Issue #6, opened 2014-12-09
Standard for "os" and "arch" definition


philips:
_From @sarnowski on December 8, 2014 15:23_

Hi,

is there a list of standard naming for the "os" and "arch" definitions?

**os**:
- linux
- windows(?)
- macosx (?)
- ... (?)

**arch**:
- amd64
- i386 (?)
- armel (?)
- ...

One cannot and should not cover all cases but I think to avoid some chaos, it makes sense to suggest some defaults.

_Copied from original issue: coreos/rocket#234_

PR #33, opened 2014-12-11
schema: expand OS/arch whitelist to allow other platforms

mpasternacki:
Introduced a map of valid OS/architecture combinations to allow
implementations supporting platforms other than linux/amd64. I can't
figure out at the moment how to warn rather than fail during JSON
validation without resorting to "just print to stderr" approach. A
workaround for an implementation would be to add own values to the
`ValidOSArch` map, but this would still not allow full flexibility.

Additionally fixed a typo in Labels.UnmarshalJSON that prevented the
validation from actually being executed (an empty `l` Labels were
validated, rather than deserialized `nl`).

Improves #6 (not a full fix, as unlisted combinations still fail)


schema/types/labels.go
```diff
@@ -2,9 +2,16 @@ package types
...
 	"encoding/json"
-	"errors"
+	"fmt"
+	"sort"
 )
 
+var ValidOSArch = map[string][]string{
+	"linux":   []string{"amd64", "i386"},
+	"freebsd": []string{"amd64", "i386", "arm"},
+	"darwin":  []string{"x86_64", "i386"},
```
stevvooe: Why is darwin inconsistent with everything else for "amd64"?

jonboulle: @stevvooe https://github.com/appc/spec/pull/33#discussion-diff-21702202

schema/types/labels.go
```diff
@@ -2,9 +2,16 @@ package types
 
 import (
 	"encoding/json"
-	"errors"
+	"fmt"
+	"sort"
 )
 
+var ValidOSArch = map[string][]string{
+	"linux":   []string{"amd64", "i386"}, // what about x86_64, i686, other platforms?
```
stevvooe: @mpasternacki Shouldn't we be providing a common interface that is invariant to the whims of hardware developers? Mapping amd64 to x86_64 is trivial and makes cross-platform understanding a lot more straightforward.

schema/types/labels.go
stevvooe: @jonboulle I see. This is a little unfortunate, since this is just being different to be different, rather than providing a benefit over GOARCH/GOOS.

steveeJ:
> I am also a little wary of amd64 vs x86_64 (would rather just alias the two as I am almost certain there is never a substantive difference, although could be proved wrong)

Would this have any downsides? If not, let's do it!

jonboulle:
I'm open to it

Stefan Junker <notifications@github.com> schrieb am Di., 1. Dez. 2015 06:04:

PR #42, opened 2014-12-15
spec: enforce absolute path to exec

jonboulle:
We should ensure that the first argument to the app.exec array is an absolute path.


yifan-gu:
@philips I missed the the reason behind this change. Actually in k8s's pod spec, the command can be in relative path. And as long as the $PATH env is given, why we need to prohibit this?
/cc @jonboulle

jonboulle:
@yifan-gu can you come up with a proposal for what path resolution would look like?

yifan-gu:
@jonboulle Will think about it tomorrow (actually today)

Issue #60, opened 2014-12-19
Clarifications on version label default value on aci discovery.


sgotti:
Hi,

should the spec define that, during the aci discovery process, if the version label isn't specified it should default to "latest" (as done in discovery/parse.go?)?

Thanks!

sgotti:
@jonboulle Something more to the above question on the "latest" value definition:

From what I asked in #73 and implemented in coreos/rocket#297 my thoughts are that the ACI imagemanifest shouldn't contain a "version" label with value "latest".

If you look at #73 "case 2" I'm imagining that "latest" is used on fetching to get the latest ACI, but the ACI in its imagemanifest has a specific version like "version" : "2.0.1" and not "latest".

What was the original idea on the latest pattern? Am I getting it correctly?

jonboulle:
@sgotti We have discussed this (and made changes) in a few different places so I am not quite sure on the latest status with this one - can we close it?

sgotti:
@jonboulle to summarize, over then #73, related to the version label, the basic doubt is this:

* Is the "latest" value for the "version" label a kind of special value?

Take this example case:
  * (case 2 of #73) On fetching I'm requesting an appname without a version label. The aci discovery will download something like https://example.com/appname-latest-linux-amd64.aci. On the http server it can be a link to an aci called appname-2.0.1-linux-amd64.aci and in its manifest the version label will be "2.0.1".

I'd like to know if this example is what you had in mind when you thought about the "latest" pattern.

If this is correct I'm thinking that:

* The ACI imagemanifest mustn't contain a "version" label with value "latest" as it will clash with the "latest" pattern.
* The ACI imagemanifest mustn't contain in its dependencies section an app with a "version" label with value "latest". As given the first point an ACI cannot contain a "version": "latest" label and so an ACI will never match the required labels (as the spec in the "dependency matching" section says that the downloaded aci's labels should be compared against the labels in the retrieved ACI)
* Should be good to define in the spec that the latest value for a version label is "the standard value to ask for the latest version". So every repository can provide it for every implementation. Leaving it free will make it different for every implementation and also for every repository maker creating a lot of combinations and different behaviors.

Or if all the above is not corrent were you thinking that my idea of the link on the http server is wrong and that can exists an ACI called appname-latest-linux-amd64.aci with a "version" : "latest" in its imagemanifest?
[...]

PR #61, opened 2014-12-19
discovery: do not set defaults for os and arch and use only fully renderable meta templates.

sgotti:
os and arch can be optional as there can be "noarch" (without machine dependent code) aci.

I'm imagining that in my repository I can have also some "noarch" packages.
For getting this, using meta discovery, I should provide a meta like this:

```
<meta name="ac-discovery" content="example.com hdfs://storage.example.com/{name}-{version}.{ext}">
```

In this way, if I'm asking for an ACI providing a name but not an os and arch the first template should fail as not completely renderable (os and arch cannot be evaluated) but the second one will work.
The second patch does this.

Plus probably the labels in meta can have any name and values so this patch will also create the template vars slice from all the available labels instead of only some predefined ones (name, version, os, arch). Or am I wrong and only some well known labels should be used (if so I think that the spec should clarify this)?

Thanks!

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -3,6 +3,7 @@ package discovery
 import (
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
 	"github.com/appc/spec/Godeps/_workspace/src/golang.org/x/net/html"
@@ -20,6 +21,10 @@ type Endpoints struct {
 	ACI  []string
 	Keys []string
 }
+
+var (
+	templateExpression = regexp.MustCompile(`{.*?}`)
+)
 
 func appendMeta(meta []acMeta, attrs []html.Attribute) []acMeta {
 	m := acMeta{}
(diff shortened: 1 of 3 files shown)


jonboulle:
>> Should this behavior be documented in the SPEC?
> 
> Yes, good call, I will add a note.

Filed https://github.com/appc/spec/pull/79

sgotti:
@jonboulle perhaps my idea to remove defaultOS and defaultArch was not so good :disappointed:
For example:

```
rkt run 'coreos.com/etcd:v2.0.0-rc.1'
```
now fails as it explicitly needs also the os and arch:
```
rkt run 'coreos.com/etcd:v2.0.0-rc.1&os=linux&arch=amd64'
```

Returning to the example where in a repository I'd like to provide both "witharch" and "noarch" ACIs:
```
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}.{ext}">
```
Trying to discover an aci without providing os and arch:

* Without setting default{Os,Arch} the first template will fail to render and the second will succeed and will be downloaded.
* Setting default{Os,Arch} both templates will render, but probably the first download will get a 404 and the other will work.

I'm starting to think the the second case will be better and the fetching function should retry all rendered templates until one succeeds.

Now my doubts are related to the fact that in the rocket code the various ac-discover meta tags are considered mirrors (see TODO from @philips in rkt/fetch.go and also cas/remote.go also if Mirrors aren't used) while my idea was to use multiple ac-discover to handle different name formats...
What is the correct use of multiple "ac-discover" meta? (mirrors? different aci name formats? both?)

Anyway, if you think that it's better I'll readd defaultOS and defaultArch (in #86 I proposed to move defaultVersion to doDiscover)

jonboulle:
@sgotti For os and arch, I am going to say that they should be handled by implementations. For example here is my proposed fix for Rocket's use of discovery:
https://github.com/coreos/rocket/pull/337

jonboulle:
@sgotti For latest, I think it's OK for it to be in the discovery "reference impl" right now, so I merged #86

sgotti:
> @sgotti For os and arch, I am going to say that they should be handled by implementations. For example here is my proposed fix for Rocket's use of discovery:
> coreos/rocket#337

@jonboulle You're right! LGTM!

> @sgotti For latest, I think it's OK for it to be in the discovery "reference impl" right now, so I merged #86

Ok! Thanks! There are some open issues about the latest pattern i tried to explain in #73 and #60 .

> Now my doubts are related to the fact that in the rocket code the various ac-discover meta tags are considered mirrors (see TODO from @philips in rkt/fetch.go and also cas/remote.go also if Mirrors aren't used) while my idea was to use multiple ac-discover to handle different name formats...
> What is the correct use of multiple "ac-discover" meta? (mirrors? different aci name formats? both?)

What do you think about the intended usage of multiple "ac-discover" meta tags?

PR #71, opened 2014-12-31
spec: Add cpu mask to isolators.

rjnagal:
This is one of the common isolator missing from the spec.
I have used the same format as the cgroup interface.

Signed-off-by: Rohit Jnagal <jnagal@google.com>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -200,8 +200,9 @@ Additional isolators will be added to this specification over time.
 
 |Name|Type|Schema|Example|
 |-------------------------|------|------------------------------------|----------------|
-|cpu/shares/              |string|"&lt;uint&gt;"                      |"4096"          |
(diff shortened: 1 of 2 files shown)


philips:
Yes, we should certainly add this. One question though: what are the semantics? I am familiar with the systemd directive of "CPUAffinity" which says which CPUs the process group can run on, are the semantics suggested by this the same?

rjnagal:
@philips Yes, this is exactly the same as CPUAffinity. It limits the CPUs for a container. Its also hierarchical, as in CPUs for  children containers have to be a subset of the parent (pod) container.

philips:
@rjnagal OK, I will merge this as-is so I don't forget about it when converting the rest of the isolators to schemas.

PR #74, opened 2015-01-04
spec: remove dependency's "root"

sgotti:
The example image manifest in the spec and in examples/image.json contain a
dependency's "root" that was removed from the spec.


jonboulle:
Agh, great catch! LGTM, ty

PR #78, opened 2015-01-05
spec: add workingDirectory to image manifest

jonboulle:
`workingDirectory` was added in 1d45bd9 but not fully specified. This adds
details to the ImageManifest definition, and also mandates that the ACE must
create the directory if it does not already exist within the rootfs in the
image.


SPEC.md
```diff
@@ -491,6 +491,7 @@ JSON Schema for the Image Manifest
     * **eventHandlers** are optional, and should be a list of eventHandler objects. eventHandlers allow the app to have several hooks based on lifecycle events. For example, you may want to execute a script before the main process starts up to download a dataset or backup onto the filesystem. An eventHandler is a simple object with two fields - an **exec** (array of strings, ACE can append or override), and a **name** (there may be only one eventHandler of a given name), which must be one of:
         * **pre-start** - will be executed and must exit before the long running main **exec** binary is launched
         * **post-stop** - if the main **exec** process is killed then this is ran. This can be used to cleanup resources in the case of clean application shutdown, but cannot be relied upon in the face of machine failure.
+    * **workingDirectory** the working directory of the launched application, relative to the application image's root (optional string, must be absolute path, defaults to "/", ACE can override). If the directory does not exist in the application's assembled rootfs, it must be created by the ACE at runtime.
```
philips: I do not think the ACE should create the directory since then we have to set the permissions, etc. If the workingDirectory does not exist then it should fail.

jonboulle: Hmm, OK, my intuition was to go this route because of my experience with this being a surprising/frustrating hurdle to a lot of simple use cases, but since this is optional anyway then that shouldn't really be an issue.

I assume it is fine if the workingDirectory exists in a dependent image rather than necessarily this one?

mpasternacki: I believe, failing when there is no workingDirectory in the assembled container is less surprising than creating that directory. I would expect that when I make a typo in workingDirectory, then the container won't start (and will tell me why). I wouldn't expect it to start in a mistyped directory that it just created (and if it did, this behavior would likely cause some frustration and debugging, because I wouldn't notice the offending typo, but I'd see unexplained behaviour after starting the application).

> I assume it is fine if the workingDirectory exists in a dependent image rather than necessarily this one?
I believe this should be checked when running the application rather than when validating the image, if this is what you are asking. The workingDirectory can also (legitimately, I think, at least I don't see a reason why it couldn't) point to a subdirectory of a volume that needs to be mounted from host, so we can't check that before we actually run.

jonboulle: Thanks for reminding me that this issue exists :-)

jonboulle:
@mpasternacki ptal

mpasternacki:
LGTM.

PR #79, opened 2015-01-05
spec: disallow labels named "name"

jonboulle:
As noted in https://github.com/appc/spec/pull/61/#issuecomment-67964235
, during discovery we have the potential confusion of a user supplying a label
with the name "name", which would essentially conflict with the actual image
name. To mitigate this, assert in the spec/schema that labels are not named
"name".


philips:
This makes me a bit nervous. But, it is a practical way of solving the template problem so I am OK with it. What I would suggest is that users namespace their labels to a domain or something too.

An alternative proposal would be to prefix the name "label" in the templates with github.com/appc or something but that is pretty cumbersome.

So, long story short, lets LGTM.

jonboulle:
> What I would suggest is that users namespace their labels to a domain or something too.

Yes, our guidance on this should be the same as all other fields like isolators and annotations IMO

PR #80, opened 2015-01-05
spec: change isolators example to list

jonboulle:
isolators was previously changed to be a list of structs rather than a struct
itself but unfortunately it seems like the actual example embedded in the
SPEC.md was not properly updated (on the other hand, the example JSON is
fine).

h/t @cdaylward


cdaylward:
LGTM!

Issue #58, opened 2014-12-19
spec: clarification for event handlers


mpasternacki:
1. If multiple event handlers of the same name are defined, should they be executed in parallel, or one after another?
2. Are event handlers executed as root, or as parent application's user/group? A good use case for handlers would be privileged initialization (e.g. creating and `chown`-ing workspace for process, which requires root) to let the main process run as a non-privileged user.
3. Do event handlers run with parent app's environment, or with a clean one?

It may be useful to have event handlers more parameterized: it should be possible to specify at least user, group, and environment for the handler. If you are not opposed to that idea, I will prepare a pull request.

jonboulle:
Closing this in favour of the slightly more specific #188

PR #82, opened 2015-01-06
schema: Modify annotations to be a list of name/value pairs.

cdaylward:
Change annotations from a map to a list of Annotation objects that represent name/values in the same manner as Labels. This maintains consistency with similar types and has the benefits already discussed in issues [58](https://github.com/appc/spec/pull/58), [62](https://github.com/appc/spec/pull/62), and [64](https://github.com/appc/spec/pull/64). Namely,1) it maintains that field names on the left are schema, values on the right are user values 2) behavior during duplicate JSON object field names are not part of the JSON spec and may be handled differently based on the library being used...but generally an ill-formed manifest will have the duplicate error swallowed during JSON parsing.

Small fixup for isolator in SPEC.md example using "value" instead of "val"


schema/types/annotations.go
```diff
@@ -1,27 +1,43 @@
 package types
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"fmt"
+)
 
 // TODO(jonboulle): this is awkward since it's inconsistent with the way we do
```
jonboulle: could you kill this TODO now please?

SPEC.md
```diff
@@ -501,7 +513,7 @@ JSON Schema for the Image Manifest
...
-* **annotations** key/value store that can be used by systems outside of the ACE (ACE can override). The key is restricted to the [AC Name](#ac-name-type) formatting. If you are defining new annotations, please consider submitting them to the specification. If you intend for your field to remain special to your application please be a good citizen and prefix an appropriate namespace to your key names. Recognized annotations include:
+* **annotations** are optional, and should be a list of annotation objects (where the *name* is restricted to the [AC Name](#ac-name-type) formatting and *val* is an arbitrary string) Annotations can be used by systems outside of the ACE (ACE can override). If you are defining new annotations, please consider submitting them to the specification. If you intend for your field to remain special to your application please be a good citizen and prefix an appropriate namespace to your key names. Recognized annotations include:
```
jonboulle: s#string) Annotations#string). Annotations#
s#key names#names#
Annotation names must be unique within the list.

schema/container.go
```diff
@@ -7,13 +7,13 @@ import (
...
-	Volumes     []types.Volume          `json:"volumes"`
-	Isolators   []types.Isolator        `json:"isolators"`
-	Annotations map[types.ACName]string `json:"annotations"`
+	ACVersion   types.SemVer       `json:"acVersion"`
+	ACKind      types.ACKind       `json:"acKind"`
+	UUID        types.UUID         `json:"uuid"`
+	Apps        AppList            `json:"apps"`
+	Volumes     []types.Volume     `json:"volumes"`
+	Isolators   []types.Isolator   `json:"isolators"`
+	Annotations []types.Annotation `json:"annotations"`
```
jonboulle: Is this intentionally `[]types.Annotation` and not `types.Annotations` because you want to avoid the validation that the latter performs? 

If so, this is a subtlety that should be called out in the code and in the CRM section of the spec.

If not, use `types.Annotations` here, and update the CRM section of the spec to clarify that the same restrictions apply as to the `Annotations` field of the Image Manifest (e.g. "**annotations** arbitrary metadata appended to the app. Must be of the same type as the `annotations` field in the Image Manifest" or so)

jonboulle:
awesome, thank you. @philips can you ack this?

philips:
ack, lgtm

philips:
I am going to merge this but it seems like val is one of the only abbreviations we have in the schema.

jonboulle:
> I am going to merge this but it seems like val is one of the only abbreviations we have in the schema.
 
@philips are you implying we should just change it to "value" everywhere?

cdaylward:
I feel it should be. It is odd to have one unnecessary abbreviation.

PR #85, opened 2015-01-06
Switch remaining Fprintf(os.stderr) to stderr()

jbergstroem:
Because nitpick

### DIFF
--- a/ace/validator.go
+++ b/ace/validator.go
@@ -91,7 +91,7 @@ type results []error
 // main outputs diagnostic information to stderr and exits 1 if validation fails
 func main() {
 	if len(os.Args) != 2 {
-		fmt.Fprintf(os.Stderr, "usage: %s [main|sidekick|preStart|postStop]\n", os.Args[0])
+		stderr("usage: %s [main|sidekick|preStart|postStop]", os.Args[0])
 		os.Exit(64)
 	}
 	mode := os.Args[1]
@@ -106,7 +106,7 @@ func main() {
 	case "poststop":
 		res = validatePoststop()
 	default:
-		fmt.Fprintf(os.Stderr, "unrecognized mode: %s\b", mode)
+		stderr("unrecognized mode: %s", mode)
 		os.Exit(64)
 	}
 	if len(res) == 0 {
--- a/actool/discover.go
+++ b/actool/discover.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"os"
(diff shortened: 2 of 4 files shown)


actool/discover.go
```diff
@@ -25,7 +25,7 @@ func init() {
 
 func runDiscover(args []string) (exit int) {
 	if len(args) < 1 {
-		fmt.Fprintf(os.Stderr, "discover: at least one name required")
+		fmt.Fprintf(os.Stderr, "discover: at least one name required\n")
```
jonboulle: Thanks for the fix. Would you mind actually just changing this to use the [`stderr` helper function](https://github.com/appc/spec/blob/master/actool/actool.go#L99) for consistency with elsehwere?

jbergstroem: Sure. Will update shortly.

jbergstroem:
I found a few additional places where the same pattern were used and replaced them as well. I avoided some corner cases where newline wasn't expected. Hope that's fine.

ace/validator.go
```diff
@@ -106,7 +106,7 @@ func main() {
 	case "poststop":
 		res = validatePoststop()
 	default:
-		fmt.Fprintf(os.Stderr, "unrecognized mode: %s\b", mode)
+		stderr("unrecognized mode: %s\b", mode)
```
jonboulle: Doh! The `\b` is a bug, could you remove it please?

jbergstroem:
Fixed.

philips:
lgtm, merging.

jonboulle:
This jumped the gun a bit :/
```
gopath/src/github.com/appc/spec/ace/validator.go:94: undefined: stderr
```

jbergstroem:
@jonboulle strange, I never got that error. Just tried building on both os x/linux with go 1.4 (ace not built on OS X).

jonboulle:
@jbergstroem I subsequently fixed it in https://github.com/appc/spec/commit/a4cff3a3f49eeeecc5cd05ad80ede739b9590392 (and enabled travis for the repo, so we should catch these in future!)

jbergstroem:
@jonboulle Got it. Thanks, appreciated. Fwiw, the issue was me not testing on linux (only os x) before putting my new round of patches up for review. Apologies.

jonboulle:
@jbergstroem No worries, it's clearly a bug in our process that we didn't have travis enabled, so thanks for catching that ;-)

Issue #73, opened 2015-01-04
Clarifications on dependency matching and split between fetching and rendering.


sgotti:
The spec in the "Dependency matching" chapter says that every dependency should be discovered using the ACI discovery process and then, just the downloaded image should be verified with the requested labels or matching ImageID.

But, in appc/spec#16 this process is split in two phases: ACI fetching and ACI rendering.
I think that the splitting between the fetching phase and the rendering phase is good. Someone can, for example, not fetch the image from a repository but just import an aci manually in the store. Or there can be an offline/caching mode.

So the ACI rendering will be executed independently from the ACI fetching.
In the ACI rendering phase the ACI will be requested to the store providing an imageID or an app name and some labels. In the latter case the store can find more than one ACI that matches the requested App name and labels.

Which one should be returned?

I have some possible cases in mind:

### Case 1 ###
If I'm requesting to the store an App name without any label, all the ACIs with that App Name will match your request.

Which one should be returned?
Should it be implementation dependent or should it be defined in the spec?

The current implementation i tried to do in coreos/rocket#297 uses the latest aci imported in the store.

### Case 2 ###
Another example can be the "latest" pattern: If I'm requesting to the store an App without any version label, how can I satisfy the "latest" pattern?

* Fetch phase: In this phase, as I'm not passing a specific version label, the aci discovery will download something like https://example.com/appname-latest-linux-amd64.aci. On the http server it can be a link to an aci called appname-2.0.1-linux-amd64.aci and in its manifest the version label will be "2.0.1".

* Rendering phase: I'm requesting from the store an ACI named appname and without any version label. The store can have other ACIs previously downloaded or imported for this app name with different version (for example "1.0.0" "strangereleasename" etc...).
As the version label is just a string without any SemVer or lexical ordering rule there isn't an ordering in these labels.

As in [Case 1](#Case 1) the implementation i tried to do in coreos/rocket#297 chooses the latest imported ACI but prefers ACIs imported with the "latest" flag (its saved in an appinfo additional index):

On fetching, if downloaded with the "latest" pattern, the ACI is saved in the store with the "latest" flag. When requesting an aci without specifying a version label the ones with the latest flag are preferred over the others (or perhaps only the one with the latest flag should be choosed?).

* Are my thoughts correct?
* If so, there should be possible ACI choosing logics other then the one I tried to implement? Should this logics be clarified in the SPEC?
* If so, should the spec be clarified (also with some examples)?

sgotti:
[...]

PR #86, opened 2015-01-06
discovery: set default version label in doDiscover and not in NewAppFromString

sgotti:
This is related to #60 and #73. If my thought are wrong please let me know.
If correct probably this should be explained in the spec and maybe also #4 and maybe the app parsing functions should be moved in a new lib outside discovery?

The app and labels serialization format parsed by NewAppFromString is not
strictly related to discovery/fetching but should also be used by other
functions (eg. asking an image to a local store without discovery and fetching).

Setting a default version label if it's missing shouldn't be done here but from
the functions that needs it (now in doDiscover).

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -22,6 +22,10 @@ type Endpoints struct {
 	Keys []string
 }
 
+const (
+	defaultVersion = "latest"
+)
+
 var (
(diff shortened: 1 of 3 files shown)


jonboulle:
This makes sense to me as an immediate improvement; we will likely adjust this in future but I agree with your thinking on relocating this for now. Let's roll forward.

PR #87, opened 2015-01-06
discovery: httpsOrHTTP: fix error management

alban:
Symptoms:
$ sudo ./bin/rkt run foo
panic: runtime error: invalid memory address or nil pointer dereference

Fixed by checking if res is nil.

Additionally, only release res.Body with closeBody(res) if the http method is
attempted. Otherwise, it would be released a second time at the end of the
function.


jonboulle:
Bah, thanks for fixing this, looks like it crept through recently.  
How do you feel about creating a test for this function? :-)

alban:
Good idea. I added a test in the next commit :)

jonboulle:
@alban awesome - thanks for the patch! LGTM

PR #89, opened 2015-01-07
acevalidator: fix stderr reference

jonboulle:
4b128a485bc00687dc8d6f3bcb1adbe6e2c255dd tried to standardise on stderr() but
unfortunately updated the validator to use the function when it's in a
different package; this wasn't caught since we did not have travis confiugred.
For now just introduce the same function to this package.


jonboulle:
This does not quite fix the build, there is still some fallout from #82, but I will clear that up in subsequent commit.

PR #102, opened 2015-01-09
aci: add helper functions for working with ACI images

kelseyhightower:
Currently there is no easy way to work with ACI images. Implementations
of the App Container Spec must understand low level details of the ACI
package format in order to complete simple tasks such as extracting
metadata and detecting the type of compression being used.

The ACI package now includes three helper functions for working with ACI
images: ManifestFromImage, NewCompressedTarReader, and NewCompressedReader.

### DIFF
--- a/aci/file.go
+++ b/aci/file.go
@@ -1,12 +1,19 @@
 package aci
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/bzip2"
+	"compress/gzip"
 	"encoding/hex"
+	"errors"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os/exec"
+
+	"github.com/appc/spec/schema"
 )
 
 type FileType string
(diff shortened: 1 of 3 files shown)


kelseyhightower:
Do not merge! Work in progress.

aci/file.go
```diff
@@ -100,3 +107,87 @@ func XzReader(r io.Reader) io.ReadCloser {
...
+	var im schema.ImageManifest
+
+	tr, err := NewCompressedTarReader(rs)
+	if err != nil {
+		return nil, err
+	}
+
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
```
jonboulle: this is kind of superfluous since the manifest branch breaks out of the loop anyway

aci/file.go
```diff
@@ -100,3 +107,87 @@ func XzReader(r io.Reader) io.ReadCloser {
...
+func NewCompressedTarReader(rs io.ReadSeeker) (*tar.Reader, error) {
+	cr, err := NewCompressedReader(rs)
+	if err != nil {
+		return nil, err
+	}
+	return tar.NewReader(cr), nil
+}
+
+// NewCompressedReader creates a new io.Reader from the given ACI image.
+func NewCompressedReader(rs io.ReadSeeker) (io.Reader, error) {
```
jonboulle: While `ReadSeeker` is a little more pure, arguably `Reader` is a bit more practical (e.g. for http response bodies). See https://github.com/coreos/rocket/blob/master/cas/cas.go#L128 for example

aci/file.go
```diff
@@ -100,3 +107,87 @@ func XzReader(r io.Reader) io.ReadCloser {
 
 	return rpipe
 }
+
+// ManifestFromImage creates a new schema.ImageManifest from the given ACI image.
```
jonboulle: how about extracts rather than creates

aci/file.go
jonboulle: Actually this is buggy because if the reader does not have a manifest at all, then the function will return a zero-valued Imagemanifest, which is not what you want (nil)

aci/file.go
kelseyhightower: I'm using a `ReadSeeker` here because I want to avoid copying the images. Seems more efficient to "rest".

jonboulle: > Seems more efficient to "rest".

I don't understand, can you elaborate? In the code in my previous link, we only copy the first 512 bytes to sniff the type, but nothing more..

kelseyhightower: That 512 copy moves the position of the reader and without using a `ReadSeeker`, I don't have an easy way to rewind.

jonboulle: Why do you need to rewind at the start of this function? Also, since this function returns a Reader, the buffered approach is fine IMHO (so you don't need to rewind after sniffing the type).

test
```diff
@@ -14,7 +14,7 @@ COVER=${COVER:-"-cover"}
 
 source ./build
 
-TESTABLE_AND_FORMATTABLE="discovery schema/types"
+TESTABLE_AND_FORMATTABLE="aci discovery schema/types"
 FORMATTABLE="$TESTABLE_AND_FORMATTABLE ace aci actool schema pkg/tarheader"
```
jonboulle: Remove "aci" from here

aci/file.go
```diff
@@ -100,3 +107,89 @@ func XzReader(r io.Reader) io.ReadCloser {
...
+// NewCompressedTarReader creates a new tar.Reader reading from the given ACI image.
+func NewCompressedTarReader(rs io.ReadSeeker) (*tar.Reader, error) {
+	cr, err := NewCompressedReader(rs)
+	if err != nil {
+		return nil, err
+	}
+	return tar.NewReader(cr), nil
+}
+
+// NewCompressedReader creates a new io.Reader from the given ACI image.
```
jonboulle: I guess we should really deduplicate this with cas/cas.go in rocket..

jonboulle:
@kelseyhightower consider the Reader/ReadSeeker thing non-blocking, feel free to land this whenever. LGTM.

Issue #84, opened 2015-01-06
spec: clarify relationship between CRM and "apps"


jonboulle:
(This is partly leading on from the discussion in https://github.com/coreos/rocket/issues/294)

A [Container Runtime Manifest](https://github.com/appc/spec/blob/master/SPEC.md#container-runtime-manifest-schema) declares one or more apps to be executed. At the time of writing the format of declaration looks like this:
> * **apps** the list of apps that will execute inside of this container
    * **app** the name of the app (string, restricted to AC Name formatting)
    * **imageID** the content hash of the image that this app will execute inside of (string, must be of the format "type-value", where "type" is "sha512" and value is the hex encoded string of the hash)
    * **isolators** the list of isolators that should be applied to this app (key is restricted to the AC Name formatting and the value can be a freeform string)
    * **annotations** arbitrary metadata appended to the app (key is restricted to the AC Name formatting and the value can be a freeform string)

There are several issues with this schema:
- Each "app" technically refers to an app _image_ (ACI), which may or may not actually contain an **app** declaration (since "app" is an _optional_ section of the Image Manifest). The spec is unclear as to what will occur if the referenced image does _not_ contain an app declaration. (Relatedly, "app" should likely more accurately be called "name")
- "app" references are not unique. There is a fairly simple use case of wanting to use a particular app-image multiple times within a container, but it is unclear how this behaviour should be represented in the spec, or at least how implementations should go about handling it (currently in Rocket we actually [mandate that apps within a CRM must be unique](https://github.com/coreos/rocket/blob/master/stage0/run.go#L156))
- For each app, there is only a partial ability to override some of the parameters defined in the Image Manifest. For example, currently **isolators** (from the **app** section of the Image Manifest) and **annotations** can be overridden/extended; but there is no such ability to override the **exec**, **eventHandlers** or **environment**

This is closely related to, but not the same as, https://github.com/appc/spec/issues/83

One possible solution is to extend the schema of app references in the CRM to match that of the Image Manifest. In this way, all of the parameters in the Image Manifest could be explicitly overridden at a later stage. However, this does potentially mean a lot of duplication between the information contained in the IM and the CRM.

wereHamster:
The first issue was clarified in #72.

jonboulle:
@wereHamster I don't think we properly understood your question there; I think we should define (or at least describe) the expected behaviour in the spec.

wereHamster:
[...]

Issue #83, opened 2015-01-06
spec: clarify relationship between CRM and ACE


jonboulle:
There is a possible conflation between the CRM and the ACE runtime environment, which might need to be teased apart. In the lifecycle of an app image, between the points of "image manifest creation" (when it is bundled into an ACI) and "application execution" (when the ACE actually executes app(s) in a container), there are conceptually two points at which parameters from the image manifest can potentially be overridden/extended:
    1. when the CRM is generated 
    2. at runtime by the ACE

Currently, in rocket, we generate a CRM at runtime and use that as the final artefact to be executed (let's call that the _executed unit_). That is, rocket essentially conflates 1) and 2), by encoding all of its "ACE overriding" into the generated CRM. 

There is a lot of value in this model because it means the CRM is essentially immutable, and so for a given executor, two invocations of the same CRM should be effectively identical. (It also makes it convenient because the executor can make various assumptions about the environment - for example where images are located, whether they are unique, etc - when generating the CRM.

However, critically, it makes it impossible to use CRMs as a portable config (let's call that a _deployable unit_), because it inherently contains executor-specific information.

Here are two potential solutions:
- Have the CRM be the _deployable unit_, and allow ACEs to override parameters at runtime. This has the disadvantage that there is now a discrepancy between what the CRM defines and what is actually being executed.
- Have the CRM be the _deployable unit_ and introduce _another_ CRM as the _executable unit_ (which may or may not be identical to the _deployable unit_); the ACE consumes a CRM and then must generate a new CRM (overriding/extending whatever parameters it desires) to execute. This retains the advantage of the CRM-that-is-executed being immutable, but introduces the potentially confusing discrepancy between the CRM supplied to the ACE (deployable unit) and the CRM that is actually run (executable unit).

cdaylward:
Consider the following basic toolchains:
* A minimal CLI tool that accepts a raw JSON CRM as input and executes the container specified therein.
* A more flexible CLI tool (e.g. Rocket) that takes command line arguments and determines what CRM to generate.
* CRM as a "task" specification submitted to a large...elastic compute environment.

[...]

PR #106, opened 2015-01-12
schema: Add labels to Container Runtime Manifest

cdaylward:
# Issue
An ACE implementation needs additional information about images that does not exist in the CRM. Currently, information in the CRM needs to be combined with user input to perform image discovery. Namely, the app name, image ID (really, a checksum), needs to be combined with labels as these three fields are together used to uniquely identify an image. In a scenario where system components are highly decoupled, an ACE implementation cannot use the CRM to "start work".

# Proposal
Add labels to the CRM. This fills the information gap (all image identifiers are now available). A CRM can then be passed between components without requiring additional information to be passed along to "start work". The CRM becomes _the_ message and can be used to remotely start a task. Essentially, this enables the use of the exact same image discovery mechanism that is used to resolve inter-image dependencies. Without it, an implementation must know a priori where to fetch an image given the imageID.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -606,6 +606,7 @@ JSON Schema for the Container Runtime Manifest (container manifest)
 * **uuid** an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt) that represents this instance of the container (string, must be in [RFC4122](http://www.ietf.org/rfc/rfc4122.txt) format)
 * **apps** the list of apps that will execute inside of this container
     * **app** the name of the app (string, restricted to AC Name formatting)
+    * **labels** are optional, and should be a list of label objects as specified under[Image Manifest Schema](#image-manifest-schema).
     * **imageID** the content hash of the image that this app will execute inside of (string, must be of the format "type-value", where "type" is "sha512" and value is the hex encoded string of the hash)
     * **isolators** the list of isolators that should be applied to this app (key is restricted to the AC Name formatting and the value can be a freeform string)
     * **annotations** optional arbitrary metadata appended to the app. Should be a list of annotation objects (where the *name* is restricted to the [AC Name](#ac-name-type) formatting and *value* is an arbitrary string). Annotation names must be unique within the list. These will be merged with annotations provided by the image manifest when queried via the metadata service; values in this list take precedence over those in the image manifest.


cdaylward:
If this seems reasonable, I can add to the examples, validator, etc.

jonboulle:
> An ACE implementation needs additional information about images that does not exist in the CRM. Currently, information in the CRM needs to be combined with user input to perform image discovery. Namely, the app name, image ID (really, a checksum), needs to be combined with labels as these three fields are together used to uniquely identify an image.

This diagnosis is not quite accurate IMO. Tautologically, the only information necessary to uniquely identify an image is the `imageID`. Labels are really just a convenience for users/discovery; without `imageID`, there is _not_ a guarantee that a particular set of (`names`, `labels`) will globally or consistently resolve to the same image (and conversely, different sets of (`names`, `labels`) could very well resolve to a single image). Hence, since `imageID` is a mandatory field, there is no ambiguity in the CRM as far as _which_ images are referred to, and with the current situation an ACE can very well use a CRM to start work as long as it can locate those `imageID`s. (Consider for example, an ACE with access to a global content-addressable-store: there is no missing information required to then resolve the image dependencies as it can simply retrieve from this store based on `imageID` - this is effectively what Rocket does today).

So I am not so sure this is an "information gap". Actually, I feel this is yet another symptom of the indeterminism described in #83 and #84.

cdaylward:
_cross-posting from IRC_

Discovery as it is discussed in the specification is a bit of a gray area between policy and mechanism. It's policy (this, then that) but there is _also_ a pretty strong implicit indication of "obvious mechanism". It _may_ use CAS mechanisms (imageID), or it _may_ use indexing based on precisely laid-out identifiers (e.g. version/os/arch tuple).

In the CRM portion of the spec however, the CRM has policy without mechanism, but the policy actually implicitly restricts the mechanism in a way that is not the case for the IM part of the spec. I.e. suddenly CAS is assumed and due to the CRM schema, now an "implicitly forced mechanism".

If I _may_ use CAS _or_ identifiers at one step (dependency resolution), doesn't it seem that I should be able to use either at another? Why the restriction?

I _do_ get the point about labels not guaranteeing a unique image. But they are still specified as part of the lookup process, why not afford that to the CRM consumer? That is, the section on "Dependency Matching" should be afforded to both.

In Dependency Matching, imageID is optional and labels _may_ be used. If labels do not guarantee a unique result, then dependency resolution in an IM is time-variant (I can get different results for the same request), but the consumer of a CRM has time-invariant lookup (imageID). I am approaching this from the perspective that a CRM is akin to a “statement of work” that can be passed around amongst components…which I understand is not exactly determined yet, so perhaps this should wait for resolution on issue/#83 and  issue/#84

sgotti:
@cdaylward To summarize if I understand it correctly:

As the idea is to also run a container starting from an user provided CRM (over than, like now, starting a container with one or more user provided ACIs provided in various way: a discovery string (appname + labels as defined in discovery/parse.go NewAppFromString, imageID or file) there's a need for the ACE to discover and fetch the apps requested in the CRM.

Providing only the ImageID like now will require the ACE to already have the ACIs in a store, as imageID cannot be used for discovery.

So a discovery mechanism like the one applied for dependencies resolution is needed if the ACIs are not already in the store.

Did I get it right?

Plus I see some concerns on how the image discovery mechanism is defined in the spec. I have some similar doubts and thoughts that I tried to summarize in #73 derived from the work I did in coreos/rocket#356 needed for coreos/rocket#297). Perhaps you can take a look at these issues/PR.

cdaylward:
@sgotti, yes, I believe that is a fair restatement. It should be possible that the CRM is the ACE "entry point" and not have an implicit assumption that whatever generated the CRM also performed various lookups and stores.

cdaylward:
Subsumed by #207

PR #111, opened 2015-01-13
spec: describe behavior of HTTP redirection during image discovery

### LINKED ISSUE #97: spec: HTTP Redirection during image discovery?
The discovery spec says:

> If the first attempt at fetching the discovery URL returns a status code other than 200 OK or does not   contain any ac-discovery meta tags then the next higher path in the name should be tried.

I think the spec should specifically state what it does if it receives a 3xx response.

I'd vote for following redirects as while good URLs should work forever, we all know that's not how the real world works.

kelseyhightower:
Currently the spec does not provided any details regarding the behavior
an implementation should exhibit when an HTTP redirect is encountered
during the image discovery process.

Add clarity by suggesting that HTTP redirects should be followed if
discovery URL returns a 3xx status code.

Fixes #97

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -281,8 +281,10 @@ keys: https://example.com/pubkeys.gpg
 
 This mechanism is only used for discovery of contents URLs.
 
-If the first attempt at fetching the discovery URL returns a status code other than `200 OK` or does not contain any `ac-discovery` meta tags then the next higher path in the name should be tried.
+If the first attempt at fetching the discovery URL returns a status code other than `200 OK`, `3xx`, or does not contain any `ac-discovery` meta tags then the next higher path in the name should be tried.
 For example if the user has `example.com/project/subproject` and we first try `example.com/project/subproject` but don't find a meta tag then try `example.com/project` then try `example.com`.
+
+All HTTP redirects should be followed when the discovery URL returns a `3xx` status code.
 
 Anything implementing this spec should enforce any signing rules set in place by the operator, and ensure the image manifest provided by the fetched ACI are all prefixed from the same domain.
 


SPEC.md
```diff
@@ -281,9 +281,11 @@ keys: https://example.com/pubkeys.gpg
 
 This mechanism is only used for discovery of contents URLs.
 
-If the first attempt at fetching the discovery URL returns a status code other than `200 OK` or does not contain any `ac-discovery` meta tags then the next higher path in the name should be tried.
+If the first attempt at fetching the discovery URL returns a status code other than `200 OK`, 3xx, or does not contain any `ac-discovery` meta tags then the next higher path in the name should be tried.
```
jonboulle: `3XX` ?

SPEC.md
```diff
@@ -281,9 +281,11 @@ keys: https://example.com/pubkeys.gpg
...
 For example if the user has `example.com/project/subproject` and we first try `example.com/project/subproject` but don't find a meta tag then try `example.com/project` then try `example.com`.
 
+All HTTP redirects should be followed when the discovery URL returns a 3xx status code.
```
jonboulle: should or must? (I am happy with should actually, discovery is not intended to be the _strictest_ part of the spec)

kelseyhightower: Should.

SPEC.md
kelseyhightower: I'm following `xx` convention from the [RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)

jonboulle: Oh sorry I just meant put it in backticks to be consistent with 200, so `3xx`

jonboulle:
lgtm.

PR #115, opened 2015-01-13
spec: authentication on discovery endpoints is optional

kelseyhightower:


SPEC.md
```diff
@@ -288,6 +288,10 @@ Anything implementing this spec should enforce any signing rules set in place by
 
 Discovery URLs that require interpolation are [RFC6570](https://tools.ietf.org/html/rfc6570) URI templates.
 
+### Discovery Authentication
+
+Authentication during the discovery process is optional. If an attempt at fetching the discovery URL, an app container image, or signature returns a `401 Unauthorized` implementations should enforce any authentication rules set by the operator.
```
jonboulle: "implementations enforce any authentication rules" doesn't make any sense to me in this context since the "implementation" refers to the client..

kelseyhightower: If the operator only wants to support endpoints that do basic auth over https for example.

jonboulle: OK, how about "implementations should enact the authentication policy set by the operator (for example, ...)

jonboulle:
I think this can be tweaked further, but to save the back and forth I am going to merge and follow up

Issue #114, opened 2015-01-13
discovery: be more discerning about HTTP errors


jonboulle:
Right now discovery just recursively walks up the tree on any error, alternately trying https and http. For example, when discovering the name `example.com/foo/bar/baz`, it will make requests like so:
`https://example.com/foo/bar/baz`
`http://example.com/foo/bar/baz`
`https://example.com/foo/bar/`
`http://example.com/foo/bar/`
...

However for a certain class of errors, it doesn't make much sense to do this; for example if dialing `example.com:443` times out the first time it's probably not worth retrying it for every branch. Some errors should just propagate directly (fail early)

kelseyhightower:
I think it's Ok to walk every branch and timeout along the way if the timeouts where quick. But it does make more sense to skip walking the tree if the host is unreachable for a given endpoint (scheme + host).

jonboulle:
The spec would also need to be updated to reflect this.

PR #118, opened 2015-01-15
discovery: implement simple discovery

sgotti:
This patch needs coreos/rocket#382 implemented in rocket (or fetch will fail as it will use only the first simple discovery url).

When applied, as it changes the name of DiscoverEndpoints to MetaDiscoverEndpoints, rocket should be changed to handle this different name (in rkt/fetch.go, done in coreos/rocket#383)

Implement simple discovery as defined in the spec.
If insecure is true, also the http scheme, in addition to https is used to
render the simple template.
This adds a new function for simple discovery so users can separately
call Simple and Meta discovery (and get per discovery type errors).
Fix and improve tests.

### DIFF
--- a/actool/discover.go
+++ b/actool/discover.go
@@ -34,15 +34,15 @@ func runDiscover(args []string) (exit int) {
 			return 1
 		}
 		eps, err := discovery.DiscoverEndpoints(*app, transportFlags.Insecure)
-
 		if err != nil {
 			stderr("error fetching %s: %s", name, err)
 			return 1
 		}
-		for _, list := range [][]string{eps.Sig, eps.ACI, eps.Keys} {
-			if len(list) != 0 {
-				fmt.Println(strings.Join(list, ","))
-			}
+		for _, aciEndpoint := range eps.ACIEndpoints {
+			fmt.Println("ACI: %s, Sig: %s\n", aciEndpoint.ACI, aciEndpoint.Sig)
+		}
+		if len(eps.Keys) > 0 {
+			fmt.Println("Keys: " + strings.Join(eps.Keys, ","))
 		}
 	}
 
--- a/actool/discover.go
+++ b/actool/discover.go
@@ -33,10 +33,21 @@ func runDiscover(args []string) (exit int) {
 			stderr("%s: %s", name, err)
 			return 1
 		}
-		eps, err := discovery.DiscoverEndpoints(*app, transportFlags.Insecure)
+		eps := &discovery.Endpoints{}
+		simpleEps, err := discovery.SimpleDiscoverEndpoints(*app, transportFlags.Insecure)
 		if err != nil {
-			stderr("error fetching %s: %s", name, err)
-			return 1
+			stderr("error doing simple discovery for %s: %s", name, err)
+		} else {
+			eps.ACIEndpoints = append(eps.ACIEndpoints, simpleEps.ACIEndpoints...)
+			eps.Keys = append(eps.Keys, simpleEps.Keys...)
+		}
+
+		metaEps, err := discovery.MetaDiscoverEndpoints(*app, transportFlags.Insecure)
+		if err != nil {
+			stderr("error doing matadata discovery for %s: %s", name, err)
+		} else {
+			eps.ACIEndpoints = append(eps.ACIEndpoints, metaEps.ACIEndpoints...)
+			eps.Keys = append(eps.Keys, metaEps.Keys...)
 		}
 		for _, aciEndpoint := range eps.ACIEndpoints {
 			fmt.Println("ACI: %s, Sig: %s\n", aciEndpoint.ACI, aciEndpoint.Sig)
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -16,10 +16,14 @@ type acMeta struct {
 	uri    string
 }
 
+type ACIEndpoint struct {
+	ACI string
+	Sig string
+}
+
 type Endpoints struct {
-	Sig  []string
-	ACI  []string
-	Keys []string
+	ACIEndpoints []ACIEndpoint
+	Keys         []string
 }
 
 const (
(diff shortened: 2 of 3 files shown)


discovery/discovery.go
```diff
@@ -161,3 +161,34 @@ func DiscoverEndpoints(app App, insecure bool) (eps *Endpoints, err error) {
...
+	de := &Endpoints{}
+
+	// Simple Discovery endpoints
+	schemes := []string{"https"}
+	if insecure {
+		schemes = append(schemes, "http")
+	}
+	for _, scheme := range schemes {
+		// Ignore not handled variables as {ext} isn't already rendered.
+		uri, _ := renderTemplate(fmt.Sprintf("%s://{name}-{version}-{os}-{arch}.{ext}", scheme), tplVars...)
```
jonboulle: Can you split the tmpl into a const?

discovery/discovery.go
```diff
@@ -148,12 +148,12 @@ func doDiscover(app App, pre string, insecure bool) (*Endpoints, error) {
 // DiscoverEndpoints will make HTTPS requests to find the ac-discovery meta
```
jonboulle: I am thinking that we would really just expose one external function, `DiscoverEndpoints`, which would encapsulate both simple and meta.

discovery/discovery.go
```diff
@@ -161,3 +161,34 @@ func DiscoverEndpoints(app App, insecure bool) (eps *Endpoints, err error) {
...
+		schemes = append(schemes, "http")
+	}
+	for _, scheme := range schemes {
+		// Ignore not handled variables as {ext} isn't already rendered.
+		uri, _ := renderTemplate(fmt.Sprintf("%s://{name}-{version}-{os}-{arch}.{ext}", scheme), tplVars...)
+		sig, ok := renderTemplate(uri, "{ext}", "sig")
+		if ok {
+			de.Sig = append(de.Sig, sig)
+		}
+		aci, ok := renderTemplate(uri, "{ext}", "aci")
```
jonboulle: Serving a sig over HTTP but an ACI over HTTPS (or vice versa) seems weird, should we really support this case?

discovery/discovery.go
sgotti: I had the same doubt. I splitted parse and meta because I was thinking that maybe rocket will report if one of the two failed and for which reason (see https://github.com/sgotti/rocket/commit/d2389180c779864750e67c6ae3b92a09a4f88e96). But if this isn't needed I'll merge the two ep inside DiscoverEndpoints and return an error only if both discovery methods have failed.

discovery/discovery.go
sgotti: I think that this is strange due to how Endpoints struct is defined.
Effectively sig and aci must both render or fail, and they must be used togheter (ep.ACI[i] with ap.ACI[i]) (see also coreos/rocket#382).

Maybe refactoring Endpoint tying together aci and sig will make it clear.

sgotti:
I just pushed a new version as I previously wrongly pushed an old one (actool wasn't fixed for new functions names). The comments on the previous commit version remain open.

discovery/discovery.go
jonboulle: Yeah, for now I think it should be opaque to users, so please combine them. If simple discovery fails with an error (as opposed to just not finding anything), it should fail immediately imho.

sgotti: @jonboulle Thinking more about it, I see some problems doing an unique DiscoverEndpoints:

*) The spec says that first simple discovery should be tried, then, if nothing can be downloaded, meta discovery should be tried.
For the same reasons explained in #114, if a simple discovery template is generated but there's a connection error (not 404 or other http errors), I think that meta discovery shouldn't be tried and the fetching process should stop here with an error.
So there's the need for the functions using discovery (like in rkt/fetch.go) to first get just the simple discovery urls, try to download them, and, based on the received error, try meta discovery or fail.

*) If an unique DiscoverEndpoints function is used it should combine simplediscovery url with metadiscovery url (putting simple ones first in the arrays). But if meta discovery fails for reasons explained in #114 it should return an error and so simple discovery won't be done (going against the spec).

discovery/discovery.go
sgotti: @jonboulle I opened PR #132 trying to avoid this confusion and rebased this on top of it.

sgotti:
Rebased on top of #132.

discovery/discovery.go
jonboulle: @sgotti Sorry for the silence, I have been mulling on this for a few days as I am not really happy with the current relationship between simple/meta. What do you think of the discussion in #140 ?

actool/discover.go
```diff
@@ -33,16 +33,27 @@ func runDiscover(args []string) (exit int) {
...
+		} else {
+			eps.ACIEndpoints = append(eps.ACIEndpoints, simpleEps.ACIEndpoints...)
+			eps.Keys = append(eps.Keys, simpleEps.Keys...)
+		}
 
+		metaEps, err := discovery.MetaDiscoverEndpoints(*app, transportFlags.Insecure)
 		if err != nil {
-			stderr("error fetching %s: %s", name, err)
-			return 1
+			stderr("error doing matadata discovery for %s: %s", name, err)
```
jonboulle: s#matadata#meta#

discovery/discovery.go
sgotti: @jonboulle You're right. I tried to summarize the various doubts (and the already opened issues about discovery and label matching) in #140.
Feel free to close this and the related PR (also on coreos/rocket) as I think that this is going to be outdated.

philips:
@sgotti What is going on with this? I am confused on what the next steps are.

sgotti:
@philips This PR tries to implement the simple discovery behavior actually explained in the SPEC. But there's a related discussion on discovery in #140 (see also this comment: https://github.com/appc/spec/pull/118#discussion_r23422541) so I was waiting for it to advance.

jonboulle:
I am going to close this in favour of @cdaylward chasing up #140, we can re-open if necessary. Sorry about the runaround.

PR #119, opened 2015-01-15
Fix Annotations.Set()

mpasternacki:
`Annotations.Set` has been `append()`ing new Annotation to a private copy of the slice, which had effect only inside the `Set` function, unless `Annotations`' capacity has been greater than its length. This causes inconsistent behaviour, and – in most cases – no visible effect of setting a new annotation. If `Annotations.Set` behaves like `append` that it uses – i.e. returns an updated slice – the caller is able to replace old annotations slice with the updated one.
Second commit `SetAnnotation` methods on `ImageManifest` and `RuntimeApp`, to make `Annotations.Set` more useful, and `GetAnnotation` on `RuntimeApp` for consistency.

### DIFF
--- a/schema/container.go
+++ b/schema/container.go
@@ -75,3 +75,11 @@ type RuntimeApp struct {
 	Isolators   []types.Isolator  `json:"isolators"`
 	Annotations types.Annotations `json:"annotations"`
 }
+
+func (app *RuntimeApp) GetAnnotation(name string) (val string, ok bool) {
+	return app.Annotations.Get(name)
+}
+
+func (app *RuntimeApp) SetAnnotation(name types.ACName, value string) {
+	app.Annotations = app.Annotations.Set(name, value)
+}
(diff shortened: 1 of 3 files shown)


jonboulle:
@mpasternacki well that is stupid and dumb, `Set()` with a value receiver never makes sense :-(. Thanks for catching.

However I'm not really sure why your approach is preferable to just fixing `Set` - see https://github.com/jonboulle/spec/commit/ebe057b0fbb52b64e3b94983e4607f4678cc8832. This makes it more consistent with the name and other `Set` functions in the types package.

mpasternacki:
Yes, your approach seems much better than mine. I haven't thought about that, even though it's scattered all over the code - just solved my immediate issue of annotations not being set. Closing this one.

jonboulle:
https://github.com/appc/spec/pull/120

PR #120, opened 2015-01-15
types: fix annotations Set method

jonboulle:
Set was acting on an Annotations value instead of a pointer, which is never 
going to work for obvious reasons. This changes the receiver to a pointer and
corrects the logic to properly set the value in-place if necessary.

Also adds a test case for Annotations.Set which should have been there all
along and catches this issue.

Thanks to @mpasternacki for catching this silliness.


mpasternacki:
:+1: this will let me revert an ugly hot-patch in Jetpack

Issue #109, opened 2015-01-13
spec: missing expected behaviour


frntn:
This is more like a **general purpose issue** to mention that the specs lacks of informations on expected behaviours. While reading the spec, many times I thought "how would I handle it with this particular use case if I had to implement it ?"

For example for `pathWhitelist` : 

- does the runtime (container executor) juste filter out the allowed paths from dependency acis, or will it raise an error if some dependency container contains files outside the allowed paths ?
- **"This field is only required if the app has dependencies and you wish to remove files from the rootfs before running the container"** => Ok i now have my answer (files from dependency container but outside the folder path are just dropped) but will it raise an error if the field is missing ? (does "required" means "mandatory" here, not sure..)
- **"an empty value means that all files in this image and any dependencies will be available in the rootfs."** => here again : what about a missing value ?

jonboulle:
> does the runtime (container executor) juste filter out the allowed paths from dependency acis, or will it raise an error if some dependency container contains files outside the allowed paths ?

It filters; what's in the pathWhitelist is all that ends up in the resolved layout

> "This field is only required if the app has dependencies and you wish to remove files from the rootfs before running the container" => Ok i now have my answer (files from dependency container but outside the folder path are just dropped) but will it raise an error if the field is missing ? (does "required" means "mandatory" here, not sure..)

No, "required" here means "necessary if you wish to accomplish this". The pathWhitelist field is always an _optional_ part of the manifest, but if you want to filter paths, you need to include it. Any ideas how we can subtly tweak this to make that clearer?

> "an empty value means that all files in this image and any dependencies will be available in the rootfs." => here again : what about a missing value ?

A missing value means the field is not defined, does that make sense?

frntn:
> No, "required" here means "necessary if you wish to accomplish this". The pathWhitelist field is always an optional part of the manifest, but if you want to filter paths, you need to include it. Any ideas how we can subtly tweak this to make that clearer?

Like in RFCs I would use keywords (https://www.ietf.org/rfc/rfc2119.txt) through the whole document.
In this example the sentence could be : 
```
The pathWhitelist is an OPTIONAL field. 
You MUST add it to filter in allowed paths from the app dependent rootfs to final container rootfs. 
An empty value MUST be considered as a missing field so no files from rootfs dependencies will be filtered out"
```

kelseyhightower:
@frntn I like the use of the RFC keywords.

kelseyhightower:
[...]

PR #122, opened 2015-01-16
spec: using standard RFC 2119 proposal (example on pathWhitelist -- #109)

frntn:
See #109


jonboulle:
I actually think what's on master for this is fine now so I'm going to close this out.

PR #123, opened 2015-01-16
tarheader: correctly set major/minor on /dev files

alban:
This fix would help with flanneld using /dev/net/tun.
See https://github.com/coreos/rocket/issues/389


vbatts:
Perhaps, we could just ahead the next step https://github.com/vbatts/spec/commit/6f1fab9fb0df0455d6238d691bf5f322abd400a8

alban:
@vbatts: good idea! Can you make a pull request with it?

I wonder what's the best way to report a missing capability CAP_MKNOD when running the unit test...
```
if ...:
  t.Skip("Test requiring CAP_MKNOD disabled.")
```

jonboulle:
:+1:, please file that PR :-)

PR #124, opened 2015-01-16
schema: don't discard existing data when unmarshaling complex structures

mpasternacki:
These changes make it easier to programmatically construct and modify manifests.

Would it be a good idea to take that further and allow "delta JSON" for slice types (Annotations, Labels, etc): JSON unmarshaled on top of existing slice would then:
- add new element if it doesn't exist
- update existing element if there is one
- delete existing element on some marker (nil value? empty string value? special attribute that wouldn't make it into final manifest?)

### DIFF
--- a/schema/container.go
+++ b/schema/container.go
@@ -7,6 +7,8 @@ import (
 	"github.com/appc/spec/schema/types"
 )
 
+const ContainerRuntimeManifestKind = types.ACKind("ContainerRuntimeManifest")
+
 type ContainerRuntimeManifest struct {
 	ACVersion   types.SemVer      `json:"acVersion"`
 	ACKind      types.ACKind      `json:"acKind"`
@@ -20,6 +22,8 @@ type ContainerRuntimeManifest struct {
 // containerRuntimeManifest is a model to facilitate extra validation during the
 // unmarshalling of the ContainerRuntimeManifest
 type containerRuntimeManifest ContainerRuntimeManifest
+
+var BlankContainerRuntimeManifest = ContainerRuntimeManifest{ACKind: ContainerRuntimeManifestKind}
 
 func (cm *ContainerRuntimeManifest) UnmarshalJSON(data []byte) error {
 	c := containerRuntimeManifest{}
@@ -42,14 +46,16 @@ func (cm ContainerRuntimeManifest) MarshalJSON() ([]byte, error) {
 	return json.Marshal(containerRuntimeManifest(cm))
 }
 
+var cmKindError = types.InvalidACKindError(ContainerRuntimeManifestKind)
+
 // assertValid performs extra assertions on an ContainerRuntimeManifest to
 // ensure that fields are set appropriately, etc. It is used exclusively when
 // marshalling and unmarshalling an ContainerRuntimeManifest. Most
 // field-specific validation is performed through the individual types being
 // marshalled; assertValid() should only deal with higher-level validation.
 func (cm *ContainerRuntimeManifest) assertValid() error {
-	if cm.ACKind != "ContainerRuntimeManifest" {
-		return types.ACKindError(`missing or bad ACKind (must be "ContainerRuntimeManifest")`)
+	if cm.ACKind != ContainerRuntimeManifestKind {
+		return cmKindError
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


mpasternacki:
Bump? Does it need any improvements to get merged?

philips:
lgtm. Any objections @jonboulle ?

jonboulle:
@mpasternacki sorry about the delay on this one. The central idea of the change makes total sense, but I am not really sure about the `Blank*Manifest` vars. Do you have an example use case? What about having a function to generate them instead?

```
func NewImageManifest() *ImageManifest {
    return &ImageManifest{
        ACKind: ContainerRuntimeManifestKind,
        ACVersion: version.Version,
    }
}
```

mpasternacki:
Regarding the `Blank*Manifest`, the reason for this approach was that:

 - it seemed inelegant to make a function that actually returns a constant value - OTOH, Go compiler should be able to optimize that out, and the pointer function would return a new pointer every time, which would prevent possibility of accidental overwrite by carelessly using `var myImg = &schema.BlankImageManifest`
 - It seemed wrong to write a function that would return an invalid manifest (i.e. without name), and I needed a "blank" manifest to fill in: I don't have a name yet when I start composing the new manifest, so parameterized `NewImageManifest(name ACName)` wouldn't be useful to me. As an API user, I'd expect `NewFoo()` function to return a valid value. Maybe this would be solved by using the function approach, but naming it `Blank*Manifest()` rather than `New*Manifest()` to indicate that the returned values are not complete?

schema/container.go
```diff
@@ -41,14 +45,16 @@ func (cm ContainerRuntimeManifest) MarshalJSON() ([]byte, error) {
 	return json.Marshal(containerRuntimeManifest(cm))
 }
 
+var cmKindError = types.InvalidACKindError(ContainerRuntimeManifestKind)
```
jonboulle: It's a bit overkill to split this declaration out imho

jonboulle:
Bah! GH lost my comment. One sec

jonboulle:
>  accidental overwrite by carelessly using var myImg = &schema.BlankImageManifest

Yeah, that is basically my concern with the var. Not that is a pattern unseen in the standard lib (e.g. `http.DefaultTransport`), but I'm still not fond of it.

> I'd expect NewFoo() function to return a valid value.

Fair point. I don't think it would be crazy to add more default parameters so that it _does_ return a valid value, though. Otherwise, `Blank*Manifest()` helpers seems OK

Getting back to one of your original questions:
> Would it be a good idea to take that further and allow "delta JSON"

I am a bit wary of this magic during unmarshalling. If people want this behaviour I would rather encourage it at the JSON level with a standard mechanism there like http://jsonpatch.com/

schema/container.go
mpasternacki: I have inserted it as to DRY (to have the `"ContainerRuntimeManifest"` / `"ImageManifest"` strings defined in one place, and to enforce consistent error message by wrapping the error in `types.InvalidACKindError()`), and intent for the variable was to avoid instantiating same thing many times, using a private variable to avoid the accidental overwrite risk. If you find it overengineered and prefer calling `types.InvalidACKindError()` directly from the validation method, I'm fine with changing that (actually, Go compiler may be smart enough to not only inline the call, but realize it is actually a constant value, so there may be no real difference).

mpasternacki:
As I think about it, the exported variable pattern in the standard lib seems to be used where it makes sense for user to modify the variable (I can imagine application that wants to just replace `http.DefaultTransport` to avoid qualifying every call). I'll rewrite the PR to use functions.

Parameterizing `NewFoo()` to enforce required parameters would make it useless for me, though this may be an edge case not worth explicitly supporting: I use a blank boilerplate ImageManifest in Jetpack's image builder code, so that the image can specify only relevant pieces. I treat ACType/ACVersion as boilerplate, since the manifest is constructed and re-serialized by Jetpack code, and Jetpack ensures validity and version - so it seems to make more sense to have Jetpack stamp the final manifest with type/version. Obviously, when I create the initial boilerplate, name is not known – this comes from JSON that I'm reading into the boilerplate.

A way to make everyone happy would be to parameterize `NewFoo()` with name etc, but accept `nil` or `""`, to let user explicitly declare they want a blank. Would that be a good approach?

mpasternacki:
Ah, re "delta JSON": agreed. If I want to "patch" parent image's manifest, I'll find an external way to do it.

jonboulle:
Fair points. I would say the `NewFoo()` constructors seem like overkill at this stage; let's err on the side of simplicity and leave them out until there is a clearer need. Happy to ship this with the `Blank*` functions.

mpasternacki:
Changed `Blank*Manifest` to be functions, rebased on top of current master.

jonboulle:
Thanks for coming back to this, I was planning to do it myself but fell of my plate. LGTM

Issue #127, opened 2015-01-17
spec: minimal set of os-specific device nodes


vcaputo:
For the apps to be able to have reasonable expectations of what to find at places like /dev and /proc, the spec should probably explicitly state what the execution environment must provide.

For linux we could just conform to http://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/, though there's quite a bit going on there, some of which probably isn't necessary in the app's root.  Perhaps we should instead require LSB conformance: http://refspecs.linuxfoundation.org/LSB_2.0.0/LSB-Core/LSB-Core/execenvfhs.html

What about the other operating systems?

Is it considered beyond the scope of the spec?

alban:
Yes I think the spec should specific that. There is another list of mount points and device nodes:
https://github.com/docker/libcontainer/blob/master/SPEC.md#filesystem

jonboulle:
Hijacking this issue!

Per out-of-band discussion: we reached a consensus that of the various device-file related possibilities this approach probably makes the most sense.

To restate: within the spec, we define for each OS a minimum set of device files that need to be provided by the executor (i.e. made available in an app's chroot filesystem). So on Linux, this might look like for example the LSB standard mentioned above.

@mpasternacki would you be able to help out on the BSD side for this?

mpasternacki:
On FreeBSD side (jails are FreeBSD-specific, not a generic BSD mechanism), I would need to check out the standard, but here is what I gathered from practice:

Side Note: for FreeBSD guests, it doesn't seem to make much sense *not* to have whole basesystem (base.txz) inside the container's chroot. Because of how the system is developed (userspace basesystem in lockstep with kernel), much of the software - including pieces of basesystem and stdlib - relies on basesystem being present: directory structure, binaries, files, etc. Trying to get things working without a basesystem is going to be difficult, and I wouldn't expect any of the system maintainers to help or support that. Given that there is no disk space gained by not having all images include the basesystem (because of ZFS copy-on-write clones), and that the basesystem can be considered secure (if there is a security issue in the basesystem, I have bigger problems outside the jails), I just go with what the full basesystem as the supported basis of FreeBSD guests.

[...]

Issue #99, opened 2015-01-08
Passn/dev/


stereoit:

jonboulle:
@stereoit I think this is something we would anticipate implementations providing, but I am not sure there is a way that makes sense to incorporate it into the spec. For example since the goal is for the spec to be OS agnostic, I'm not sure of a sufficiently generic way to describe this behaviour.

I suppose we could describe it as an optional/example extension that the spec explicitly doesn't dictate but that implementations might include, would that suffice for you?

stereoit:
I think this is perfectly fine (I know docker supports this) so as long as
it is at least as optional I am happy.

kelseyhightower:
@stereoit Would you like to take a shot at updating the spec to call out optional support for adding a host device to the container at runtime?

alban:
Different kind of devices:
* Run-time-specified devices. E.g. "/dev/sdc" would not make sense to be specified at build-time in the ACI. A GParted container could be executed ("rkt run --device /dev/sdc") to partition a specific disk. The major/minor numbers of devices can be dynamic (e.g. device mapper) so the device nodes cannot be in the ACI tar archive. I am not convinced we need to add something in the spec, we could just implement it in "rkt run".
* Build-time-specified basic devices ("/dev/null") specified in the "systemd container interface" document. The major/minor numbers for those basic devices are static on Linux so they could be in the ACI tar archive (usually in the /dev directory), or they could just be created by stage0 if the spec is updated to request them (#127), or the manifest in the ACI could optionally request stage0 to create them.
* Build-time-specified other devices. E.g. "/dev/net/tun" for coreos/rocket#389. They could just be in the ACI in some cases. It would work fine for "/dev/net/tun" because the major/minor numbers for tun are static on Linux. But there might cases where it does not work: /dev/v4l/by-id/some-camera might have a minor allocated dynamically. Maybe rootfs/dev/v4l/by-id/some-camera could exist in the tar archive with major/minor 0/0 and the manifest could request stage0 to look at the correct major/minor on the host when extracting the files...

jonboulle:
Is this covered by #165 ?

PR #128, opened 2015-01-19
schema: add "deviceFiles"

alban:
In order to add character device files or block device files in the ACI, one
could create them in the directory with mknod and let "actool build" pick them
up. But mknod requires root and it is desirable to be able to build the ACI
without being root.

This patch adds the specification for "deviceFiles":

{
    "acVersion": "0.1.1",
    "acKind": "ImageManifest",
    ...
    "deviceFiles": [
        "/dev/null",
        "/dev/zero",
        "/dev/net/tun"
    ]
}

Note that the device file type (character or block) and the major/minor are not
specified in the manifest: the device files will not be added in the ACI tar
archive by "actool build". Some device files use a dynamically allocated
major/minor so it might be different at build-time and at run-time.

stage1 must checks the device file type, major and minor of each file listed in
deviceFiles at run-time on the host and create them in the container root
filesystem.

appc/spec#99
appc/spec#127
coreos/rocket#389

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -483,6 +483,14 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest)
         "/etc/reduce-worker.conf",
         "/etc/systemd/system/"
     ],
+    "deviceFiles": [
+        "/dev/null",
+        "/dev/zero",
+        "/dev/full",
+        "/dev/random",
+        "/dev/urandom",
+        "/dev/net/tun",
+    ]
     "annotations": [
         {
             "name": "authors",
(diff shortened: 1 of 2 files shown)


markllama:
so if I read this correctly, https://github.com/coreos/rocket/issues/396 requires this change to the spec *and* an update to the actool validation code?

jonboulle:
@markllama yes, because it leverages the validation code to do its processing of manifests

@alban how would you propose dealing with operating systems that don't support listed devices?

vcaputo:
@alban and what is supposed to happen when the host doesn't possess an included device path?

Also, security implications of reproducing whatever device node an .aci mentioned within a container's filesystem?  It's not too dissimilar to simply bind-mounting in the host's /dev, when the .aci controls which nodes are pulled in.

I'm more in favor of the spec requiring a minimum set of device nodes be pulled in, on a per-os basis, as mentioned in https://github.com/appc/spec/issues/127.

alban:
@jonboulle if an OS does not support tun and the file /dev/net/tun is missing on the host (e.g. Linux compiled with CONFIG_TUN=n), then the executor will fail to stat the file and will not get the major/minor. In that case, it should not start the container and report an error to the user instead.

@vcaputo the .aci also controls the isolators and the manifest could include "capabilities/bounding-set"="CAP_MKNOD". Both "CAP_MKNOD" and "deviceFiles" introduce similar security implications. (alternatively, a device file could also be included in the tar archive: Rocket extracts device files without complaining at the moment). Do we have a way to warn the user against .aci files with such security implications? Executors could refuse to start an aci with unless the user explicitely allows it (something like "rkt run --allow-priviledged-aci"?).

I agree that a minimum set of device nodes should be pulled in (#127) regardless of the content of deviceFiles in the manifest. /dev/net/tun should not be pulled in automatically though, so a mechanism for adding more devices should exist.

As an alternative to deviceFiles, it was suggested to use bind-mounting on specific device files. The manifest could have:
```
        "mountPoints": [
            {
                "name": "dev-net-tun",
                "path": "/dev/net/tun",
                "readOnly": false
            }
        ]
```

And the command:
```
bin/rkt run -volume=dev-net-tun:/dev/net/tun /tmp/flannel-v0.2.0+git-linux-amd64.aci
```

The security implications of having a char device in the container would be more explicit on the command line, but CAP_NET_ADMIN would still be hidden from the command line. So I am not sure the more complex command line would be a gain.

Bind-mounting a device file does not work in Rocket at the moment because systemd-nspawn has code to explicitely refuse to create a mount point for block and char devices: systemd/src/nspawn/nspawn.c:
```
/* Create the mount point, but be conservative -- refuse to create block
 * and char devices. */
...
} else {
        log_error("Refusing to create mountpoint for file: %s", *x);
        return -ENOTSUP;
}

```

* If we choose the deviceFiles route, I can update the patch to explain what happens when the OS does not support the listed device.
* If we choose the bind-mounting route, I can patch systemd-nspawn to support that.

alban:
The security regarding read/write/mknod access to /dev files can be handled separately with an isolator based on the Device Whitelist Controller cgroup:
https://www.kernel.org/doc/Documentation/cgroups/devices.txt

jonboulle:
Closing this in favour of #127

PR #130, opened 2015-01-20
schema: Clarify kind, source, and readOnly for image manifest volumes.

cdaylward:
Made the follow modifications to the spec based on my extrapolation of what the **volumes** fields meant.

- Clarify what should occur based on kind.
- Add missing definition for source.
- Add missing definition for readOnly.


jonboulle:
This makes sense to me. @philips ack?

jonboulle:
Actually, I am just going to land this now since I think it is an accurate description of the situation today. We can follow up as necessary when @philips figures out the rest of the volume stuff..

PR #131, opened 2015-01-20
spec: fix isolator names

alban:


SPEC.md
```diff
@@ -208,8 +208,8 @@ Additional isolators will be added to this specification over time.
 |blockIO/writeBandwidth   |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
 |networkIO/readBandwidth  |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
 |networkIO/writeBandwidth |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|privateNetwork           |string|"&lt;true&#124;false&gt;"           |"true"                              |
-|capabilities/boundingSet |string|"&lt;cap&gt; &lt;cap&gt; ..."       |"CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"|
+|private-network          |string|"&lt;true&#124;false&gt;"           |"true"                              |
```
jonboulle: Good catch. Clearly isolators are an area that haven't received much love in a while since none of the names have been updated to the AC Name type! Would you mind changing the others too? (blockIO and networkIO)

vcaputo:
lgtm

philips:
This is fine. The whole thing needs to be fixed up though to make the isolators value objects instead of strings.

schema/types/isolator.go
```diff
@@ -1,6 +1,6 @@
 package types
 
 type Isolator struct {
-	Name ACName `json:"name"`
-	Val  string `json:"value"`
+	Name  string `json:"name"`
```
jonboulle: er, I just realised you are doing conflicting things in this review: changing the isolator names _to_ `ACName`s (which is what it is in the explanatory part of the spec) but then changing this type back to a string - why?

alban: Sorry I got confused with the meaning of ACName. I will fix this.

alban:
Patch updated:
- Fix other isolator names (blockIO and networkIO)
- Do not change ACName type to strings.

jonboulle:
Thanks a lot! LGTM

PR #132, opened 2015-01-20
discovery: Tie together ACI and Sig URLs.

sgotti:
This patch, should be merged together with the rocket part (coreos/rocket#398).

As the aci urls is tied with its signature url, to avoid confusion, tie them
together in their own struct.

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -16,10 +16,14 @@ type acMeta struct {
 	uri    string
 }
 
+type ACIEndpoint struct {
+	ACI string
+	Sig string
+}
+
 type Endpoints struct {
-	Sig  []string
-	ACI  []string
-	Keys []string
+	ACIEndpoints []ACIEndpoint
+	Keys         []string
 }
 
 const (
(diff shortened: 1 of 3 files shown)


discovery/discovery.go
```diff
@@ -125,20 +129,21 @@ func doDiscover(app App, pre string, insecure bool) (*Endpoints, error) {
 			// Ignore not handled variables as {ext} isn't already rendered.
 			uri, _ := renderTemplate(m.uri, tplVars...)
 			sig, ok := renderTemplate(uri, "{ext}", "sig")
-			if ok {
-				de.Sig = append(de.Sig, sig)
+			if !ok {
```
jonboulle: +1 on the idea of the `ACIEndpoint` struct, but I don't know about this change in behaviour, what about ACIs without signatures?

sgotti: @jonboulle The behaviour isn't changed:
* The spec says (or better, It's inferred and I think it should be clarified, I'm going to open a PR on this) that a template (simple or meta) should contain an {ext} token. So if {ext} is present it will be rendered to "aci" and "sig", if not present the rendering will fail. So there isn't a possibility that the ACI url is rendered but not the Sig url.

* Given the above, it's the ACE that will download or not the signature and verify it or not. Actually rocket does this:
  * if -insecure-skip-verify=false (default) it tries do download also the sig and verify it
  * if -insecure-skip-verify=true it doesn't try to download the sig.

sgotti: Trying to clarify the SPEC I opened #137.

jonboulle: Sorry, you are correct, I was misreading what this does. This LGTM for now and we can follow up with clarifications from #137. Thanks!!

PR #135, opened 2015-01-21
validator: make /opt/acvalidtor in aci as expected

eyakubovich:


jonboulle:
More context? Why is /opt expected?

jonboulle:
Oh, because someone actually made that the working directory. Carry on

jonboulle:
LGTM but nit, s#acvalidtor#acvalidator# in your commit before merging (also, adding context would be nice - "workingDirectory was set to this path but the directory was not created" or so)

jonboulle:
thanks

PR #136, opened 2015-01-21
Metadatasvc validation fixes

eyakubovich:


ace/validator.go
```diff
@@ -84,6 +81,8 @@ var (
 	}
 	// "Name"
 	an = "coreos.com/ace-validator-main"
+	// value of AC_METADATA_URL
+	metadataURL = ""
```
jonboulle: how about just passing this instead of using a global var?

schema/container.go
```diff
@@ -26,6 +26,9 @@ func (cm *ContainerRuntimeManifest) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
+	if c.Annotations == nil {
```
jonboulle: isn't this a no-op? http://play.golang.org/p/UiDVPZLgx-

vcaputo: it doesn't look like a noop to me, what am I missing?

jonboulle: I mean, effectively a no-op, not strictly identical, since nil slices are generally usable as zero-length slices are (appendable, iterable, etc).

Maybe I should rephrase: why exactly is it preferable to replace it explicitly with a zero-length slice here?

(Note also this makes it inconsistent with Volumes, Isolators)

eyakubovich: You're right, the value is questionable. My validation code was using reflect.DeepEquals and failing as it was comparing nil and empty. But I mistakenly thought there's also a problem with calling annotations.Set() when the underlying slice is nil -- there's not.

I can revert this and easily work around this in the validator code.

ace/validator.go
eyakubovich: It seemed like such a good idea late at night :) I'll clean up my laziness.

jonboulle:
LGTM, but please rebase/squash

eyakubovich:
I was actually trying to keep them as logically separate commits but I'll squash.

jonboulle:
@eyakubovich sorry I was really just referring to the "gofmt" one

PR #138, opened 2015-01-21
aci/actool: remove superfluous AddFile argument

jonboulle:
At some point in the past, AddFile relied on the path to be explicitly passed
in, but now it simply uses the path that is set on the tar header it is
passed, so we can remove the argument now.


philips:
lgtm

PR #139, opened 2015-01-22
s/app/manifest/ in tar line

keunwoo:
Commit 408d3374966450045b34fa69422384bfeda1a405 updated the image description (line 59) but not the tar line.


jonboulle:
ah! good find, thanks.

PR #141, opened 2015-01-23
schema: Modify App Environment to be a list of name/value objects.

cdaylward:
Make environment consistent with isolators and annotations (schema on "left", user values on "right", etc).

- Add EnvironmentVariable type.
- Add Environment alias for list of EnvironmentVariables
- Replace App.Environment field with new Environment type
- Plumb App.assertValid through to Environment.assertValid()
- Call out and test for IEEE Std 1003.1-2001 in environment variable
  identifiers.

### DIFF
--- a/schema/types/environment.go
+++ b/schema/types/environment.go
@@ -1,0 +1,96 @@
+package types
+
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+)
+
+var (
+	envPattern = regexp.MustCompile("^[A-Za-z_][A-Za-z_0-9]*$")
+)
+
+type Environment []EnvironmentVariable
+
+type environment Environment
+
+type EnvironmentVariable struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (ev EnvironmentVariable) assertValid() error {
+	if len(ev.Name) == 0 {
+		return fmt.Errorf(`environment variable name must not be empty`)
+	}
+	if !envPattern.MatchString(ev.Name) {
+		return fmt.Errorf(`environment variable does not have valid identifier %q`, ev.Name)
+	}
+	return nil
+}
+
+func (e Environment) assertValid() error {
+	seen := map[string]bool{}
+	for _, env := range e {
+		if err := env.assertValid(); err != nil {
+			return err
+		}
+		_, ok := seen[env.Name]
+		if ok {
+			return fmt.Errorf(`duplicate environment variable of name %q`, env.Name)
+		}
+		seen[env.Name] = true
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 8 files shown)


schema/types/environment.go
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf(`environment variable name must not be empty`)
+	}
+	if !envPattern.MatchString(ev.Name) {
+		return fmt.Errorf(`environment variable does not have valid identifier %q`, ev.Name)
+	}
+	return nil
+}
+
+func (e Environment) assertValid() error {
+	seen := map[string]string{}
```
jonboulle: is this the exact same pattern we use elsewhere? `map[string]bool` or `map[string]struct{}` is slightly more idiomatic to represent a set

jonboulle:
One question (and this certainly again raises the idea that we should be sharing code amongst these lists of things), but looks good to me

philips:
I am on the fence on this one. Environment variables on all operating systems are simple key/value pairs so I don't know if there is really a schema here that we plan to expand on. The one advantage is that we clearly see the ordering of which value should override which.

schema/types/environment.go
cdaylward: It's consistent with the way that Annotations are stored. Isolators as well (however this too is a bit inconsistent in the code, e.g. unaliased []Isolator). If they are all technically sets, then they should probably have consistent behavior when duplicates exist in the schema (error).

cdaylward:
Annotations and Isolators are also simple key/value pairs. Given that all three (Annotations, Isolators, and Environment) have the same model, they should probably have the same representation in the schema. It seems a bit like splitting hairs when we're talking about a map at the end of the day...but I think consistency should probably win. Having a consistent schema for maps also affords the application of constraints to be done in a similar fashion (each might have different behavior upon duplicates...e.g. map vs multimap, etc). Which to @jonboulle's point, means they could probably share some abstract base code.

schema/types/environment.go
jonboulle: sorry, more specifically I was talking about this uniqueness check within `assertValid`. For example [here](https://github.com/appc/spec/blob/master/schema/types/annotations.go#L18) we care about the values because we subsequently access them, whereas in the case of Environment we do not use the values at all.

jonboulle: (nbd)

philips:
@cdaylward Good argument. Lets do it. +1 from me.

schema/types/environment.go
cdaylward: Oh, I see what you're saying. No problem, I'll amend.

jonboulle:
[LGTM](http://www.lgtm.in/i/wHu3)

PR #147, opened 2015-01-26
spec: remove private-network isolator

eyakubovich:
Having private-network as an optional isolator is problematic:
1. No private-network makes identity for metadata svc very difficult
2. The network is shared between all apps, reconciling conflicting
isolator values in manifests is impossible.


vcaputo:
lgtm

philips:
lgtm, it has been suggested that "private" is too nuanced to be useful.

philips:
I would like to have a term for "only localhost to a pod" though.

jonboulle:
+100!

eyakubovich:
Can one of the collaborators please merge it.

PR #148, opened 2015-01-26
spec: sign/verify: use hmac-512 and separate content from sig

eyakubovich:
- Use HMAC-512 to be consistent with the rest of the spec.
- Have verify take in object that was signed so signature doesn't
have to store it (or its digest).


jonboulle:
Makes total sense to me; the client should keep track of the object, not the metadata server.

jonboulle:
@eyakubovich Can I just merge this? I am not really sure what part you thought might be controversial?

jonboulle:
LGTM

PR #151, opened 2015-01-28
spec: exec: add parameter examples

alban:
I thought that exec could contain several executables to run. But the
array is just here to contain parameters. Adding parameters in the
example will make it more obvious that "exec" can only contain *one*
executable with potential parameters.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -399,7 +399,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest)
     ],
     "app": {
         "exec": [
-            "/usr/bin/reduce-worker"
+            "/usr/bin/reduce-worker",
+            "--quiet"
         ],
         "user": "100",
         "group": "300",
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -399,20 +399,23 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest)
...
+                    "/usr/bin/data-downloader",
+                    "--enable-https"
                 ],
                 "name": "pre-start"
             },
             {
                 "exec": [
-                    "/usr/bin/deregister-worker"
+                    "/usr/bin/deregister-worker",
+                    "--verbose"
```
jonboulle: Bit of a nit, but how about mixing it up so at least one of them has no args and just the executable?

alban: Ok, I removed the --enable-https. So we can see that both app/exec and app/eventHandlers/exec can have parameters.

jonboulle: thanks!

jonboulle:
lgtm

PR #152, opened 2015-01-28
spec: chroot: more details

alban:


SPEC.md
```diff
@@ -162,6 +162,8 @@ This example container will use a set of three apps:
 | example.com/worker-backup          | 1.0.0   | sha512-3e86b59982e49066c5d813af1c2e2579cbf57... |
 | example.com/reduce-worker-register | 1.0.0   | sha512-86298e1fdb95ec9a45b5935504e26ec29b8fe... |
 
+The chroot means the app `reduce-worker` will not see the files from `worker-backup.aci` or `reduce-worker-register.aci`. If the three apps have a dependency on a common ancestror image, the files from that ancestror image will be visible by all apps, however a change performed by one app will not be shared to the other apps. Apps willing to share read-write files should set up volumes.
```
jonboulle: Hmm, frankly I think this makes things more confusing (nowhere else do we talk about "ancestor images", for example, and I think "unique read-write rootfs" should be explicit enough to capture the fact that they are independent?). Let me think on it a bit, I kind of see where you're coming from and we can probably add more detail on each of these pieces but I would like to approach it  a bit differently.

alban:
I wonder if the spec should avoid talking specifically about chroot and instead allow implementations to use an additional per-app mount namespace + pivot_root if they want to. This would have the benefit of:
* blocking a malicious/compromised app from accessing the files of another app in the same container by escaping the chroot with the chdir("..") method: http://linux-vserver.org/Secure_chroot_Barrier
* allowing a container to start a nested container with a new user namespace:
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=3151527ee007b73a0ebd296010f1c0454a919c7d

philips:
@alban I think this would be a good one to add to the ABI.md doc or maybe OS.md

cdaylward:
I believe all of this is conceptually covered in #207 now, is it not?

philips:
@cdaylward Yes, this either belongs in #207 or in the OS-specific spec in #221.

jonboulle:
Closing, let's revisit after those two land.

PR #156, opened 2015-01-30
spec: clarify ACI format and signing/encryption

jonboulle:
This patch attempts to be much more explicit and strict about what an ACI file
actually is and tighten up a lot of the language.
- Be clearer about ordering of encryption/signing
- Move towards RFC2119 wording ("MUST", "MAY", "SHOULD").
- Breaks out Image ID into its own section to reflect that it is a first
 class concept.
- Remove second person references

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -48,7 +48,7 @@ Now, let's dive into the pieces that took us from three URLs to a running contai
 ## App Container Image
 
 An *App Container Image* (ACI) contains all files and metadata needed to execute a given app.
-In some ways you can think of an ACI as equivalent to a static binary.
+In some ways an ACI can be thought of as equivalent to a static binary.
 
 ### Image Layout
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -64,50 +64,55 @@ It includes a *rootfs* with all of the files that will exist in the root of the
...
+The ACI file format ("image archive") aims for flexibility and relies on standard and common technologies: HTTP, gpg, tar and gzip.
+This set of formats makes it easy to build, host and secure a container using technologies that are widely available and battle-tested.
 
-Image archives MUST be a tar formatted file.
-The image may be optionally compressed with gzip, bzip2, or xz.
-After compression, images may also be encrypted with AES symmetric encryption.
+- Image archives MUST be a tar formatted file.
+- All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
+- Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
+- Image archives SHOULD be signed using PGP, in detached signature mode.
```
jonboulle: Note that the current implementation in Rocket expects _late signing_ - e.g. the etcd ACI is signed *after* compression. This is contrary to what was implicitly described here (in the example commands below), where the signature is taken of the canonical image format (uncompressed, unencrypted tar) - which I am now making explicit here.

I have been going back and forth on this a little bit since there are some arguments both ways.

Signing the final (compressed/encrypted ACI) means that signature validation can happen as early as possible after transport and before any other processing, and also means that a `.sig` sitting next to an `.aci` can always be expected to match, which is a pretty nice property. On the other hand, since we would generally advise implementations to store the canonical image, it means that signature validation is probably _only_ useful at transport time (otherwise implementations would need to store information about compression/encryption and then reconstruct it later, ugh). I am actually OK with this (limiting signature validation to a property of the retrieval process), since we always have the cryptographic image ID to verify contents anyway at any other time.

Signing the uncompressed tar means that it is super easy to reproduce and the same signature can be used across a variety of transport formats. However, then the filenames are decoupled and working with ACIs and signatures on the command line becomes a total pain since you always have to check what you are working with before being able to check the signature. (Possible solution to this would be to introduce different file extensions to represent the different forms). It also means greater computational expense of having to always decrypt/decompress before verifying signatures. (Arguably this last point is just a short-term overhead since we will eventually be exploring streaming formats anyway).

vbatts: when a compressed image is pulled, is the CAS blob sitting on the disk always the uncompress tar?

jonboulle: @vbatts in rocket, yes

vbatts: honestly, I feel the signed artifact should be the greatest common denominator. Despite the expense of (de)compressing, I feel the base tar gets the signature.

SPEC.md
```diff
@@ -64,50 +64,55 @@ It includes a *rootfs* with all of the files that will exist in the root of the
...
+- Image signatures MUST be named with the suffix `.sig`.
 
+There are two further transformations that may be applied to image archives for transport:
+- Image archives MAY be compressed with `gzip`, `bzip2`, or `xz`.
```
vbatts: and by signing the base tar, what is the harm in having `myimage.aci.xz`?

SPEC.md
mpasternacki: This would be *bad* security practice. Verifying signature is making sure that the input is trusted. Software for verifying signatures is designed to work with untrusted output; other software is usually not (at least can't be assumed to). Thus, signature check should happen *before any processing*, including decompression, to avoid processing untrusted output. Otherwise, it doesn't protect against, say, buffer overflows in decompression software.

I could make the same argument against checksumming uncompressed tar too. The way I see it, SHA-512 is an identifier and protection against non-malicious transfer errors, not an authentication mechanism, so this is less of an issue. Signatures are definitely an authentication mechanism, so they should be checked before attempting any hind of processing.

vbatts: agreed. I would like to have a way to preserve that the contents on disk are still validate by a known identity and signature. That is presently the lost step.

mpasternacki: Right: keeping the compressed file to be able to re-validate and make sure that no content was substituted after validation & decompression is also an important point.

SPEC.md
```diff
@@ -64,50 +64,55 @@ It includes a *rootfs* with all of the files that will exist in the root of the
...
+- Image archives MAY be encrypted with AES symmetric encryption, after (optional) compression.
+
+The following example demonstrates the creation of a simple ACI using common command-line tools.
+In this case, the ACI is both compressed and encrypted.
 ```
 tar cvf reduce-worker.tar manifest rootfs
 gpg --output reduce-worker.sig --detach-sig reduce-worker.tar
```
vbatts: @jonboulle are the sig's ever armored? This way, additional header information could be stored in the sig, without affecting the validation of payload and signer identity.

SPEC.md
philips: @vbatts It makes discovery of the URLs harder and doesn't add much value since we can detect the compression via header snooping.

SPEC.md
philips: This should be armored I think. Also, this reminds me we should make it aci.sig

SPEC.md
philips: @mpasternacki Hrm, I don't agree that we should sign the compressed artifact. I see compression as part of the transport, just like https is part of the transport. I understand the concern that someone might attempt to exploit the transport layer but we are already making an https call too.

The downside of signing the compressed artifact is that the user needs to know the original transport compression method in order to be able to reconstruct the artifact and do a signature validation. And, currently, there is no place in the signature to hold that metadata.

philips: To clarify what I am saying is that we either:

- Use a signature format where we can know the type of compression used on the image (JOSE?)
- Do not sign the compressed image

SPEC.md
kelseyhightower: @vbatts The sig is armored for that very reason. Needs to be easy to add header info.

kelseyhightower: Based on the RFC we should be able to use custom headers for this: https://tools.ietf.org/html/rfc4880#section-6.2

```

```

SPEC.md
philips: After thinking through this more I think we should do late signing of the image. The downside is that we will need to store the type of compression used if we wish to check the signature but it is the right thing to do from a security perspective. In the next revision of the spec we can use something like JOSE so that the signature includes the type of compression used.

SPEC.md
```diff
@@ -64,50 +64,55 @@ It includes a *rootfs* with all of the files that will exist in the root of the
...
+This set of formats makes it easy to build, host and secure a container using technologies that are widely available and battle-tested.
 
-Image archives MUST be a tar formatted file.
-The image may be optionally compressed with gzip, bzip2, or xz.
-After compression, images may also be encrypted with AES symmetric encryption.
+- Image archives MUST be a tar formatted file.
+- All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
+- Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
+- Image archives SHOULD be signed using PGP, in detached signature mode.
+- Image signatures MUST be named with the suffix `.sig`.
```
philips: Lets make the `.aci.asc` and say they are armored PGP signatures.

SPEC.md
vbatts: late signing of the _compressed_ image? ...

Uncompressed:
```
vbatts@valse ~ (master *) $ sha256sum ./true.tar
5b921ba77402ea4c618d4e9bafb2031296fde1e3b997ddf274142c51323328bd  ./true.tar
vbatts@valse ~ (master *) $ sha256sum ./true.tar
5b921ba77402ea4c618d4e9bafb2031296fde1e3b997ddf274142c51323328bd  ./true.tar
```

After Compression:
```
vbatts@valse ~ (master *) $ cat true.tar | gzip -9 | sha256sum 
2c5e64edaa32da17d3f0ce8d3a986367fb40c2a6c0f9e230d71f07f62c63ed17  -
vbatts@valse ~ (master *) $ cat true.tar | gzip -9 | sha256sum 
aa244153f70eb6dfb10d012068b7d2499bd7b21e8ba523f242cfb98ddca00522  -
vbatts@valse ~ (master *) $ cat true.tar | gzip -9 | sha256sum 
accd6abd99ff3960aadb13cf83e0a36ee608d3060f5996ba0c8edae8405bcb4f  -
```

if we sign the compressed archive, it'll will largely be a one time sig validation.

SPEC.md
vbatts: right. I am game for armored. Though the example didn't show this. That's all. :-)

SPEC.md
vbatts: fair.

SPEC.md
vbatts: understood. Was just thinking through was to disambiguate the compression and what is being signed.
If i had `myimage.aci.gz` and `myimage.aci.sig` I would expect to validate the sig against the gunzip'ed `myimage.aci`
But i understand the difficulty in discovery.

SPEC.md
philips: @vbatts `gzip -n`

vbatts: <3 TIL

philips:
I think everyone is roughly in agreement here on the crypto and naming, I am going to fix this up since Jon is currently traveling and merge.

SPEC.md
philips: @vbatts The deterministic build stuff from Debian has lots of neat technical details https://wiki.debian.org/ReproducibleBuilds/Howto

cdaylward:
+1 SGTM

vbatts:
SGTM

philips:
Merged via 55f134d2ac05eb703091c2581548ab7960a772c6. Closing.

PR #155, opened 2015-01-30
SPEC: Redundant "http" scheme in metadata URLs

jonboulle:
https://github.com/appc/spec/blob/master/SPEC.md#metadata-server list examples for URL endpoints like "Retrievable at http://$AC_METADATA_URL/acMetadata/v1/container". The default for "AC_METADATA_URL" is listed as "http://169.254.169.255" (already includes a scheme).

This is at least confusing, but probably just wrong.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -322,7 +322,7 @@ Clients querying any of these endpoints must specify the `Metadata-Flavor: AppCo
 
 Information about the container that this app is executing in.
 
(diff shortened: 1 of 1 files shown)


jonboulle:
You're right, thanks for pointing that out. Patch posted.

jonboulle:
You're right, thanks for pointing that out. Patch posted.

philips:
lgtm

PR #143, opened 2015-01-23
SPEC: add mountPoints and remove fulfills

philips:
Add the concept of a mountPoint to the CRM and remove the idea that a volume
"fulfills" app mountPoints. The fulfills concept we troublesome because it was
putting the binding of a volume to a name at image/definition time instead of
runtime.

On the command line you could imagine it working something like this:

```
ace run
 example.com/webapp
 --volume ssl-share,type=empty
 example.com/ssl-key-downloader
 --mount ssl-share=example.com/ssl-key-downloader:ssl-out
 example.com/nginx-proxy
 --mount ssl-share=example.com/nginx-proxy:private-certs
```

This fixes #13

PR #158, opened 2015-01-30
schema/types: update volume docstrings

jonboulle:
type -> kind


jonboulle:
Minor cleanup following #143

PR #161, opened 2015-02-02
spec: be more specific about JSON

jonboulle:
@cdaylward thoughts? I am not really sure about 4627 vs.
[RFC7159](https://tools.ietf.org/html/rfc7159)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -99,7 +99,7 @@ An example application container image builder is [actool](https://github.com/ap
 
 ### Image Manifest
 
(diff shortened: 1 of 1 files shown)


cdaylward:
The existing reference implementation uses 4627 due to its go-ness. I don't see an issue with choosing RFC 4627. LGTM.

jonboulle:
OK, we can revisit if it emerges that we should move to RFC7159 for some reason.

PR #164, opened 2015-02-03
tarheader: use minor and major macros

vbatts:
on posix system's, include the minor and major device number, and use the sysmacros

/cc @jonboulle @alban

### DIFF
--- a/pkg/tarheader/pop_posix_test.go
+++ b/pkg/tarheader/pop_posix_test.go
@@ -1,0 +1,61 @@
+package tarheader
+
+import (
+	"archive/tar"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+)
+
+// mknod requires privilege ...
+func TestHeaderUnixDev(t *testing.T) {
+	hExpect := tar.Header{
+		Name:     "./dev/test0",
+		Size:     0,
+		Typeflag: tar.TypeBlock,
+		Devminor: 5,
+		Devmajor: 233,
+	}
+	// make our test block device
+	var path string
+	{
+		var err error
+		path, err = ioutil.TempDir("", "tarheader-test-")
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer os.RemoveAll(path)
+		if err := os.Mkdir(filepath.Join(path, "dev"), os.FileMode(0755)); err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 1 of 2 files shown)


sgotti:
@jonboulle @vbatts I added a related comment here: https://github.com/coreos/rocket/pull/276#issuecomment-72630500

vbatts:
also, as far as cgo is concerned with this, the segment of code is under the `_posix.go` build tag, so if the `tarheader` library were built for/on windows, it wouldn't be affected anyways.

sgotti:
@vbatts I was just curious to know the advantages between the pure go implementation like in 2f87893 or the use of external macros.

It the external macros path is used, for consistency, also https://github.com/coreos/rocket/blob/master/pkg/tar/tar.go#L184-L187 should be updated to use the ```makedev``` macro?

vbatts:
Advantages are usually that you are relieved of ensuring you keep the implementations identical. Granted, this bit is not likely to change.

The other can get updated as well. Should be a different PR, as it would involve a little rearranging for posix only logic.

sgotti:
LGTM!

pkg/tarheader/pop_posix_test.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"archive/tar"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+)
+
+// mknod requires privilege ...
```
jonboulle: how about adding a check and skipping this test?

vbatts:
@jonboulle PTAL

Now,
```
vbatts@valse ~/src/appc/spec/pkg/tarheader (vbatts-tar-dev-files) $ go test -v
=== RUN TestHeaderUnixDev
--- SKIP: TestHeaderUnixDev (0.00 seconds)
        pop_posix_test.go:37: no permission to CAP_MKNOD
PASS
ok      _/home/vbatts/src/appc/spec/pkg/tarheader       0.002s
vbatts@valse ~/src/appc/spec/pkg/tarheader (vbatts-tar-dev-files) $ sudo go test -v
=== RUN TestHeaderUnixDev
--- PASS: TestHeaderUnixDev (0.00 seconds)
PASS
ok      _/home/vbatts/src/appc/spec/pkg/tarheader       0.004s
```

jonboulle:
thank you! LGTM

PR #169, opened 2015-02-04
actool: split out BuildWalker into aci package

jonboulle:
This facilitates spec vendoring and also things like jonboulle/goaci reusing this code.

### DIFF
--- a/aci/build.go
+++ b/aci/build.go
@@ -9,9 +9,9 @@ import (
 	"github.com/appc/spec/pkg/tarheader"
 )
 
-// BuildWalker creates a filepath.WalkFunc that walks over
-// the given root and adds files to the given
-// ArchiveWriter
+// BuildWalker creates a filepath.WalkFunc that walks over the given root
+// (which should represent an ACI layout on disk) and adds the files in the
+// rootfs/ subdirectory to the given ArchiveWriter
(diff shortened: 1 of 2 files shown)


aci/build.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/appc/spec/pkg/tarheader"
+)
+
+// BuildWalker creates a filepath.WalkFunc that walks over
+// the given root and adds files to the given
+// ArchiveWriter
+func BuildWalker(root string, aw ArchiveWriter) filepath.WalkFunc {
```
philips: would the common case use case for this be appending a "root/" to the path?

philips:
lgtm, after the question about API.

aci/build.go
jonboulle: I am not really sure what you mean? The API now is a little weird (see line 31) since we expect a top-level ACI layout including the manifest rather than being pointed directly at the `rootfs/`. This is basically just because it was the compromise I reached after trying to keep `actool build` simple in terms of having to specify manifest paths, rootfs paths, etc (and keep it clear to users how an ACI is actually laid out on disk).

I am going to merge as-is for now and we can come back to it, I don't expect too many consumers of this out the gate :-)

PR #170, opened 2015-02-04
aci: remove dead code

jonboulle:
This is not useful, we can introduce as necessary if we decide to implement
some of this behaviour within the spec tooling (for now it's being implemented
by e.g. Rocket directly)


philips:
lgtm

PR #173, opened 2015-02-06
discovery: fix walk for keys and add debugging

### LINKED ISSUE #166: discovery: don't require ACI endpoints when discovery pubkeys
Discovery currently returns an error (and hence fails completely) in all cases in which it cannot find ACI endpoints:  https://github.com/appc/spec/blob/2fe56d3eeef90f9655e53bbb62ec00d9419a8c27/discovery/discovery.go#L146

This needs to be refactored a bit, since the "discover trusted key" use case does not work with this logic (it's quite possible for there to be no ACI corresponding exactly to the key being trusted, e.g. on coreos.com)

philips:
Add two new functions to let a user control the Discover walking process and
then also add attempts list to keep track of errors that were encountered along
the way.

Fixes #166

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"regexp"
@@ -26,12 +27,18 @@ type Endpoints struct {
 	Keys         []string
 }
 
+func (e *Endpoints) Append(ep Endpoints) {
+	e.ACIEndpoints = append(e.ACIEndpoints, ep.ACIEndpoints...)
+	e.Keys = append(e.Keys, ep.Keys...)
+}
+
 const (
 	defaultVersion = "latest"
 )
 
 var (
 	templateExpression = regexp.MustCompile(`{.*?}`)
+	errEnough          = errors.New("enough discovery information found")
 )
 
 func appendMeta(meta []acMeta, attrs []html.Attribute) []acMeta {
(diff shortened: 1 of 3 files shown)


discovery/discovery.go
```diff
@@ -143,26 +143,72 @@ func doDiscover(app App, pre string, insecure bool) (*Endpoints, error) {
...
+// and user feedback.
+type FailedAttempt struct {
+	Prefix string
+	Error  error
+}
+
+// DiscoverEndpoints will make HTTPS requests to find the ac-discovery meta
+// tags and optionally will use HTTP if insecure is set. It will not give up
+// until it has exhausted the path or found an image discovery.
+func DiscoverEndpoints(app App, insecure bool) (eps *Endpoints, attempts []FailedAttempt, err error) {
```
eyakubovich: Wouldn't this cause the whole thing to abort if the first iteration was not fruitful?

discovery/discovery.go
```diff
@@ -143,26 +143,72 @@ func doDiscover(app App, pre string, insecure bool) (*Endpoints, error) {
...
+	dfn := func(pre string, eps *Endpoints, err error) error {
+		if err != nil {
+			attempts = append(attempts, FailedAttempt{pre, err})
+			return nil
+		}
+		if len(eps.ACIEndpoints) == 0 {
+			return fmt.Errorf("found no ACI meta tags")
```
eyakubovich: Wouldn't this cause the whole thing to abort if the first iteration was not fruitful?

philips: Yea, this is how the old code worked too. So, it would assume that the first 200 website it got should have meta info.

jonboulle:
This could do with some cleaning up, but lgtm for now.

PR #176, opened 2015-02-06
spec: remove mention of metadata svc on 169.254.169.255

eyakubovich:
Since the metadata service address is specified in
AC_METADATA_URL env var, there's little value in
setting expectation that the service will be on a
particular IP. From the impl point of view, it may
be much easier to not have it listen on a particular
link-local address.


philips:
This makes sense to me. If we are having an environment variable then there is no point in having a convention.

jonboulle:
+1

PR #179, opened 2015-02-09
Iaguis/sanitize acname

iaguis:
With this PR users can sanitize an ACName string so they can be sure it's a valid one.

### DIFF
--- a/schema/types/acname.go
+++ b/schema/types/acname.go
@@ -2,12 +2,19 @@ package types
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 )
 
 const (
 	valchars = `abcdefghijklmnopqrstuvwxyz0123456789.-/`
+)
+
+var (
+	invalidChars = regexp.MustCompile("[^a-z0-9./-]")
+	invalidEdges = regexp.MustCompile("(^[./-]+)|([./-]+$)")
 )
 
 // ACName (an App-Container Name) is a format used by keys in different
@@ -71,4 +78,23 @@ func (n *ACName) UnmarshalJSON(data []byte) error {
 // MarshalJSON implements the json.Marshaler interface
 func (n *ACName) MarshalJSON() ([]byte, error) {
 	return json.Marshal(n.String())
+}
+
+// SanitizeACName replaces every invalid ACName character in s with a dash
+// making it a legal ACName string. If the character is an upper case letter it
+// replaces it with its lower case. It also removes illegal edge characters
+// (hyphens, periods and slashes).
+//
+// This is a helper function and its algorithm is not part of the spec. It
+// should not be called without the user explicitly asking for a suggestion.
+func SanitizeACName(s string) (string, error) {
+	s = strings.ToLower(s)
+	s = invalidChars.ReplaceAllString(s, "-")
+	s = invalidEdges.ReplaceAllString(s, "")
+
+	if s == "" {
+		return "", errors.New("must contain at least one valid character")
+	}
+
+	return s, nil
 }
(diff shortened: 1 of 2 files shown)


schema/types/acname_test.go
```diff
@@ -40,3 +40,21 @@ func TestNewACNameBad(t *testing.T) {
 		}
 	}
 }
+
+func TestSanitizeACName(t *testing.T) {
+	tests := []string{
+		"foo#",
```
jonboulle: Please make this a map of input -> expected output

iaguis: Right, it makes sense and uncovered a bug :)

schema/types/acname_test.go
```diff
@@ -40,3 +40,21 @@ func TestNewACNameBad(t *testing.T) {
 		}
 	}
 }
+
+func TestSanitizeACName(t *testing.T) {
+	tests := map[string]string{
+		"foo#":              "foo-",
```
jonboulle: Forgot to ask: can you please add one test-case of a name that _shouldn't_ be modified?

schema/types/acname.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 const (
-	valchars = `abcdefghijklmnopqrstuvwxyz0123456789.-/`
+	ACNameValChars = `abcdefghijklmnopqrstuvwxyz0123456789.-/`
```
jonboulle: also a comment for this now that it's exported

jonboulle: docstring I mean, not comment
(and the docstring for ACName should reference this)

philips:
@iaguis Who is the intended caller of Sanitize? I would rather we defensively reject invalid ACNames instead of having a thing that randomly casts invalid input to something valid.

iaguis:
We do reject invalid ACNames, the user needs to call `SanitizeACName` explicitly to make the input valid. It seemed like a reasonable feature for cases like appc/docker2aci#16

jonboulle:
Right, the idea is this is just a helper function for people working with the schema library. We would absolutely never do this implicitly. I suggested that iaguis move it here instead of docker2aci since it seemed of more general utility.

schema/types/acname_test.go
```diff
@@ -40,3 +40,22 @@ func TestNewACNameBad(t *testing.T) {
 		}
 	}
 }
+
+func TestSanitizeACName(t *testing.T) {
+	tests := map[string]string{
+		"foo#":                                             "foo-",
```
cdaylward: I have a conflicting proposed change in #185, namely that a valid ACName should not begin or end with namespace or separation characters (the last three characters in ACNameValChars). Could this be modified to check for and handle those cases?

iaguis: I replaced the implementation of Sanitize, it should not conflict with #185

cdaylward:
LGTM based on test results. @jonboulle?

philips:
@jonboulle we should add a comment saying this is a helper and should never be done without the user explicitly asking for a suggestion and that the downvert algo is not part of the spec.

schema/types/acname.go
```diff
@@ -72,3 +74,22 @@ func (n *ACName) UnmarshalJSON(data []byte) error {
...
 	return json.Marshal(n.String())
 }
+
+// SanitizeACName replaces every invalid ACName character in s with a dash
+// making it a legal ACName string. If the character is an upper case letter it
+// replaces it with its lower case. It also removes illegal edge characters
+// (hyphens, periods and slashes).
+func SanitizeACName(s string) (string, error) {
+	s = strings.ToLower(s)
+	invalidChars := regexp.MustCompile("[^a-z0-9./-]")
```
jonboulle: Please initialise these globally vars so that MustCompile would capture them at runtime initialisation

iaguis:
Initialised invalidChars and invalidEdges and added the suggested comment.

jonboulle:
LGTM anticipating #185, thanks.

PR #180, opened 2015-02-09
SPEC: move to .aci.asc file extension

philips:
Use the .aci.asc file extension since we are expecting an ASCII armored
extension of the aci file.

Discussed on: https://github.com/appc/spec/pull/156

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -70,7 +70,7 @@ This set of formats makes it easy to build, host and secure a container using te
 - Image archives MUST be a tar formatted file.
 - All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
 - Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
-- Image archives SHOULD be signed using PGP, in detached signature mode.
+- Image archives SHOULD be signed using PGP, the format MUST be ascii-armored detached signature mode.
 - Image signatures MUST be named with the suffix `.aci.asc`.
 
 There are two further transformations that may be applied to image archives for transport:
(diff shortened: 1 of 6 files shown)


SPEC.md
```diff
@@ -280,7 +280,7 @@ where _appname_, _version_, _os_, and _arch_ are set to their respective values
 In our example above this would be:
```
jonboulle: Line 278 needs update (GH won't let me comment on it directly)
bad grep? ;-)

discovery/discovery.go
```diff
@@ -19,7 +19,7 @@ type acMeta struct {
 
 type ACIEndpoint struct {
 	ACI string
-	Sig string
+	ASC string
```
jonboulle: nit, can you call this `Asc`, `ASC` makes me think it is a new initialism

philips: I figured Asc was a shortening of ASCII which is an initialism :-/

SPEC.md
philips: Fixed. Hard to grep for just `sig` without a dot in front.

philips:
@cdaylward Mind giving a review?

SPEC.md
```diff
@@ -82,7 +82,7 @@ In this case, the ACI is both compressed and encrypted.
 
 ```
 tar cvf reduce-worker.tar manifest rootfs
-gpg --output reduce-worker.sig --detach-sig reduce-worker.tar
+gpg --output reduce-worker.aci.asc --detach-sig reduce-worker.tar
```
cdaylward: Change to: ```gpg --armor --output ```... so that this command always outputs an ascii-armored signature (otherwise this might be relying on specific version/configuration of gpg). Call out that it should (must?) be ascii-armored on line 73?

> - Image archives SHOULD be signed using PGP, in detached signature mode.

cdaylward:
+1 on the change in general. It makes it more clear and consistent.

philips:
@jonboulle this is good to go if you are OK with ASC vs Asc.

philips:
@cdaylward thanks :)

discovery/discovery.go
jonboulle: an abbrevinitialism?!?
in that case can't we just use ASCII :-)

jonboulle:
I would still prefer ASCII or Asc over ASC, but if I am in the minority, go for it.

Issue #146, opened 2015-01-26
SPEC: propose erroring on overlapping files in tar


philips:
Tar has this pretty sad property that dentries can appear multiple times in the same file. I was thinking it would be rather pragmatic to just say this is always an error and advise executors to reject such files.

Thoughts?

cdaylward:
It's a great property when tar is used to actually store on tape! :)
+1 on this. My approach would be to reject any ACI that does not _strictly_ follow the format (manifest must be a regular file, rootfs must be a directory, error on any extraneous top-level files).

jonboulle:
+1, we have been dancing a little around the fact that the tar stuff is slightly underspecified, this is as good a place to start as any for being stricter/clearer.

jonboulle:
Fixed by #186, thanks.

PR #186, opened 2015-02-10
schema: Restrict duplicate/extraneous files in ACI tar layout.

cdaylward:
- The image archive must not have duplicate entries (possible with tar)
- The image archive must not have extraneous top-level pathnames. Only
  manifest and rootfs are valid.

Relates to issue #146. Too much? Too little?

### DIFF
--- a/aci/layout.go
+++ b/aci/layout.go
@@ -90,7 +90,7 @@ func ValidateLayout(dir string) error {
 // Application Container Image format.  If any errors are encountered during
 // the validation, it will abort and return the first one.
 func ValidateArchive(tr *tar.Reader) error {
-	var flist []string
+	var fseen map[string]bool = make(map[string]bool)
 	var imOK, rfsOK bool
 	var im bytes.Buffer
 Tar:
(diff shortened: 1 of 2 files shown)


jonboulle:
@cdaylward can you update `actool validate` to deal with this?

cdaylward:
@jonboulle Is this some scheme to get me to learn Golang?

cdaylward:
Updated `actool validate` with basic duplicate checking in ValidateArchive.

aci/layout.go
```diff
@@ -118,9 +118,16 @@ Tar:
 			}
 			rfsOK = true
 		default:
-			flist = append(flist, name)
+			if _, seen := fseen[name]; seen {
+				return fmt.Errorf("Duplicate file entry in archive: %s", name)
```
jonboulle: Errors should start lowercase, please

aci/layout.go
```diff
@@ -118,9 +118,16 @@ Tar:
...
+			}
+			fseen[name] = true
 		}
 	}
+	var flist []string
```
jonboulle: random sidenote: it bothers me that there isn't a better way to preallocate here, unless I'm just too dumb to think of it

jonboulle:
cool, lgtm, thanks!
(I wonder what existing ACIs are going to break ;-)

aci/layout.go
cdaylward: Seems ok. You have to pay the cost of decompression as well as reading all of the tar entry headers to know how much you need to allocate. The Golang append complexity is probably well amortized by archive read overhead in a single pass. *waves hands*

jonboulle: yeah, waves hands.

PR #187, opened 2015-02-10
Pass TERM variable to application's environment if appropriate

mpasternacki:
Currently, `$TERM` is always unset, which is problematic when one attempts to work interactively within a container. Since runtime is responsible for providing stdin/stdout/stderr to the app, it should also set `$TERM` if application runs on a terminal device.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -195,6 +195,7 @@ Note that logging mechanisms other than stdout and stderr are not required by th
 * **USER, LOGNAME** username of the user executing this app
 * **HOME** home directory of the user
 * **SHELL** login shell of the user
+* **TERM** terminal type if app is started on a tty/pty device; if application inherits stdio from host and it is a terminal, `$TERM` is copied from host environment.
 * **AC_APP_NAME** name of the application, as defined in the image manifest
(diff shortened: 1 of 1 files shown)


mpasternacki:
bump?

philips:
Err, /cc @vcaputo

vcaputo:
Seems reasonable to me, as long as it can be unset for non-interactive containers.

Things quickly get muddy in the context of pods though, are we placing a requirement on the ACE to provide a pty per application for interactive containers having multiple apps? (think screen/tmux)  They can't all share a pty if they do termio, and caring about the TERM variable implies termio is happening.

philips:
@mpasternacki Yes, can you give an example of what you would like to do? Do you want a container with just a single process and attach it to the current TERM?

philips:
Ping.

vbatts:
detecting tty/pty is not impossible, but can be tricky. Though, `$TERM` is something I would expect to pass to a container...

vbatts:
Is seems almost more sane to just have a default `$TERM` like screen(1) does, and allow users to override it.

PR #189, opened 2015-02-10
aci: clean the manifest path

vbatts:
the packed *.aci, may have path prefixes of "./", therefore the explicit
path of "manifest" would be missed.

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/aci/file.go
+++ b/aci/file.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"net/http"
 	"os/exec"
+	"path"
 
 	"github.com/appc/spec/schema"
 )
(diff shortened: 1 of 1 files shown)


philips:
Weird, which tar impls that do this? lgtm though.

aci/file.go
```diff
@@ -123,7 +124,7 @@ func ManifestFromImage(rs io.ReadSeeker) (*schema.ImageManifest, error) {
 		if err != nil {
 			return nil, err
 		}
-		if hdr.Name == "manifest" {
+		if path.Clean(hdr.Name) == "manifest" {
```
jonboulle: filepath.Clean?
also, could you please actually change "manifest" to ManifestFile?
https://github.com/appc/spec/blob/master/aci/layout.go#L106

vbatts:
@jonboulle `path` has a simpler offering and `filepath` overlaps it with additional functionality.
http://golang.org/pkg/path/#Clean and http://golang.org/pkg/path/filepath/#Clean

Also, updated to use the variable. PTAL.

jonboulle:
@vbatts I should have been more explicit abut what I meant: wouldn't it be better to use filepath.Clean to remain os.PathSeparator-agnostic?

jonboulle:
fixed by #212

sgotti:
@vbatts @jonboulle sorry! Didn't saw this issue!

PR #190, opened 2015-02-10
Support ports in the ACName

vbatts:
This is a subtlety of the ACName, which might require some discussion.

I realize the port is not part of the DNS RFC, but for the purposes of the name of an image, it is more like an URL. In this way it can still be accessed the same way and use the same ac-discovery, But would need to support having the port number present.

This adds a couple of additional parse tests and modifications to support port on the image name.

### DIFF
--- a/discovery/parse.go
+++ b/discovery/parse.go
@@ -32,13 +32,20 @@ func NewApp(name string, labels map[string]string) (*App, error) {
 // Example app parameters:
 // 	example.com/reduce-worker:1.0.0
 // 	example.com/reduce-worker,channel=alpha,label=value
+// 	example.com:8080/reduce-worker:1.0.0
+// 	example.com:8080/reduce-worker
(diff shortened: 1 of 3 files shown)


philips:
@vbatts Yea, I have been struggling with this because I really do not want the name of a container to imply a transport. But, discovery is a bit of a special case since we append a transport onto it. The best suggestion I have seen is letting a user specify a discovery port.

Related to #110

philips:
@vbatts Besides container naming for discovery where else would a ACName need a port?

vbatts:
discovery is all, really.

jonboulle:
I would rather avoid this since as mentioned it takes us back to the name implying transport. @vbatts what do you think of #110 as an alternative?

philips:
I am going to close this out since it is a duplicate discussion to #110.

Issue #54, opened 2014-12-18
spec: updating isolators


jonboulle:
The spec should describe how executors are able to update isolators dynamically during the lifetime of a container.

tnachen:
Also I think the spec should include how isolators can be pluggable, where it can be loaded and described dependencies of different isolators.

philips:
@tnachen Can you take a look at #192. I don't understand what you mean by pluggable.

philips:
I have no idea how we do this without adding a runtime API of some sort.

jonboulle:
I am pretty sure I jotted down this issue because you suggested it so I figured you had something in mind :stuck_out_tongue:

jonboulle:
At the very least we should just be vague about the _means_ but explicit about the _consequences_ - e.g. that the isolators exposed into the container via the metadata service may no longer be accurate.

jonboulle:
related: #276

PR #192, opened 2015-02-11
SPEC: make isolators objects instead of strings

philips:
Until now isolators have not had a schema and had to be serialized to a
string. Fix this by making an isolator an object with only a single
required field: "name". Everything else goes into isolator specific
fields.

Credits and influences:

http://www.freedesktop.org/software/systemd/man/systemd.exec.html#CapabilityBoundingSet=
https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/resources.md
https://github.com/google/lmctfy/blob/master/include/lmctfy.proto
https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt
https://www.kernel.org/doc/Documentation/cgroups/net_cls.txt
https://www.kernel.org/doc/Documentation/cgroups/blkio-controller.txt

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -201,25 +201,154 @@ Note that logging mechanisms other than stdout and stderr are not required by th
 ### Isolators
 
 Isolators enforce resource constraints rather than namespacing.
-Isolators may be applied to individual applications, to whole containers, or to both.
+Isolators may be scoped to individual applications, to whole containers, or to both.
 Some well known isolators can be verified by the specification.
 Additional isolators will be added to this specification over time.
 
-|Name|Type|Schema|Example|
-|---------------------------|------|------------------------------------|------------------------------------|
-|cpu/shares                 |string|"&lt;uint&gt;"                      |"4096"                              |
-|memory/limit               |string|"&lt;bytes&gt;"                     |"1G", "5T", "4K"                    |
-|cpu/mask                   |string|"&lt;cpu range&gt;"                 |"0", "0-3", "0,2,4-7"               |
-|block-io/read-bandwidth    |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
-|block-io/write-bandwidth   |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
-|network-io/read-bandwidth  |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|network-io/write-bandwidth |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|capabilities/bounding-set  |string|"&lt;cap&gt; &lt;cap&gt; ..."       |"CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"|
+An isolator is a standalone JSON object with only one required field: "name".
+All other fields are specific to the isolator.
 
-#### Types
+An executor MAY ignore isolators that it does not understand and run the container without them.
+But, an executor MUST make information about which isolators were ignored, enforced or modified available to the user.
+An executor MAY implement a "strict mode" where an image cannot run unless all isolators are in place.
 
-* uint: base 10 formatted unsigned int as a string
-* bytes: Suffix to a base 10 int to make it a K, M, G, or T for base 1024
+### Linux Isolators
+
+These isolators are specific to the Linux kernel and are impossible to represent as a 1-to-1 mapping on other kernels.
+The first example is "capabilities" but this will be expanded to include things such as SELinux, SMACK or AppArmor.
+
+#### linux/capabilities-remove-set
+
+* Scope: app
+
+**Parameters:**
+* **set** list of capabilities that will be removed from the process's capabilities bounding set
+
+```
+"name": "linux/capabilities-remove-set",
+"value": {
+  "set": [
+    "CAP_SYS_PTRACE",
+  ]
+}
+```
+
+#### linux/capabilities-retain-set
+
+* Scope: app
+
+**Parameters:**
+* **set** list of capabilities that will be retained in the process's capabilities bounding set
+
+```
+"name": "linux/capabilities-retain-set",
+"value": {
+  "set": [
+    "CAP_KILL",
+    "CAP_CHOWN"
+  ]
+}
+```
+
+### Resource Isolators
+
+A resource is something that can be consumed by a container such as memory (RAM), CPU, and network bandwidth. These resource isolators will have a request and limit quantity:
+
+- request will be the guaranteed quantity resources given to the container. Going over the request may result in throttling or denial. If request is omitted, it defaults to limit.
+
+- limit is the cap on the maximum amount of resources that will be made available to the container. If more resources than the limit are consumed, it may be terminated or throttled to no more than the limit.
+
+Limit and requests will always be a resource type's natural base units (e.g., bytes, not MB). These quantities may either be unsuffixed, have suffices (E, P, T, G, M, K, m) or power-of-two suffices (Ei, Pi, Ti, Gi, Mi, Ki). For example, the following represent roughly the same value: 128974848, "129e6", "129M" , "123Mi". Small quantities can be represented directly as decimals (e.g., 0.3), or using milli-units (e.g., "300m").
+
+#### resource/block-bandwidth
+
+* Scope: app/container
+
+**Parameters:**
+* **devicePaths** the block devices that this limit will be placed on
+* **limit** read/write bytes per second
+
+```
+"name": "resouce/block-bandwidth",
+"value": {
+  "default": true,
+  "limit": "2M"
+}
+```
+
+**TODO**: The default=true behvaior probably covers nearly 80% of all use cases but we need to define a way to target particular devices.
+
+#### resource/block-iops
+
+* Scope: app/container
+
+**Parameters:**
+* **default** must be set to true and means that this bandwidth limit applies to all interfaces except localhost by default.
+* **limit** read/write input/output operations per second
+
+```
+"name": "resource/block-iops",
+"value": {
+  "default": true,
+  "limit": "1000"
+}
+```
+
+**TODO**: The default=true behvaior probably covers nearly 80% of all use cases but we need to define a way to target particular devices.
+
+
+#### resource/cpu
+
+* Scope: app/container
+
+**Parameters:**
+* **request** milli-cores that are requested
+* **limit** milli-cores that can be consumed before the kernel temporarily throttles the process
+
+```
+"name": "resource/cpu",
+"value": {
+  "request": "250",
+  "limit": "500"
+}
+```
+
+**Note**: a milli-core is the milli-seconds/second that the app will be able to run. e.g. 1000 would represent full use of a single CPU core every second.
+
+#### resource/memory
+
+* Scope: app/container
+
+**Parameters:**
+* **request** bytes of memory that the app is requesting to use and allocations over this request will be reclaimed in case of contention
+* **limit** bytes of memory that the app can allocate before the kernel considers the container out of memory and stops allowing allocations.
+
+```
+"name": "resource/memory",
+"value": {
+  "request": "1G",
+  "limit": "2G"
+}
+```
+
+#### resource/network-bandwidth
+
+* Scope: app/container
+
+**Parameters:**
+* **default** must be set to true and means that this bandwidth limit applies to all interfaces except localhost by default.
+* **limit** read/write bytes per second
+
+```
+"name": "resource/network-bandwidth",
+"value": {
+  "default": true,
+  "limit": "1G"
+}
+```
+
+**NOTE**: Limits SHOULD NOT apply to localhost communication between apps in a container.
+**TODO**: The default=true behvaior probably covers nearly 80% of all use cases but we need to define a network interface tagging spec for appc.
 
 ## App Container Image Discovery
 
@@ -434,20 +563,24 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
         ],
         "isolators": [
             {
-                "name": "cpu/shares",
-                "value": "20"
+                "name": "resources/cpu",
+                "value": {
+                    "request": "250",
+                    "limit": "500"
+                }
             },
             {
-                "name": "memory/limit",
-                "value": "1G"
+                "name": "resource/memory",
+                "value": {
+                    "request": "1G",
+                    "limit": "2G"
+                }
             },
             {
-                "name": "cpu/mask",
-                "value": "0-3"
-            },
-            {
-                "name": "capabilities/bounding-set",
-                "value": "CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"
+                "name": "linux/capabilities-retain-set",
+                "value": {
+                    "set": ["CAP_NET_BIND_SERVICE", "CAP_SYS_ADMIN"]
+                }
             }
         ],
         "mountPoints": [
@@ -524,7 +657,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
         * **post-stop** - executed if the main **exec** process is killed. This can be used to cleanup resources in the case of clean application shutdown, but cannot be relied upon in the face of machine failure.
     * **workingDirectory** (string, optional) working directory of the launched application, relative to the application image's root (must be an absolute path, defaults to "/", ACE can override). If the directory does not exist in the application's assembled rootfs (including any dependent images and mounted volumes), the ACE must fail execution.
     * **environment** (list of objects, optional) represents the app's environment variables (ACE can append). The listed objects must have two key-value pairs: **name** and **value**. The **name** must consist solely of letters, digits, and underscores '_' as outlined in [IEEE Std 1003.1-2001](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html). The **value** is an arbitrary string.
-    * **isolators** (list of objects, optional) well-known isolation steps that should be applied to the app. The listed objects should have two key-value pairs: **name** is restricted to the [AC Name](#ac-name-type) formatting and **value** can be a freeform string. Any isolators specified in the Image Manifest can be overridden at runtime via the Container Runtime Manifest. The executor can either ignore isolator keys it does not understand or error. In practice this means there might be certain isolators (for example, an AppArmor policy) that an executor doesn't understand so it will simply skip that entry.
+    * **isolators** (optional) list of isolation steps that should be applied to the app.
+        * **name** is restricted to the [AC Name](#ac-name-type) formatting
     * **mountPoints** (list of objects, optional) locations where a container is expecting external data to be mounted. The listed objects should contain three key-value pairs: the **name** indicates an executor-defined label to look up a mount point, and the **path** stipulates where it should actually be mounted inside the rootfs. The name is restricted to the AC Name Type formatting. **readOnly" should be a boolean indicating whether or not the mount point should be read-only (defaults to "false" if unsupplied).
     * **ports** (list of objects, optional) are protocols and port numbers that the container will be listening on once started. All of the keys in the listed objects are restricted to the AC Name formatting. This information is primarily to help the user find ports that are not well known. It could also optionally be used to limit the inbound connections to the container via firewall rules to only ports that are explicitly exposed.
         * **socketActivated** (boolean, optional) if set to true, the application expects to be [socket activated](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) on these ports. The ACE must pass file descriptors using the [socket activation protocol](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) that are listening on these ports when starting this container. If multiple apps in the same container are using socket activation then the ACE must match the sockets to the correct apps using getsockopt() and getsockname().
@@ -608,6 +742,15 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming 
                         "value": "1.0.0"
                     }
                 ],
+                "isolators": [
+                    {
+                        "name": "resource/memory",
+                        "value": {
+                            "request": "1G",
+                            "limit": "2G"
+                        }
+                    }
+                ],
                 "annotations": [
                     {
                         "name": "foo",
@@ -618,6 +761,15 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming 
                     {"volume": "work", "mountPoint": "backup"}
                 ]
             }
+            "annotations": [
+                {
+                    "name": "foo",
+                    "value": "baz"
+                }
+            ],
+            "mounts": [
+                 {"volume": "work", "mountPoint": "backup"}
+            ]
         },
         {
             "name": "register",
@@ -643,9 +795,11 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming 
     ],
     "isolators": [
         {
-           "name": "memory/limit",
-           "value": "4G"
-        }
+            "name": "resource/memory",
+            "value": {
+                "limit": "4G"
+            }
+        },
     ],
     "annotations": [
         {
--- a/SPEC.md
+++ b/SPEC.md
@@ -225,7 +225,7 @@ The first example is "capabilities" but this will be expanded to include things 
 * **set** list of capabilities that will be removed from the process's capabilities bounding set
 
 ```
-"name": "linux/capabilities-remove-set",
+"name": "os/linux/capabilities-remove-set",
 "value": {
   "set": [
     "CAP_SYS_PTRACE",
@@ -241,7 +241,7 @@ The first example is "capabilities" but this will be expanded to include things 
 * **set** list of capabilities that will be retained in the process's capabilities bounding set
 
 ```
-"name": "linux/capabilities-retain-set",
+"name": "os/linux/capabilities-retain-set",
 "value": {
   "set": [
     "CAP_KILL",
--- a/schema/types/isolator.go
+++ b/schema/types/isolator.go
@@ -1,6 +1,83 @@
 package types
 
+import (
+	"encoding/json"
+	"errors"
+)
+
+var (
+	isolatorMap map[string]IsolatorValueConstructor
+)
+
+func init() {
+	isolatorMap = make(map[string]IsolatorValueConstructor)
+}
+
+type IsolatorValueConstructor func() IsolatorValue
+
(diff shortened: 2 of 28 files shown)


philips:
@vishh @rjnagal @vmarmol Any feedback from all of you would be much appreciated since I borrowed heavily from the resource proposal and lmctfy when it seemed practical. The big outlier is the block-weight since there is really no way to introspect the iops/bandwidth of a device AFAIK.

SPEC.md
```diff
@@ -201,25 +201,170 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+"name": "resource/network-bandwidth",
+"limit": {
+	"read": "1G",
+	"write": "2G"
+}
+```
+
+**NOTE**: Limits SHOULD not apply to localhost communication between apps in a container.
+
+#### resource/network-bandwidth
```
jheiss: Looks like a duplicate of the previous stanza

SPEC.md
```diff
@@ -157,7 +157,7 @@ These details are orthogonal to the runtime environment.
 
 A container executes one or more apps with shared PID namespace, network namespace, mount namespace, IPC namespace and UTS namespace.
 Each app will start chrooted into its own unique read-write rootfs before execution.
-The definition of the container is a list of apps that should be launched together, along with isolators that should apply to the entire container.
+The definition of the container is a list of apps that should be launched together, along with solators that should apply to the entire container.
```
jheiss: typo?

SPEC.md
```diff
@@ -201,25 +201,170 @@ Note that logging mechanisms other than stdout and stderr are not required by th
 ### Isolators
 
 Isolators enforce resource constraints rather than namespacing.
-Isolators may be applied to individual applications, to whole containers, or to both.
+Isolators may be scoped to individual applications, to whole containers, or to both.
```
jonboulle: We should be very explicit here about the required/desired behaviour of executors when a specified isolator cannot be fulfilled; I've been asked this a number of times now (even if it is sort of implied later on that they are optional). 

And especially so if we are introducing platform-specific isolators.

jonboulle:
Perhaps @thockin could also provide feedback.

SPEC.md
philips: @jonboulle Thoughts on this? d6b759dabb82e7d567c9b68e007f38d07422d7e8

jonboulle: I suppose it is OK, but "notifying the user" seems a bit vague. @cdaylward any better ideas?

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
-#### Types
+An executor MAY ignore isolators that it does not understand and run the container without them.
+But, an executor MUST notify the user of the isolators that were ignored during the run.
+An executor MAY implement a "strict mode" where an image cannot run unless all isolators are in place.
 
-* uint: base 10 formatted unsigned int as a string
-* bytes: Suffix to a base 10 int to make it a K, M, G, or T for base 1024
+### Linux Isolators
+
+These isolators are specific to the Linux Kernel and are impossible to represent as a 1-to-1 mapping on other Kernels.
```
jonboulle: Kernel is never capitalised

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+}
+```
+
+#### resource/block-iops
+
+* Scope: app/container
+
+**Parameters:**
+* **devicePaths** the block devices that this limit will be placed on
+* **limit** read/write iops per second
```
jonboulle: input output operations per second per second

SPEC.md
```diff
@@ -526,7 +656,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
-    * **isolators** (optional) list of well-known isolation steps that should be applied to the app. **name** is restricted to the [AC Name](#ac-name-type) formatting and **value** can be a freeform string. Any isolators specified in the Image Manifest can be overridden at runtime via the Container Runtime Manifest. The executor can either ignore isolator keys it does not understand or error. In practice this means there might be certain isolators (for example, an AppArmor policy) that an executor doesn't understand so it will simply skip that entry.
+    * **isolators** (optional) list of well-known isolation steps that should be applied to the app. 
+        * **name** is restricted to the [AC Name](#ac-name-type) formatting and
```
jonboulle: I think you forgot to finish a

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+"request": "1G",
+"limit": "2G"
+```
+
+#### resource/block-weight
+
+* Scope: app/container
+
+**Parameters:**
+* **devicePaths** the block devices that this weight will be placed on, if this is omitted then this weight will be the default for all devices
```
jonboulle: what if it's omitted but others are specified per-device?

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+These isolators are specific to the Linux Kernel and are impossible to represent as a 1-to-1 mapping on other Kernels.
+The first example is "capabilities" but this will be expanded to include things such as SELinux, SMACK or AppArmor.
+
+#### linux/capabilities-remove-set
+
+* Scope: app
+
+**Parameters:**
+* **set** list of capabilities that will be removed from the processes's capabilities bounding set
```
jonboulle: process's

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+	"CAP_SYS_PTRACE",
+]
+```
+
+#### linux/capabilities-retain-set
+
+* Scope: app
+
+**Parameters:**
+* **set** list of capabilities that will be retained in the processes's capabilities bounding set
```
jonboulle: process's

jonboulle:
I think it's going to be slightly horrid to implement the un/marshalling for this :-(

jonboulle:
While you are here do you think you could cover #54?

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+### Resource Isolators
+
+#### resource/cpu
+
+* Scope: app/container
+
+**Parameters:**
+* **request** milli-cores that are requested
+* **limit** milli-cores that can be consumed before the Kernel unschedules
```
vmarmol: maybe: s/unschedules/temporarily throttles/

vmarmol: There is always the question of the period of throttling, but we probably don't need to specify it here.

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+```
+"name": "resource/cpu",
+"request": "250",
+"limit": "500"
+```
+
+**Note**: a milli-core is the milli-seconds/second that the app will be able to run. e.g. 1000 would represent full use of a single CPU core.
```
vmarmol: of a single CPU core "every second"

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+#### resource/memory
+
+* Scope: app/container
+
+**Parameters:**
+* **request** bytes of memory that the app is requesting to use and will be pushed back to as a result of contention
+* **limit** bytes of memory that the app can allocate before the Kernel stops allowing allocations
```
vmarmol: explicitly mention OOMs? Maybe: before the kernel considers the container out of memory.

The OOM killer may not run if it was disabled.

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+```
+
+**Note**: a milli-core is the milli-seconds/second that the app will be able to run. e.g. 1000 would represent full use of a single CPU core.
+
+#### resource/memory
+
+* Scope: app/container
+
+**Parameters:**
+* **request** bytes of memory that the app is requesting to use and will be pushed back to as a result of contention
```
vmarmol: maybe: that will be reclaimed to in case of contention?

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+	"read": "2000",
+	"write": "512"
+}
+```
+
+#### resource/network-bandwidth
+
+* Scope: app/container
+
+**Parameters:**
```
vmarmol: Maybe specify device? That would assume a container can have multiple devices. May be simpler to assume only one for now.

vmarmol:
All nits and comments, LGTM overall.

SPEC.md
cdaylward: I think "notify the user" is ok but does somewhat conjure the image of a human in front of a keyboard (e.g. Rocket). An executor might be a component in a larger system. Something like "made available"? Related, should isolation information (ignored/enforced/modified) be made available to the container by way of the metadata server?

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
-|cpu/shares                 |string|"&lt;uint&gt;"                      |"4096"                              |
-|memory/limit               |string|"&lt;bytes&gt;"                     |"1G", "5T", "4K"                    |
-|cpu/mask                   |string|"&lt;cpu range&gt;"                 |"0", "0-3", "0,2,4-7"               |
-|block-io/read-bandwidth    |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
-|block-io/write-bandwidth   |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
-|network-io/read-bandwidth  |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|network-io/write-bandwidth |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|capabilities/bounding-set  |string|"&lt;cap&gt; &lt;cap&gt; ..."       |"CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"|
+An isolator is a standalone JSON object with only one required field: "name".
+All other fields are specific to the isolator.
```
cdaylward: I feel that this violates the strong schema convention we've been pushing for. Specifically that "keys" are always part of the schema and never user-supplied fields. What if the isolator is a user-specified JSON object under a key? Perhaps "parameters"? e.g.
```
"name": "linux/capabilities-remove-set",
"parameters": { 
                "set": [
                    "CAP_SYS_PTRACE",
                ]
}
```

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+* Scope: app/container
+
+**Parameters:**
+* **devicePaths** the block devices that this weight will be placed on, if this is omitted then this weight will be the default for all devices
+* **request** read/write fraction of device I/O time between 1 and 100 allocated to this container
+
+```
+"name": "resource/block-weight",
+"default: false,
```
cdaylward: missing end quote for default.

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+"name": "resource/memory",
+"request": "1G",
+"limit": "2G"
+```
+
+#### resource/block-weight
+
+* Scope: app/container
+
+**Parameters:**
```
cdaylward: Missing definition for "default" parameter.

SPEC.md
```diff
@@ -201,25 +201,157 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+```
+"name": "resource/network-bandwidth",
+"limit": {
+	"read": "1G",
+	"write": "2G"
+}
+```
+
+**NOTE**: Limits SHOULD not apply to localhost communication between apps in a container.
```
cdaylward: RFC nit: SHOULD NOT

vishh:
LGTM. Are there plans to add other resources like fd limits and disk usage?

philips:
@vishh Absolutely, I think we need to do all of the rlimit stuff. I added an issue for that: https://github.com/appc/spec/issues/197

I was considering the disk usage one but I skipped it because I was unsure how to actually implement it on Linux. Also, would it apply to volumes or just the app rootfs? Any pointers on that one? We can discuss here: https://github.com/appc/spec/issues/198

SPEC.md
philips: how about this, and we can discuss making this information available via the metadata service: "But, an executor MUST make information about which isolators were ignored, enforced or modified to the user."

SPEC.md
philips: thinking about it overnight I am just going to make default an explicit flag.

SPEC.md
philips: Yes, I am going to add a "default"  and then device list like the block-bandwidth. It is straightforward enough to implement.

philips: Fixed.

SPEC.md
philips: fixed.

philips:
@cdaylward @jonboulle Looking for an LGTM from one of you now :)

thockin:
This seems like a reasonable interpretation of the problem space.
Kubernetes turned this around a bit and made the field name the primary
index, asserting that all resource types have the same basic fields.  We
haven't implemented it very far yet, but we'll see how it plays out.

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
-|block-io/write-bandwidth   |string|"&lt;path to file&gt; &lt;bytes&gt;"|"/tmp 1K"                           |
-|network-io/read-bandwidth  |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|network-io/write-bandwidth |string|"&lt;device name&gt; &lt;bytes&gt;" |"eth0 100M"                         |
-|capabilities/bounding-set  |string|"&lt;cap&gt; &lt;cap&gt; ..."       |"CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"|
+An isolator is a standalone JSON object with only one required field: "name".
+All other fields are specific to the isolator.
 
-#### Types
+An executor MAY ignore isolators that it does not understand and run the container without them.
+But, an executor MUST make information about which isolators were ignored, enforced or modified to the user.
```
cdaylward: Missing a final ___ to the user but looks good.

philips:
On Thu, Feb 12, 2015 at 9:16 PM, Tim Hockin <notifications@github.com>
wrote:

> This seems like a reasonable interpretation of the problem space.
> Kubernetes turned this around a bit and made the field name the primary
> index, asserting that all resource types have the same basic fields. We
> haven't implemented it very far yet, but we'll see how it plays out.
>

Hrm, the struggle with that are things like asymmetric resources like
having more outbound bandwidth than inbound. Or does the new thinking have
the constraints serialize direction into the name too?

thockin:
Internally we have never dealt with isolation in that way - too little
return for the complexity.  Embedding the direction into the name is
plausible.

philips:
@thockin A number of people think of bandwidth as asymmetric outside of Google, I think. At the two hosting places I have worked at inbound/outbound balance is an important consideration.

thockin:
Sorry, I didn't mean to imply it wasn't important, just that WE don't have
an existence proof of it on which to rely :)

philips:
@jonboulle @cdaylward Thoughts on all of this? LGTY?

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+### Resource Isolators
+
+#### resource/cpu
+
+* Scope: app/container
+
+**Parameters:**
+* **request** milli-cores that are requested
+* **limit** milli-cores that can be consumed before the Kernel temporarily throttles the process
```
jonboulle: kernel

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+```
+
+**Note**: a milli-core is the milli-seconds/second that the app will be able to run. e.g. 1000 would represent full use of a single CPU core every second.
+
+#### resource/memory
+
+* Scope: app/container
+
+**Parameters:**
+* **request** bytes of memory that the app is requesting to use and will be reclaimed to in case of contention
```
jonboulle: s#to in#in#

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+"name": "resource/network-bandwidth",
+"default": true,
+"limit": {
+	"read": "1G",
+	"write": "2G"
+}
+```
+
+**NOTE**: Limits SHOULD NOT apply to localhost communication between apps in a container.
+**TODO**: The default=true behvaior probably covers nearly 80% of all use cases but we need to define a network interface tagging spec for appc.
```
jonboulle: behavio(u)r

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+An isolator is a standalone JSON object with only one required field: "name".
+All other fields are specific to the isolator.
 
-#### Types
+An executor MAY ignore isolators that it does not understand and run the container without them.
+But, an executor MUST make information about which isolators were ignored, enforced or modified to the user.
+An executor MAY implement a "strict mode" where an image cannot run unless all isolators are in place.
 
-* uint: base 10 formatted unsigned int as a string
-* bytes: Suffix to a base 10 int to make it a K, M, G, or T for base 1024
```
jonboulle: We still need to define these, you mix suffix-less/G/GB below

jonboulle:
@philips what about moving to a name:value model as elsewhere, where value is an arbitrary object? For example - http://play.golang.org/p/RHS8YdsO_p

SPEC.md
cdaylward: Still missing an adjective here. Available?

cdaylward:
@jonboulle @philips My initial reaction to this change was this as well but I withdrew my comment. As soon as a free-form object is introduced, the _schema_ quality is lost. Also, it effectively makes parsing the JSON an ACE-specific endeavor and could limit code reuse here.

SPEC.md
cdaylward: Agreed, specifying which prefixes should be used should remain (e.g. binary-prefixes, K = 1024).  And, please don't mention kibibytes. ;)

thockin: As much as I hate the kibibytes junk, some resources are NATURALLY base 10
and some are base 2.  Disk is base 10.  Network is base 10.  Memory is base
2.

For kubernetes we sucked it up and distinguished k (I am pretty sure SI
says k not K) from Ki and M from Mi

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+**Parameters:**
+* **set** list of capabilities that will be retained in the process's capabilities bounding set
+
+```
+"name": "linux/capabilities-retain-set",
+"set": [
+	"CAP_KILL",
+	"CAP_CHOWN"
+]
+```
```
alban: What happens to the capabilities that are neither listed in linux/capabilities-remove-set or linux/capabilities-retain-set? The spec does not say whether it is a blacklist approach or a whilelist approach. This should be the latter so new dangerous capabilities added in future kernels would not be enabled by default.

In Rocket, this is likely to be implemented with systemd's option CapabilityBoundingSet and it takes a whitespace-separated list of capability names. So the change from strings to isolators objects does not help for caps.
http://www.freedesktop.org/software/systemd/man/systemd.exec.html#CapabilityBoundingSet=

I would use only one isolator "linux/capabilities-bounding-set" with the same definition as systemd's CapabilityBoundingSet (including the "~" for all).

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+* **request** milli-cores that are requested
+* **limit** milli-cores that can be consumed before the Kernel temporarily throttles the process
+
+```
+"name": "resource/cpu",
+"request": "250",
+"limit": "500"
+```
+
+**Note**: a milli-core is the milli-seconds/second that the app will be able to run. e.g. 1000 would represent full use of a single CPU core every second.
```
alban: Where does the name "milli-core" come from?
It seems close to the CPUShares option in systemd:
http://www.freedesktop.org/software/systemd/man/systemd.cgroup.html#CPUShares=weight
And it has a default of 1024 for a single CPU core.

https://www.kernel.org/doc/Documentation/scheduler/sched-design-CFS.txt

thockin: Millicores are a concrete unit.  Shares are a proportional unit.  Guess
which one makes sense to users and which one makes sense to kernel hackers.

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+"limit": "2G"
+```
+
+#### resource/block-weight
+
+* Scope: app/container
+
+**Parameters:**
+* **default** if set to true this weight will be the default for all devices. If set devicePaths must be empty.
+* **devicePaths** the block devices that this weight will be placed on.
```
alban: The isolators are specified in the Image Manifest (at build-time). But when building the ACI, it is not yet known whether the executor will extract the rootfs on /dev/sda or somewhere else. So it makes it impractical to specify the device paths at ACI build-time.

Do we have a use-case for devicePaths?

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+}
+```
+
+#### resource/network-bandwidth
+
+* Scope: app/container
+
+**Parameters:**
+* **default** must be set to true and means that this bandwidth limit applies to all interfaces except localhost by default.
+* **limit** read/write bytes per second
```
alban: Is there any cgroup or systemd helper to implement "resource/network-bandwidth" per app and per network interface in the executor? I am concerned that it might add something impossible to implement.

If it is implemented with Linux traffic control, it might be per-interface configuration and not per-app.

/cc @eyakubovich

SPEC.md
```diff
@@ -201,25 +201,161 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+
+```
+"name": "resource/network-bandwidth",
+"default": true,
+"limit": {
+	"read": "1G",
+	"write": "2G"
```
alban: 2 giga bytes per second of upload? I knew I had a slow internet connection here ;)

philips:
@cdaylward The schema can be enforced by the name of the thing. I don't know how else to do disjoint things like resources and capabilities. One is a single value unit and the other is a list.

philips:
Thinking about the resources more I am going to take @thockin 's suggestion and use a simple single value instead of having read/write types. In the future we can add a network-bandwidth-upload, etc resource constraint.

SPEC.md
thockin: The way we handle this internally is by late-binding.  We can ask for X bytes of storage on no less than N but no more than M devices, and params are assigned by relative index.  This is then late-bound by the scheduler and/or node agent to real devices.

b:
Asked this over in another issue, as well: is there a plan to distinguish between mandatory and optional isolators?

philips:
@b We could certainly do it but we would need some concrete use cases. Would you mind writing something up?

philips:
@cdaylward @jonboulle Alright, I added code to validate and backup the SPEC now.

b:
There are at least two issues I'm hoping to address with the mandatory vs optional question:

1) Non-Linux platforms (maybe out of scope?)
2) Linux platforms with differences in features (RHEL still doesn't have user namespaces and won't for some time)

In addition, I don't think defaulting to permissive on this is the best plan. If a container expects certain isolators and those isolators are unavailable, there should at least be a strict flag to cause the launch to fail, rather than allowing it to continue. Generating an error message while running the container in a way it wasn't built to operate seems a dangerous default.

philips:
@b OK, if you could write-up language I think it makes sense to specify it.

philips:
@b We should do it in a separate PR since this certainly allows for expansion of the schema but I want to get this merged.

b:
No problem, will do.

SPEC.md
philips: The plan here is to tag everything coming out of an app with the same tag and tc it. I think that works.

philips:
This whole thing is ready for review and has tests to back it up. @cdaylward @jonboulle

SPEC.md
```diff
@@ -201,25 +201,154 @@ Note that logging mechanisms other than stdout and stderr are not required by th
...
+#### linux/capabilities-remove-set
+
+* Scope: app
+
+**Parameters:**
+* **set** list of capabilities that will be removed from the process's capabilities bounding set
+
+```
+"name": "os/linux/capabilities-remove-set",
+"value": {
```
eyakubovich: Is there an aversion to making `value` have polymorphic type? It maybe slightly harder to parse in some languages/parsers but is more user friendly. So have `"value": ["CAP_SYS_PTRACE"]`

SPEC.md
eyakubovich: I think it should work but I don't know if it'll always be possible to police on the host side. Depending on the type of network interface, you may need to police on the container side. If done on the container side, an app can just remove the limits (if it has proper capabilities).

philips: @eyakubovich yea, that is fine.

SPEC.md
philips: @eyakubovich No aversion but it is simpler. If someone complains we can change it.

SPEC.md
```diff
@@ -434,20 +563,22 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
+                "value": {
+                    "request": "1G",
+                    "limit": "2G"
+                }
             },
             {
-                "name": "capabilities/bounding-set",
-                "value": "CAP_NET_BIND_SERVICE CAP_SYS_ADMIN"
+                "name": "linux/capabilities-retain-set",
+                "set": ["CAP_NET_BIND_SERVICE", "CAP_SYS_ADMIN"]
```
cdaylward: Missed a "value": conversion here.

cdaylward:
Ok, other than the one fixup in the example JSON, this looks good to me!

schema/types/isolator.go
```diff
@@ -1,6 +1,83 @@
...
+		return err
+	}
+
+	i.value = dst
+	i.Name = ii.Name
+
+	return nil
+}
+
+func (i *Isolator) assertValid() error {
```
jonboulle: You're not calling this anywhere!

PR #194, opened 2015-02-11
README: correct path to build_aci script

kaorimatz:
We have moved from coreos/rocket repository so app-container directory doesn't exist anymore.


jonboulle:
Welp, this should have been caught a long time ago. Thanks for the fix!

PR #200, opened 2015-02-16
correct typo in README

stormasm:
I fixed a TYPO in the README --- looks like a comma instead of colon is appropriate as well,
since you are not really delineating a list.  Other option was a semicolon, but no other semicolons
are used in the text (at all) so it blends in better with the style of writing.


philips:
lgtm, ty.

Issue #201, opened 2015-02-16
spec: Untrusted ACI are decompressed.


skohanim:
According to the spec ACIs are decompressed before signature verification. Since the compressed images are untrusted before signature verification, it is unsafe to run them through a decompressor since they may contain an exploit payload targeting the decompressor.
The solution would be signing the compressed ACI instead. In this case the signature can be verified before anything else takes place.
The same is true for decryption.

philips:
We thought we fixed this verify before decompress ordering thing in the spec a few weeks ago but I don't see it applied. Let me research. Very much a bug.

jonboulle:
Hmm, weird, this was supposed to be covered by a recent change but appears to have slipped through the cracks. Thanks for bringing it to our attention again.

jonboulle:
#19

jonboulle:
Oops, this was long since fixed by #204

PR #204, opened 2015-02-16
SPEC: clarify that compression/encryption should happen before signing

### LINKED ISSUE #19: spec: clarify compression/encryption ordering of ACIs
_From @sarnowski on December 8, 2014 15:4_

Hello,

I am currently implementing writing and reading of ACIs and I am unsure about the exact ordering where encryption fits in. I think I know how its meant in the specification but maybe it makes sense to make it more clear. From my understanding the procedure is as follows:

**Writing:**

1. write tarball
2. hash tarball with sha256 for *.sig file (currently, maybe 512 in the future?)
3. [optionally] compress with gzip/bzip2/xz
4. [optionally] encrypt

**Reading:**

1. [optionally] unencrypt with a key (how do I determine the key and algorithm for it? how do I find out that I have to decrypt first, try it out if no compression file format is detected?)
2. [optionally] uncompress the ACI (is there any way planned how to more easily detect the file format besides besides introspection?)
3. hash tarball and compare with hash from the *.sig file
4. unpack tarball

Thanks,
Tobi

_Copied from original issue: coreos/rocket#233_

philips:
Signing should happen after compression so the user is not decompressing
things that it may not trust.

Fixes #19 and #201

We decided to fix this via
https://github.com/appc/spec/pull/156/files#r23838943 but it never made
it into code.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -67,24 +67,22 @@ It includes a *rootfs* with all of the files that will exist in the root of the 
 The ACI file format ("image archive") aims for flexibility and relies on standard and common technologies: HTTP, gpg, tar and gzip.
 This set of formats makes it easy to build, host and secure a container using technologies that are widely available and battle-tested.
 
+- Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
 - Image archives MUST be a tar formatted file with no duplicate entries.
 - Image archives MUST have only two top-level pathnames, `manifest` (a regular file) and `rootfs` (a directory). Image archives with additional files outside of `rootfs` are not valid.
 - All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
-- Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
-- Image archives SHOULD be signed using PGP, the format MUST be ascii-armored detached signature mode.
-- Image signatures MUST be named with the suffix `.aci.asc`.
-
-There are two further transformations that may be applied to image archives for transport:
 - Image archives MAY be compressed with `gzip`, `bzip2`, or `xz`.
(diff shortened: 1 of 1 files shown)


jonboulle:
Hmm, only issue is that we aren't yet annotating the armored sig which would be nice, maybe create a follow up issue for that? (In the meantime I suppose we just sniff for known types and if that fails then attempt to decrypt?)

philips:
@jonboulle I don't remember the details of that. Can you file an issue we can merge this?

SPEC.md
```diff
@@ -70,21 +70,19 @@ This set of formats makes it easy to build, host and secure a container using te
 - Image archives MUST be a tar formatted file with no duplicate entries.
 - Image archives MUST have only two top-level pathnames, `manifest` (a regular file) and `rootfs` (a directory). Image archives with additional files outside of `rootfs` are not valid.
 - All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
+- Image archives MAY be compressed with `gzip`, `bzip2`, or `xz`.
+- Image archives MAY be encrypted with AES symmetric encryption, after (optional) compression.
 - Image archives MUST be named with the suffix `.aci`, irrespective of compression/encryption (see below).
 - Image archives SHOULD be signed using PGP, the format MUST be ascii-armored detached signature mode.
```
jonboulle: actually could you please add "after (optional) compression/encryption" to be completely explicit, as the list is not technically ordered

jonboulle:
> it seems like gpg armor should be on the last line

ack, good catch

jonboulle:
https://github.com/appc/spec/issues/206

jonboulle:
ping

philips:
Refreshed and merging.

Issue #103, opened 2015-01-09
spec: clarify pod uuids


jonboulle:
[CRMs](https://github.com/appc/spec/blob/master/SPEC.md#container-runtime-manifest-schema) have a **uuid** but it is underspecified; we should explicate the purpose and how it might be generated.

Also, in the [metadata server](https://github.com/appc/spec/blob/master/SPEC.md#container-metadata) it's referred to as *uid*, so this should be changed to *uuid* for consistency.

jonboulle:
Another aspect to this is that since UUIDs are "practically unique" and not "guaranteed unique" we might want to document the theoretical possibility of a conflict, particularly when it comes to the identity verification in the metadata service. 

Extremely security-conscious implementations of the spec would likely leverage a genuine (shared/distributed) unique ID rather than just generating a UUID locally in the executor.

eyakubovich:
UUIDs are conceptually unique in time and space. The whole point of using them is that you don't worry that they're not. All types, if properly implemented, are unique enough that you don't have to concern yourself otherwise. If you're worried about probabilities on those scales, then most of cryptography in use is broken anyway and "extremely security conscious" couldn't even use encrypt their data.

jonboulle:
Sure, assuming "properly implemented", but there are also other reasons a centralised generator is useful, e.g. auditing/policy.

Anyway, feel free to do the wordsmithing to your own satisfaction :-)

updated references:
https://github.com/appc/spec/blob/de93ed4e7c40739b90011fdc4d93553417f0d9f3/SPEC.md#pod-uuid
https://github.com/appc/spec/blob/de93ed4e7c40739b90011fdc4d93553417f0d9f3/SPEC.md#metadata-service

jonboulle:
better covered by #324

PR #207, opened 2015-02-18
spec: introduce Pods, remove CRMs

jonboulle:
Following on from #150, this is the next iteration at an attempt to introduce
the concept of pods.

Following the most recent discussion on the topic, this patch attempts to
simplify things greatly by removing the ContainerRuntimeManifest entirely and
settling instead on a single PodManifest.

To achieve this, we give the PodManifest a dual purpose: it is both a
_deployable template_ (which can be used to stamp out instances of pods) which
before execution must be resolved into a final/reified version. The allowed
reified versions are a _subset_ of the template versions; for example, app
references in the reified version of a Pod Manifest MUST be deterministic and
contain image IDs (i.e. they cannot rely on labels and dependency matching).

An important distinction from the previous CRM model is that a reified 
PodManifest no longer _necessarily_ includes runtime information specific to
the executor (i.e., the UUID). It is hence possible that a fully reified pod
could be passed around as a deployable unit to different executors.

This patch also goes to some length to eliminate the (frustratingly ambiguous)
word "container" from the spec, except for in qualified contexts (e.g. "App
Container Image").


jonboulle:
@eyakubovich @vcaputo @cdaylward @philips @mpasternacki have at it, please

SPEC.md
```diff
@@ -2,33 +2,37 @@
...
+The specification defines an image format, image discovery mechanism, and execution environment that can exist in several independent implementations.
+The core goals of the specification include:
 
-* Design for fast downloads and starts of the containers
-* Ensure images are cryptographically verifiable and highly-cacheable
-* Design for composability and independent implementations
-* Use common technologies for crypto, archive, compression and transport
-* Use the DNS namespace to name and discover container images
+* Designing for fast downloads and starts of App Containers
+* Ensuring images are cryptographically verifiable and highly cacheable
```
jonboulle: Sorry for all the noise in this diff, but while trying to eliminate the word "container" I couldn't help polish a lot of these edges

SPEC.md
```diff
@@ -118,32 +122,60 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
...
+The "*executor*" perspective consists of the steps that the App Container Executor (ACE) must take to set up the environment for the pod and applications.
+The "*app*" perspective is how the app processes inside the pod see the environment.
 
 ### Executor Perspective
 
+#### Pod Manifest
+
+The deployable, executable unit in the App Container specification is the **pod**.
+A **pod** executes one or more apps inside a shared context.
+The shared context is defined as the conjunction of the following namespaces:
```
jonboulle: As I reached this part it became evident that we will probably need to move towards a more agnostic (i.e. less Linux centric) definition of these namespaces

SPEC.md
```diff
@@ -118,32 +122,60 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
...
+- UTS namespace
+Each app in a pod will start chrooted into its own unique read-write rootfs before execution.
+The definition of the **pod** is a list of apps that should be launched together, along with isolators that should apply to the entire pod.
+This is codified in a [Pod Manifest](#pod-manifest-schema).
+Pod Manifest serve the role of _deployable templates_  and _runtime manifest_: a template can be a candidate for a series of transformations before execution.
+For example, a Pod Manifest might reference an app with a label requirement of `version=latest`, which another tool might subsequently resolve to a specific version.
+Another example would be that volumes are "late-bound" by the executor; alternatively, an executor might add annotations.
+The fully resolved, finalised pod manifest is exposed through the metadata service.
+
+TODO(jonboulle): define what a reified/finalised pod manifest is
```
jonboulle: here is where I got tired...

also: thoughts on terminology for the two forms? I think "template" is relatively uncontroversial, but what about the other side: "runtime"? "reified"? "resolved"? (something not starting with R? ;-)

SPEC.md
```diff
@@ -621,9 +643,8 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming
...
-* **acKind** (string, required) must be set to "ContainerRuntimeManifest"
-* **uuid** (string, required) must be an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt) that represents this instance of the container (must be in [RFC4122](http://www.ietf.org/rfc/rfc4122.txt) format)
-* **apps** (list of objects, required) list of apps that will execute inside of this container. Each app object has the following set of key-value pairs:
+* **acKind** (string, required) must be set to "PodManifest"
+* **apps** (list of objects, required) list of apps that will execute inside of this pod. Each app object has the following set of key-value pairs:
     * **app** (string, optional) name of the app (restricted to AC Name formatting)
     * **imageID** (string, required) content hash of the image that this app will execute inside of (must be of the format "type-value", where "type" is "sha512" and value is the hex encoded string of the hash)
     * **mounts** (list of objects, optional) list of mounts mapping an app mountPoint to a volume. Each mount has the following set of key-value pairs:
```
eyakubovich: Do we need `mounts` in the PodManifest? If it's not in the ImageManifest, an app doesn't expect that this path will escape containment. Are we breaking the basic premise that a container explicitly states which parts are "exposed"?

SPEC.md
```diff
@@ -633,8 +654,8 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming
...
 * **volumes** (list of objects, optional) list of volumes which should be mounted into each application's filesystem
     * **name** (string, required) used to map the volume to an app's mountPoint at runtime. (restricted to the AC Name formatting)
-    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (writes go to container). "host" fulfills a mount point with a bind mount from a **source**.
-    * **source** (string, required if **kind** is "host") absolute path on host to be bind mounted into the container under a mount point.
+    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (writes go to pod). "host" fulfills a mount point with a bind mount from a **source**.
```
eyakubovich: This seems restrictive. What if I want to satisfy a volume via something remote via nfs, smb, etc? Do I have to first mount it onto the host and then bind-mount it into the container?

SPEC.md
philips: Yea, at this point all containers are linux container so I am not terribly concerned.

jonboulle: Jetpack, pal!

SPEC.md
jonboulle: Maybe? Are you proposing adding a set of well-known mount protocols?

SPEC.md
jonboulle: > Do we need mounts in the PodManifest?

>  Are we breaking the basic premise that a container explicitly states which parts are "exposed"?

What does "container" mean in this sentence?

eyakubovich: @jonboulle By container I meant app (aka image). When I package up my app into an ACI, can I assume that when it runs, the files in /usr/lib will be exactly the ones that are in the ACI unless I specified that directory (or its sub-dir) in the image manifest? That seems like a desirable property that maybe broken here. On the flip side, giving an ability for execution environment to override anything in the ACI maybe a necessary reality.

SPEC.md
eyakubovich: I would at least allow for other kinds. Change the `either "emtpy" or "host".` to something like `"empty", "host", or executor specific value.`

SPEC.md
jonboulle: To make sure I understand correctly, are you talking about mounts shadowing files/directories in the ACI?

SPEC.md
jonboulle: I think this makes sense. @cdaylward @philips ?

SPEC.md
```diff
@@ -2,33 +2,37 @@
 
 _For version information, see [VERSION](VERSION)_
 
-The "App Container" defines an image format, image discovery mechanism and execution environment that can exist in several independent implementations. The core goals include:
+"App Container" (appc) is a specification describing how applications can be packaged, distributed, and executed in a portable and self-contained way.
+The specification defines an image format, image discovery mechanism, and execution environment that can exist in several independent implementations.
```
cdaylward: s#that can exist in several independent implementations##
This is called out as a core goal below (composable, implementation independent). Keep this sentence short and sweet?

SPEC.md
```diff
@@ -2,33 +2,37 @@
...
 
 
 ## Example Use Case
 
 To provide context to the specs outlined below we will walk through an example.
 
-A user wants to launch a container running three processes.
+A user wants to launch an "App Container" running three processes.
 The three processes the user wants to run are the apps named `example.com/reduce-worker-register`, `example.com/reduce-worker`, and `example.com/reduce-backup`.
 First, the executor will check the cache and find that it doesn't have images available for these apps.
```
cdaylward: cache is not introduced but used in this line as if a necessary part of an executor. Perhaps just strike this line and start on line 35, "First, an executor will make an HTTPS request..."

SPEC.md
```diff
@@ -2,33 +2,37 @@
...
-So, it will make an HTTPS request to example.com and using the `<meta>` tags there finds that the containers can be found at:
+So, it will make an HTTPS request to example.com and using the `<meta>` tags there finds that the images can be found at:
```
cdaylward: "First, the executor makes an HTTPS request to example.com and inspect the `<meta>` tags to determine where the images can be found:" ?

SPEC.md
```diff
@@ -38,12 +42,12 @@ The executor downloads these three images and puts them into its local on-disk c
 Then the executor extracts three fresh copies of the images to create instances of the "on-disk app format" and reads the three image manifests to figure out what binaries will need to be executed.
 
 Based on user input, the executor now sets up the necessary cgroups, network interfaces, etc. and runs the `pre-start` event handlers for each app.
-Next, it forks the `reduce-worker`, `worker-backup`, and `register` processes in their shared namespaces inside the container, chrooted into their respective root filesystems.
+Next, it forks the `reduce-worker`, `worker-backup`, and `register` processes in their shared namespaces, chrooted into their respective root filesystems.
 
-At some point, the container will get some notification that it needs to stop.
+At some point, the App Container will get some notification that it needs to stop.
```
cdaylward: This might not be clear as to why. Parenthetical (e.g. upon host shutdown)?

SPEC.md
```diff
@@ -118,32 +122,60 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
...
 There are two "perspectives" in this process.
-The "*executor*" perspective consists of the steps that the container executor must take to set up the containers. The "*app*" perspective is how the app processes inside the container see the environment.
+The "*executor*" perspective consists of the steps that the App Container Executor (ACE) must take to set up the environment for the pod and applications.
+The "*app*" perspective is how the app processes inside the pod see the environment.
 
 ### Executor Perspective
 
+#### Pod Manifest
+
+The deployable, executable unit in the App Container specification is the **pod**.
```
cdaylward: It might be more clear to introduce the pod concept separately from the manifest and before introducing the details of the executor. A small top-level section right before ## App Container Executor?

## Pods

The deployable, executable unit in the App Container specification is the **pod**. A **pod** is a list of apps that should be launched together inside a shared context. The shared context is defined as the conjunction of the following namespaces:
- PID namespace
- network namespace
- mount namespace
- IPC namespace
- UTS namespace

---

As with the CRM, it's still not fully clear if a pod manifest is something that is transportable and ingestible, or just an executor-generated artifact made available to the apps via the metadata service. Should we clarify this?

SPEC.md
cdaylward: People really do use jail() :) Perhaps something like...

A shared context is defined as the restricted access to or view of the following:

- process table
- network
- devices
- mount table
- file system
- IPC
- UTS identifiers

SPEC.md
cdaylward: "Template" is certainly fine. "Reified" makes the most sense to me in the context of a template that is now concrete having its parameters defined.

SPEC.md
```diff
@@ -118,32 +122,60 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
...
+
+The executor must assign an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt) to each pod.
+This UUID is exposed to the pod through the [Metadata Service](#app-container-metadata-service).
+
 #### Filesystem Setup
 
-Every execution of an app container should start from a clean copy of the app image.
-The simplest implementation will take an application container image (ACI) and extract it into a new directory:
+Every execution of an app should start from a clean copy of its image (ACI).
+The simplest implementation will take an ACI with no dependencies and extract it into a new directory:
```
cdaylward: implementation or case?

SPEC.md
```diff
@@ -154,37 +186,24 @@ tar xzvf /var/lib/pce/hello.aci -C hello
...
-An App Container must have a [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interface, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
-The network interface should be configured with an IPv4/IPv6 address that is reachable from other containers.
+A Pod must have a [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interface, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
+The network interface should be configured with an IPv4/IPv6 address that is reachable from other pods.
 
 #### Logging
 
-Apps should log to stdout and stderr. The container executor is responsible for capturing and persisting the output.
+Apps should log to stdout and stderr. 
+The ACE is responsible for capturing and persisting the output.
```
cdaylward: "may capture and persist"?

SPEC.md
```diff
@@ -512,32 +534,32 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
-    * **created** date on which this container was built (string, must be in [RFC3339](https://www.ietf.org/rfc/rfc3339.txt) format)
-    * **authors** contact details of the people or organization responsible for the containers (freeform string)
-    * **homepage** URL to find more information on the container (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on this container (string, must be a URL with scheme HTTP or HTTPS)
+    * **created** date on which the image was built (string, must be in [RFC3339](https://www.ietf.org/rfc/rfc3339.txt) format)
+    * **authors** contact details of the people or organization responsible for the image (freeform string)
+    * **homepage** URL to find more information on the imgae (string, must be a URL with scheme HTTP or HTTPS)
```
cdaylward: s#imgae#image#

SPEC.md
cdaylward: Both "writes go to container" and "writes go to pod" feel wrong here now. Each app in a pod has its own chroot. "Host" mounts would be shared between apps via the bind mount in each chroot, but "empty" now seems to imply writes go to each app's respective chroot, not "to pod". With the introduction of pods, would changing "empty" to "pod" make it more clear that this is a shared volume for apps in a pod? Would probably make sense to reword the description too ("path exists" isn't sufficient to imply a shared volume). Leave "empty" as dummy volume per chroot, add "pod" as shared volume?

SPEC.md
```diff
@@ -633,8 +654,8 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming
...
 * **volumes** (list of objects, optional) list of volumes which should be mounted into each application's filesystem
     * **name** (string, required) used to map the volume to an app's mountPoint at runtime. (restricted to the AC Name formatting)
-    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (writes go to container). "host" fulfills a mount point with a bind mount from a **source**.
-    * **source** (string, required if **kind** is "host") absolute path on host to be bind mounted into the container under a mount point.
+    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (writes go to pod). "host" fulfills a mount point with a bind mount from a **source**.
+    * **source** (string, required if **kind** is "host") absolute path on host to be bind mounted into the pod under a mount point.
```
cdaylward: Each app in the pod might have a different mount point for the same volume source. Correct? "[...] bind mounted under a mount point in each app's chroot."? Not sure about the wording but do you agree that "into the pod" is not fully clear?

examples/pod_runtime.json
```diff
@@ -13,7 +12,7 @@
                 {"volume": "database", "mountPoint": "db"},
                 {"volume": "buildoutput", "mountPoint": "out"}
             ],
-            "imageID": "sha512-...",
+            "imageID": "sha512-qwerty",
```
cdaylward: Make this hex as the example is sha512, please. In fact it would be nice to make this a proper sha512 string so the examples can actually be passed through a validator.

examples/pod_runtime.json
```diff
@@ -29,7 +28,7 @@
         },
         {
             "app": "example.com/reduce-worker-register",
-            "imageID": "sha512-..."
+            "imageID": "sha512-zxcvbn"
```
cdaylward: Same, hex please. Or, preferably a proper sha512 string.

SPEC.md
jonboulle: What is "devices"?

jonboulle: SGTM other than that one. @philips ?

SPEC.md
jonboulle: ack, done

SPEC.md
jonboulle: mmm, I want to rip out this example entirely soon (and put it in EXAMPLE.md or so)

SPEC.md
jonboulle: tweaked

SPEC.md
jonboulle: added

SPEC.md
jonboulle: thnaks

SPEC.md
jonboulle: yeah, not sure on wording

examples/pod_runtime.json
jonboulle: hgurghgurghle

SPEC.md
jonboulle: Tweaked slightly, we can revisit when considering other mount types.

SPEC.md
jonboulle: Implementation imho

SPEC.md
jonboulle: > It might be more clear to introduce the pod concept separately from the manifest and before introducing the details of the executor. A small top-level section right before ## App Container Executor?

I am going back and forth on this, struggling to make it fit with the greater structure (e.g. the top level three key sections of the spec)

jonboulle: > As with the CRM, it's still not fully clear if a pod manifest is something that is transportable and ingestible, or just an executor-generated artifact made available to the apps via the metadata service. Should we clarify this?

I thought I had made this clear :disappointed: , will rework a bit more

jonboulle:
@cdaylward ready for another review

/cc @thockin

SPEC.md
```diff
@@ -116,34 +119,62 @@ In this case, the app can reference the name "example.com/trusted-certificate-au
...
+- PID namespace
+- network namespace
+- mount namespace
+- IPC namespace
+- UTS namespace
+The context may also consist of one or more isolators.
+
+Each app in a pod will start chrooted into its own unique read-write rootfs before execution.
+
+The definition of the **pod** - namely, the list of constituent apps, and any isolators that should apply to the entire pod - is codified in a [Pod Manifest](#pod-manifest-schema).
```
thockin: Take this for what it is worth (not much) but "manifest" feels like the wrong word here (and it has legacy meaning in kubernetes, but that is not your problem :).  I would expect a "pod manifest" to be a list of pods.  We originally called the same object a "container manifest" (here's the legacy) but now we just call it "pod"

SPEC.md
```diff
@@ -116,34 +119,62 @@ In this case, the app can reference the name "example.com/trusted-certificate-au
...
+At runtime, the reified Pod manifest is exposed to applications through the metadata service.
+
+Pod TODO:
+* Define the garbage collection lifecycle of the pod filesystem including:
     * Format of app exit code and signal
     * The refcounting plan for resources consumed by the ACE such as volumes
-* Define the lifecycle of the container as all exit or first to exit
-* Define security requirements for a container. In particular is any isolation of users required between containers? What user does each application run under and can this be root (i.e. "real" root in the host).
-* Define how apps are supposed to communicate; can they/do they 'see' each other (a section in the apps perspective would help)?
+* Define the lifecycle of the pod as all exit or first to exit
```
thockin: We define pod lifecycle as done when all containers have exited.  We spec the restart policy on the pod, not on each container.  In effect, a "restart always" pod can never terminate except by administrative or system action (e.g. node is decommissioned).

SPEC.md
```diff
@@ -116,34 +119,62 @@ In this case, the app can reference the name "example.com/trusted-certificate-au
...
+
+The deployable, executable unit in the App Container specification is the **pod**.
+A **pod** is a list of apps that should be launched together inside a shared context.
+The shared context is defined as the conjunction of the following Linux namespaces (or equivalents on other operating systems):
+- PID namespace
+- network namespace
+- mount namespace
+- IPC namespace
+- UTS namespace
+The context may also consist of one or more isolators.
```
thockin: Maybe specifically call out that pod-scope isolators bound container-scope isolators?  Also what is the semantic if a container specs an isolator that is not present on the pod?

jonboulle: Will call this out in the "Isolators" section below; also will follow up to clarify the wording once #192 lands (nudge @philips )

SPEC.md
jonboulle: Hrm, we use the term "Image Manifest" for the JSON blob which describes a single image, not a list of images, so this is internally consistent. I am struggling to think of another term right now.

thockin: Fair enough, naming is hard and I do not have a strong track record
(lmctfy) :)

jonboulle:
Also fixes #103

SPEC.md
cdaylward: This makes sense from my perspective on things. For one, I think we can agree that "process still resident" isn't any real sort of health metric, so pod health and pod lifecycle management are likely to be, in effect, out of band from the executor (e.g. a concern of monitoring/scheduling system). Additionally, there are likely scenarios where a pod has several apps but one is expected to, for instance, do some work and terminate.

SPEC.md
```diff
@@ -117,34 +120,62 @@ The dependencies are applied in order and each image dependency can overwrite fi
...
+
+Each app in a pod will start chrooted into its own unique read-write rootfs before execution.
```
cdaylward: Execution detail in the middle of definitions. Move into or after paragraph at starting line 149.

SPEC.md
```diff
@@ -117,34 +120,62 @@ The dependencies are applied in order and each image dependency can overwrite fi
...
+The definition of the **pod** - namely, the list of constituent apps, and any isolators that should apply to the entire pod - is codified in a [Pod Manifest](#pod-manifest-schema).
+Pod Manifests can serve the role of both _deployable template_ and _runtime manifest_: a template can be a candidate for a series of transformations before execution.
+For example, a Pod Manifest might reference an app with a label requirement of `version=latest`, which another tool might subsequently resolve to a specific version.
+Another example would be that volumes are "late-bound" by the executor; alternatively, an executor might add annotations.
+
+A Pod Manifest must be fully resolved (_reified_) before execution.
+Specifically, a Pod Manifest must have all `mountPoint`s satisfied by `volume`s, and must reference all applications deterministically (by image ID).
+At runtime, the reified Pod manifest is exposed to applications through the metadata service.
+
+Pod TODO:
```
cdaylward: If start and stop order of apps within a pod is an executor implementation detail, that should probably be briefly clarified.

SPEC.md
philips: yes, lets wait until #192

SPEC.md
philips: I think these are generic enough terms that I am OK with it.

SPEC.md
philips: @thockin If I want to update the pod it seems hard to express this: "If you want to change the isolators on the 'running pod' you can submit a modified 'pod' to the API"?

SPEC.md
philips: @eyakubovich ping

philips:
this lgtm

SPEC.md
thockin: I'm not clear what "the API" is here.  Update is more than just isolation changes.

alban:
I think the Pods spec should talk about user ids and group ids of containers in Pods.

If a pod contain one mysql app and one apache app, they might want to run as users mysql and www-data respectively.

The files in the apache aci might belong to user ID 33 but unless apache.aci contains /etc/passwd, uid=33 does not mean anything.

At the moment, Rocket does not have a /etc/passwd and it cannot start non-root processes:
https://github.com/coreos/rocket/issues/539

examples/pod_template.json
```diff
@@ -0,0 +1,34 @@
+{
+    "acVersion": "0.3.0",
+    "acKind": "PodManifest",
+    "apps": [
+        {
+            "app": "example.com/reduce-worker",
```
philips: I think we should consider making this an object so:

```
"image": {
  "name": "example.com/worker-backup",
  "labels": ...,
  "id": "sha512-"
}
```

SPEC.md
eyakubovich: @jonboulle yes. So the image might assume that /opt/foo always has contents "foo" because it didn't specify /opt as dest for any volume. And then it runs and finds that /opt/foo contains "bar".

SPEC.md
jonboulle: Any suggestions on wording? struggling a bit

jonboulle:
@alban Can you file another issue to discuss uids/gids? It's not strictly related to this change (since it is still a problem before this PR)

jonboulle:
@cdaylward @philips ready for re-review

SPEC.md
philips: By lifecycle I mean things like when/if additional processes be launched in a pod and does a pod live for sometime before a process is executed inside of it (for configuring namespaces, etc).

philips:
lgtm

SPEC.md
philips: Possible alternatives to manifest for the image. If we called it one of these then 

- metadata file: "An ACI contains the rootfs and the image metadata file"
- file: "An ACI contains the rootfs and the image file"
- spec: "An ACI contains the rootfs and the image spec"
- struct: "An ACI contains the rootfs and the image struct"

I dislike "metadata file" the least because it is a file with metadata about the image it is part of. So the naming would be Image Metadata Schema and Pod Schema.

alban:
@jonboulle : ok, I wrote https://github.com/appc/spec/issues/231 for uids/gids.

SPEC.md
```diff
@@ -2,33 +2,36 @@
...
-To achieve these goals this specification is split out into a number of smaller sections.
+The specification consists of several key sections:
 
 1. The **[App Container Image](#app-container-image)** defines: how files are assembled together into a single image, verified on download and placed onto disk to be run.
 
-2. The **[App Container Executor](#app-container-executor)** defines: how an app container image on disk is run and the environment it is run inside including cgroups, namespaces and networking.
+2. The **[App Container Executor](#app-container-executor)** defines: how applications are grouped into execution units and the environment they are run inside (including, for example, filesystem layout, resource constraints, and networking).
 
-    * The [Metadata Server](#app-container-metadata-service) defines how a container can introspect and get a cryptographically verifiable identity from the execution environment.
+    * The [Pod Manifest](#pod-manifest) defines how one or more App Container Images are grouped into a deployable, executable unit.
```
cdaylward: A pod template is a producible, transmittable unit. One might be a producer of images and pod templates and that's independent of a particular executor. Therefore, I think the Pod Manifest could be moved out as a top-level concept here.

1. App Container Image
2. App Container Image Discovery
3. Pod
4. App Container Executor

jonboulle: Yes! I agree. I have a branch somewhere that started to do this but I think
I got distracted 😕. Will look when in front of computer.
Am 18.03.2015 17:12 schrieb "Charles D. Aylward" <notifications@github.com>:

jonboulle: Hmm, "App Container Pod" to be consistent? Seems like a bit of a mouthful..
Am 18.03.2015 17:14 schrieb jonathan.boulle@coreos.com:

SPEC.md
```diff
@@ -115,34 +118,62 @@ The dependencies are applied in order and each image dependency can overwrite fi
 Execution details specified in image dependencies are ignored.
 An optional *path whitelist* can be provided, in which case all non-specified files from all dependencies will be omitted in the final, assembled rootfs.
 
-Image Format TODO
+## App Container Executor
```
cdaylward: *waves hands* This seems better to me as coming after the Pods section. Additionally, the section that is titled "Executor Perspective" should just follow on after the introduction of this section (i.e. also remove the "Executor Perspective" subsection)

SPEC.md
cdaylward: Actually, this needs more thought I believe, so perhaps we can add it as part of the "define the lifecycle of the pod" TODO item?

SPEC.md
cdaylward: Still seems out of place to me. Feels like it should be part of the discussion on what how an executor executes a pod but I can leave this one be.

SPEC.md
cdaylward: It does...but then again, no more than App Container Executor. ACI, ACP, ACE ;)

cdaylward:
Beyond my inline comments today which I feel are mostly for clarity, this looks good to me now in terms of the actual definitions and language.

SPEC.md
jonboulle: I am going to look at this in a follow up.

SPEC.md
jonboulle: Again, will consider in aforementioned follow up.

SPEC.md
jonboulle: Yep, I will add that to do the TODO

jonboulle:
Landing. Thanks for the feedback!

PR #63, opened 2015-02-18
add "guidelines/best practices" for implementations

jonboulle:
Many of the details in the app container spec are carefully designed to facilitate optimisations and advanced use cases. However, this is not always immediately obvious from reading the text of the spec. Right now we just give it a passing mention in one section:

> ...implementations could increase performance and de-duplicate data by building on top of overlay filesystems, copy-on-write block devices, or a content-addressed file store. These details are orthogonal to the runtime environment.

Similarly, there are some areas (for example, metrics) which are explicitly not covered by the spec but would benefit greatly from general guidance on how implementations might address them.

We should have a document (or documents) describing suggested guidelines and best practices for implementations of the spec. A few examples to start with:
- Heavy use of content-addressable-stores to cache content. This is useful for both the executor (e.g. at the basic level, Rocket's [cas](https://github.com/coreos/rocket/tree/master/cas)) and for image discovery/retrieval (e.g. services that serve images over HTTP could cache highly used images)
- Use of specific technologies like union mounting (e.g. overlayfs) to optimise the setup of image layouts
- Metrics and monitoring of containers - e.g. in Linux we would anticipate implementations might heavily leverage [cAdvisor](https://github.com/google/cadvisor)

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -1,0 +1,51 @@
+# Implementation Guide
+
+This document contains guidelines, tips and suggested best practices for implementers of the App Container specification.
+
+It should not be considered prescriptive, but is instead a general resource to guide usability and optimization of implementations.
+
+This is a living document; those implementing the spec are encouraged to submit feedback and improvements.
+
+## Storing and working with App Container Images (ACIs)
+
+The canonical ACI format is a simple tarball, which conveniently encapsulates the filesystem and metadata of an image in a single file.
+This also provides a simple, unique and canonical way to reference an image, using the cryptographic hash of this artefact.
+Very simple implementations of the spec might use this tarball format as their primary format for transport and local storage, extracting the tarball as necessary every time an image needs to be accessed.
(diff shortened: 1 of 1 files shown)


GUIDE.md
```diff
@@ -0,0 +1,51 @@
...
+This document contains guidelines, tips and suggested best practices for implementers of the App Container specification.
+
+It should not be considered prescriptive, but is instead a general resource to guide usability and optimization of implementations.
+
+This is a living document; those implementing the spec are encouraged to submit feedback and improvements.
+
+## Storing and working with App Container Images (ACIs)
+
+The canonical ACI format is a simple tarball, which handily encapsulates the filesystem and metadata of an image in a single file.
+This also provides a simple, unique and canonical way to reference an image, using the cryptographic hash of this artefact.
```
philips: artifact?

jonboulle: just let me have _one_!!!

jonboulle:
Ping

cdaylward:
I feel like we have quite a few notes stashed around that can go into this, but I haven't quite had the time to contribute here. LGTM as a starting point.

philips:
lgtm, merging.

PR #209, opened 2015-02-19
schema/types: add tests for acname/ackind

jonboulle:
Started hacking these up a while back but forgot to file a PR. 

There's no good reason for the schema to not have ~100% coverage, so this is
a small step towards that goal.

### DIFF
--- a/schema/types/ackind_test.go
+++ b/schema/types/ackind_test.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"testing"
 )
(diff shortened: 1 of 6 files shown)


schema/types/ackind_test.go
```diff
@@ -43,3 +44,38 @@ func TestACKindMarshalGood(t *testing.T) {
 		}
 	}
 }
+
+func TestACKindUnmarshalBad(t *testing.T) {
+	tests := []string{
+		"garbage",
+		"ImageManifest",
```
cdaylward: shouldn't ImageManifest create non-nill non-empty here?

jonboulle: This is a subtlety which I should probably call out; this is not actually valid JSON, because it is not enclosed by quotes

jonboulle: c.f. \`"AppManifest"\` below

jonboulle: updated

cdaylward: oops, gotcha

schema/types/acname.go
```diff
@@ -11,6 +11,10 @@ var (
 	validACName  = regexp.MustCompile("^[a-z0-9]+([-./][a-z0-9]+)*$")
 	invalidChars = regexp.MustCompile("[^a-z0-9./-]")
 	invalidEdges = regexp.MustCompile("(^[./-]+)|([./-]+$)")
+
+	ErrEmptyACName = ACNameError("ACName cannot be empty")
+	ErrInvalidChar = ACNameError("ACName must contain only lower case " +
+		`alphanumeric characters plus ".", "-", "/"`)
```
cdaylward: Add must begin & end w/ alphanumeric?

cdaylward: Other than the above nitpick, LGTM.

PR #212, opened 2015-02-19
aci: fix ManifestFromImage to handle ./manifest

sgotti:
The tar Header.Name (for example with gnu tar) can start with ./ (for example
./manifest). In this case ManifestFromImage will fail.

Plus, I noticed that the error was strange: unexpected error EOF. This is due
to the io.EOF error not being handled. So I rewrote the code to handle it (now
it's similar to github.com/coreos/rocket/pkg/tar.ExtractTar).

### DIFF
--- a/aci/file.go
+++ b/aci/file.go
@@ -7,11 +7,13 @@ import (
 	"compress/gzip"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


jonboulle:
oops, #189 never landed. Since this is slightly more comprehensive I'm going to merge this instead. Thanks!

PR #215, opened 2015-02-20
spec: Add one-liner that apps in image dependencies are ignored.

### LINKED ISSUE #202: spec: clarify which parts of dependencies are used
The question has come up a couple of times as to how dependent images work exactly: e.g. what if dependent images specify app sections? We should tweak some of the language to be clearer that dependencies are _only_ for filesystems and do not bring in dependent `app`s.

/cc @cdaylward

cdaylward:
Fixes #202

Short and sweet enough?


jonboulle:
@cdaylward if this makes it clear enough for you it's fine with me...

philips:
lgtm

PR #216, opened 2015-02-20
pkg: add an acirenderer library.

sgotti:
As discussed in coreos/rocket#464 I'm opening a PR here.

This library renders an aci given a well defined list of dependencies.
A future patch will also provide the dependency resolving logic.

It adds two interfaces: ACIProvider and ACIRegistry.

The first interface is the one used by this patch and the cas has been changed
to satisfy it.
The second interface will be used by the dependency resolution functions. Also
this interface is satisfied by the cas.

### DIFF
--- a/pkg/acirenderer/acirenderer.go
+++ b/pkg/acirenderer/acirenderer.go
@@ -1,0 +1,202 @@
+package acirenderer
+
+import (
+	"archive/tar"
+	"crypto/sha512"
+	"fmt"
+	"hash"
+	"io"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+
+	"github.com/appc/spec/schema"
+	"github.com/appc/spec/schema/types"
+)
+
+// An ACIRegistry provides all functions of an ACIProvider plus functions to
+// search for an aci and get its contents
+type ACIRegistry interface {
+	ACIProvider
+	GetImageManifest(key string) (*schema.ImageManifest, error)
+	GetACI(name types.ACName, labels types.Labels) (string, error)
+}
+
+// An ACIProvider provides functions to get an ACI contents, to convert an
+// ACI hash to the key under which the ACI is known to the provider and to resolve an
+// ImageID to the key under which it's known to the provider.
+type ACIProvider interface {
+	// Read the ACI contents stream given the key. Use ResolveKey to
+	// convert an ImageID to the relative provider's key.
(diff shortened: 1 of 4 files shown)


pkg/acirenderer/acirenderer.go
```diff
@@ -0,0 +1,214 @@
+// Copyright 2015 CoreOS, Inc.
```
jonboulle: Could you please drop this? We aren't licensing appc/spec as a CoreOS project (though we should figure out the license header soon, @philips )

pkg/acirenderer/acirenderer.go
```diff
@@ -0,0 +1,214 @@
...
+
+// An Image contains the ImageManifest, the ACIProvider's key and its Level in
+// the dependency tree.
+type Image struct {
+	Im    *schema.ImageManifest
+	Key   string
+	Level uint16
+}
+
+// An ordered slice made of Image structs. Represents a flat dependency tree.
```
jonboulle: Images encapsulates an ordered slice of Image structs ...

pkg/acirenderer/acirenderer.go
```diff
@@ -0,0 +1,214 @@
...
+// ACIFiles represents which files to extract for every ACI
+type ACIFiles struct {
+	Key     string
+	FileMap map[string]struct{}
+}
+
+// RenderedACI is an (ordered) slice of ACIFiles
+type RenderedACI []*ACIFiles
+
+// Return the RenderedACI list. All file outside rootfs are excluded (at the
```
jonboulle: GetRenderedACIFromList returns the ..

jonboulle:
LGTM, but another pair of eyes would be good.
@cdaylward @philips @iaguis

pkg/acirenderer/acirenderer.go
sgotti: Done!

pkg/acirenderer/acirenderer.go
sgotti: Looks better :smile:

pkg/acirenderer/acirenderer.go
sgotti: Done for all the functions.

iaguis:
After fixing a couple of nits LGTM

philips:
lgtm, merging.

PR #219, opened 2015-03-04
schema: introduce appOverride to the CRM RuntimeApp

vcaputo:
When the ACE has need for amending the app defaults in any way, it
duplicates the original app object from the image manifest into the
appOverride making amendments to the CRM copy.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -569,34 +569,70 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming 
     "uuid": "6733C088-A507-4694-AABF-EDBE4FC5266F",
     "apps": [
         {
-            "app": "example.com/reduce-worker-1.0.0",
-            "imageID": "sha512-...",
+            "name": "reduce-worker",
+            "image": {
+                "name": "example.com/reduce-worker",
+                "id": "sha512-...",
+                "labels": [
+                    {
+                        "name":  "version",
+                        "value": "1.0.0"
+                    }
+                ]
+            },
+            "app": {
+                "exec": [
+                    "/bin/reduce-worker",
+                    "--debug=true",
+                    "--data-dir=/mnt/foo"
+                ],
+                "group": "0",
+                "user": "0",
+                "mountPoints": [
+                    {
+                        "name": "work",
+                        "path": "/mnt/foo"
+                    }
+                ]
+            },
             "mounts": [
-                 {"volume": "work", "mountPoint": "work"}
+                {"volume": "work", "mountPoint": "work"}
             ]
         },
         {
-            "app": "example.com/worker-backup-1.0.0",
-            "imageID": "sha512-...",
-            "isolators": [
-                {
-                    "name": "memory/limit",
-                    "value": "1G"
-                }
-            ],
-            "annotations": [
-                {
-                    "name": "foo",
-                    "value": "baz"
-                }
-            ],
-            "mounts": [
-                 {"volume": "work", "mountPoint": "backup"}
-            ]
+            "name": "backup",
+            "image": {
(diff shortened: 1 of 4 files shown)


philips:
@vcaputo This cleans up the inconsistency problems but makes it a bit harder to make the CRM a static  thing that could be checked into git for example as described in #207. I am not in love with the name, perhaps we just call it app and do the rename as suggested here: https://github.com/appc/spec/pull/207/files#r25831517

philips:
cc @jonboulle @cdaylward

philips:
@vcaputo How do you feel about overriding fields that are seen here instead of replacing the whole original app? For example @alban brings up a good point in https://github.com/appc/spec/pull/207#issuecomment-77173130 that a user may want to have two things running as the same uid/gid for sharing of files. In that case it seems OK to just override those two fields from the app.

vcaputo:
I prefer just substituting the whole app object when the image-resident app object is unsatisfactory as-is.

You seed the replacement using the original and modify whatever was undesirable.  It's like a copy-on-write approach to the app's execution characteristics.

This approach addresses the needs of supporting overrides without getting into a discussion of merge strategies or precedence issues.  That stuff is just implementation policy for whoever/whatever tool is deriving a CRM from a UI and set of app images.

I'm also unclear on how this precludes storing CRMs in git, what's the issue with versioning a CRM containing overrides?  It seemed like an *advantage* that a portable CRM could encompass the necessary deviations from the constituent app's defaults.

The ACE is of course still free to override a CRM's overrides as well.

I think the CRM is incomplete if it can't characterize image execution which deviates from the image defaults.  It seems like a perfectly normal scenario for a CRM to be constructed which refers to images containing the required bits while entirely disregarding their image manifest defaults.

Sure it doesn't make sense to store some machine-specific overrides in git, so don't do that.  Persist a CRM without overrides, or persist a CRM with the necessary overrides and the rest identical to what the image manifest would have delivered anyways.  Machine-specific runtime overrides can be represented by the CRM equally well with this change if the ACE supports it, but I see this first and foremost as a means of overriding the image defaults with CRM-context-appropriate settings.

Perhaps I'm missing something...

philips:
@vcaputo I think this is a good compromise:

- [ ] fixup the schema to have image as an object with name, id and labels
- [ ] rename appOverrrides to app and then we can initially say it is a full copy

We can define a finer grained "merge" later.

philips:
@vcaputo Would you mind updating this PR to do that? Then we can get this going :)

vcaputo:
@philips yeah

SPEC.md
```diff
@@ -569,34 +569,70 @@ JSON Schema for the Container Runtime Manifest (container manifest), conforming
...
+                        "name":  "version",
+                        "value": "1.0.0"
+                    }
+                ]
+            },
+            "app": {
+                "exec": [
+                    "/bin/reduce-worker",
+                    "--debug=true",
+                    "--data-dir=/mnt/foo"
```
philips: Any idea why this = sign looks weird?

vcaputo: weird how? like out of place because you expect them to be space separated?

Or weird as in rendered weirdly?

If it's the latter, this is just one of those examples of why anti-aliased text isn't really well-suited for small font sizes unless your DPI is very high.  If your environment is using anti-aliasing then two perfectly horizontal bars of 100% opacity pixels will stand out in a sea of fuzzy italicized text having anti-aliased edges, since the equals sign can often be perfectly rendered on a pixel array while the rest are misaligned fractional pixels.

philips:
lgtm, I apologize to @jonboulle for the merge conflict this will cause on #207

jonboulle:
:-(((

*insert scowl*

PR #222, opened 2015-03-07
schema: add resource type name constants

philips:
This was just a stupid oversight in implementing these types. Correct it
so a user has access to the names if needed.


philips:
cc @cdaylward @jonboulle

philips:
actually, I am going to just merge this, it is a trivial code cleanup.

Issue #224, opened 2015-03-09
isolators: include isolator enforcement requirements in manifest


b:
An isolator MAY include a field named "required". If the "required" field is present and has value "true" then an executor MUST either enforce the isolator or refuse to execute the container and return an error. A missing "required" field is treated as a value of "false".

b:
Pull request #225

Issue #197, opened 2015-02-12
isolators: add rlimit and linux rlmit extension isolators


philips:
We should add the subset of rlimits that make sense from POSIX as isolators: http://pubs.opengroup.org/onlinepubs/009695399/functions/setrlimit.html

```
rlimit/open-files
rlimit/new-file-size
rlimit/cpu-limit
rlimit/core-size
```

We should probably consider adding the Linux specific ones too. See `ulimit -a`

```
core file size
data seg size
scheduling priority
file size
pending signals
max locked memory
max memory size
open files
pipe size
POSIX message queues
real-time priority
stack size
cpu time
max user processes
virtual memory
file locks
```

b:
Why not include all the POXIS rlimit resources?

The spec does not currently distinguish between mandatory and optional isolators. Is that intentional?

jonboulle:
> Why not include all the POXIS rlimit resources?

We could consider it, but there's currently no intention for the spec to approximate POSIX

> The spec does not currently distinguish between mandatory and optional isolators. Is that intentional?

Does this section not address it?
```
An executor MAY ignore isolators that it does not understand and run the container without them.
But, an executor MUST make information about which isolators were ignored, enforced or modified available to the user. 
An executor MAY implement a "strict mode" where an image cannot run unless all isolators are in place.
```

b:
I'm not proposing approximating POSIX, only suggesting that if all systems can be expected to have the POSIX rlimit resources (which they can) and all of them are potentially useful (which they are) why not include them all?

The text only describes what to do if an executor doesn't understand an isolator the container wants. Specifying mandatory isolators provides a contract on which developers can rely for what isolators are guaranteed to be available. As I said in another issue, the strict mode question should be up to the container, not the executor.

philips:
@b Sorry, this bug isn't very clear. What I am saying is we should add the posix ones as `posix/rlimits/open-files` and the non-posix linux ones as `os/linux/rlmits/foobar`.

b:
+1 on that!

Sorry for misunderstanding!

PR #225, opened 2015-03-09
#224 - Add field to specify whether an isolator is required for execution

b:


philips:
I think having this is reasonable. Letting the image say what it expects seems reasonable for a number of the isolators.

What required means for the resource isolators is more nebulous however.

/cc @cdaylward @jonboulle

b:
For things that have a minimum requested allocation, like resource/cpu and resource/memory, it seems that either required is implicit or should be available. For the other currently specified resources, I agree required may not be applicable.

jonboulle:
I can't say I'm convinced by this change. "Letting the image say what it expects" is different from the spec enforcing that implementations MUST fail execution if isolators can't be satisfied. When are isolators not a runtime-controlled characteristic?

In #197 you wrote:
> Specifying mandatory isolators provides a contract on which developers can rely for what isolators are guaranteed to be available. 

Do you have examples here? Are you thinking of capabilities? Since right now all the other isolators are just resource, and I do not see why these should be necessarily prescribed in an image. (However I have been meaning to revisit the idea of introducing language regarding "reservations" and "limits" to the resource isolators which might clarify this a bit).

jonboulle:
Oops, missed this: 
> For things that have a minimum requested allocation, like resource/cpu and resource/memory, it seems that either required is implicit or should be available

My previous suggestion re: reservations/limit would help here.

jonboulle:
gaaah, I see language to this effect _has_ been added to the spec. sorry, still catching up on things while I was out...

b:
Correct, this is a backwards-compatible change.  If you leave out the "required" field everything works as it does today.

jonboulle:
@cdaylward how are those isolator thoughts coming along? ;-)

PR #226, opened 2015-03-10
SPEC: use the "tag" label instead of the "version" label

philips:
I was thinking on this tonight and I keep going back and forth. Putting
this up and sleeping on it.


jonboulle:
Explain your rationale in a little more detail?

philips:
The rational is that people are familiar with using the term "tag" from git. However, it makes saying it outloud really annoying though: "I have the tag label set to version 1.0.0". I am going to just close this.

PR #227, opened 2015-03-10
SPEC: tweak resource isolator wording

jonboulle:
- Executors MUST act on limits, instead of MAY.
- Move to one-sentence-per-line and add some keyword highlighting


philips:
lgtm

PR #228, opened 2015-03-10
SPEC: specify only minimum set of environment vars

jonboulle:
It is useful for executor implementations to provide various environment 
variables to applications, so the specification should only mandate certain
values rather than the complete set.


philips:
lgtm

PR #234, opened 2015-03-11
schema: marshalling of isolators does not work

jonboulle:
Build a valid v0.4.0 image manifest with the following isolator using actool v0.4.0:

```
{
    "name": "os/linux/capabilities-retain-set",
    "value": {
        "set": ["CAP_NET_ADMIN"]
    }
}
```

You will get the following error message:

```
build: Unable to close image workspace/aci/kube-proxy.aci: json: error calling MarshalJSON for type *types.App: json: error calling MarshalJSON for type json.RawMessage: unexpected end of JSON input
```

### DIFF
--- a/schema/types/isolator.go
+++ b/schema/types/isolator.go
@@ -50,8 +50,8 @@ type IsolatorValue interface {
 	AssertValid() error
 }
 type Isolator struct {
-	Name     ACName          `json:"name"`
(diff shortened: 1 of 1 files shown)


bcwaldon:
Patch works for me. I expect testing to come when we refactor this per #235

PR #237, opened 2015-03-12
Iaguis/fix validation

### LINKED ISSUE #236: layout validation gets confused by "manifest" and/or "rootfs" files under rootfs/
In experimenting with actool, I discovered that if the rootfs contains a file named "manifest" anywhere in the tree, it gets used as the image manifest for validation purposes (and will fail if its not valid JSON).  It will also incorrectly fail if there's a file named "rootfs" anywhere in the tree under rootfs/.

ValidateLayout could return "filepath.SkipDir" for RootfsDir, perhaps?  Unsure how rigorous the check under rootfs/ should be...

iaguis:
We should check only the "manifest" and "rootfs/" in the root path of the layout so [using filepath.Base()](https://github.com/appc/spec/blob/80f1c924c5ca698cb15e2119aeb52245f48370ed/aci/layout.go#L63) is wrong.

Also, fix mismatching documentation.

Fixes #236

### DIFF
--- a/aci/layout.go
+++ b/aci/layout.go
@@ -60,8 +60,7 @@ func ValidateLayout(dir string) error {
 		if err != nil {
 			return err
 		}
-		name := filepath.Base(rpath)
-		switch name {
+		switch rpath {
 		case ".":
 		case ManifestFile:
 			im, err = os.Open(fpath)
--- a/aci/layout.go
+++ b/aci/layout.go
@@ -85,7 +85,7 @@ func ValidateLayout(dir string) error {
 	return validate(imOK, im, rfsOK, flist)
 }
 
-// ValidateLayout takes a *tar.Reader and validates that the layout of the
+// ValidateArchive takes a *tar.Reader and validates that the layout of the
 // filesystem the reader encapsulates matches that expected by the
 // Application Container Image format.  If any errors are encountered during
 // the validation, it will abort and return the first one.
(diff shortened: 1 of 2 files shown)


alban:
@iaguis : does it work when the aci contain an entry for "./rootfs" instead of "rootfs" (and "./manifest" instead of "manifest")?

``actool`` builds the aci without the "./" prefix.

But if I create the aci with the command ``tar cf container.aci ./rootfs ./manifest`` instead of ``tar cf container.aci rootfs manifest``, I would like if it worked too...

alban:
Nevermind, it works fine because of the filepath.Rel() call above.

iaguis:
:+1:

jonboulle:
Would you mind adding a test case for the different possibilities? (just `./manifest` and `manifest` right?)

OR we could mandate in the spec that paths must not start with `./`; thoughts @philips @cdaylward ?

jheiss:
I'd vote for supporting (and adding tests for) supporting ./ so that folks have more flexibility on how they build their images.

iaguis:
We are already testing `./manifest` and `manifest` for acis: [file_test.go](https://github.com/appc/spec/blob/a6474a8bea3b1cc2a9409494d71f7844037ba338/aci/file_test.go#L11). Actually, I'm fixing `ValidateLayout`, not `ValidateArchive`, ACIs already work fine.

I added a new test checking that a layout with another `manifest` and `rootfs` directory in its `rootfs/` validates.

jonboulle:
Oh ok, great. Thanks for the patch! :+1:

Issue #238, opened 2015-03-12
BuildWalker gets stuck on fifo files in rootfs


dardok:
aci/build.go should probably not try to open os.ModeNamedPipe (or os.ModeSocket?)

dardok:
Looks like go/tar may not be able to extract fifo types:

# rkt prepare ...
"error expanding app image: error extracting ACI: error extracting tarball: unsupported type: 54"

Investigating...

dardok:
Its an unimplemented type in rocket/pkg/tar/tar.go.  Opening issue over there...

jonboulle:
Is there still something that needs to be fixed in the spec code or is it just on the Rocket side?

jonboulle:
never mind, found #240

jonboulle:
fixed by #240

PR #240, opened 2015-03-12
actool: don't try to open named pipes or sockets

dardok:
actool will hang if fifos are opened during build.  sockets probably shouldn't be opened either.

### DIFF
--- a/aci/build.go
+++ b/aci/build.go
@@ -35,6 +35,9 @@ func BuildWalker(root string, aw ArchiveWriter) filepath.WalkFunc {
 		link := ""
 		var r io.Reader
 		switch info.Mode() & os.ModeType {
+		case os.ModeSocket:
+			return nil
+		case os.ModeNamedPipe:
 		case os.ModeCharDevice:
 		case os.ModeDevice:
 		case os.ModeDir:


alban:
LGTM

jonboulle:
fixes #238

jonboulle:
Change LGTM, but could you please rebase to tweak your commit message?
"actool: don't try to open named pipes or sockets"

dardok:
@jonboulle done

alban:
I tried to build an aci containing a socket file with ``actool build``. Before applying this patch, it failed with the following error:
```
build: Error walking rootfs: open bin/image-aci/rootfs/my.sock: no such device or address
```

After applying this patch, the aci creation succeed. The tar archive contains an empty regular file instead of the socket. I think ``actool build`` should skip the socket file and print a warning instead of silently add an empty regular file. That's what ``GNU tar`` does when the user requests to archive a socket file:
```
$ tar cvf sock.tar my.sock 
tar: my.sock: socket ignored
```

The tar format supports fifos but does not support sockets. [FileInfoHeader](http://golang.org/src/archive/tar/common.go?s=5722:5787#L238) in the tar package (inadvertently?) skips the initialization of the Typeflag field in the tar header, so it stays zero.

According to the [Basic Tar Format](http://www.gnu.org/software/tar/manual/html_node/Standard.html), the value zero in the Typeflag means it is a regular file.
```
/* Values used in typeflag field.  */
#define REGTYPE  '0'            /* regular file */
#define AREGTYPE '\0'           /* regular file */
```

jonboulle:
Hmm, yes, you're totally right, we should just ignore socket files. @dardok would you mind updating? (or just drop sockets from this commit and can do in a follow up)

jonboulle:
Thanks, lgtm!

PR #241, opened 2015-03-12
discovery: use proxy defined by environment variables

alban:


jonboulle:
LGTM, but any more context?

slallema:
@jonboulle : it is a necessary fix to take environment variables into account when you use rocket behind a http proxy like me ;)

alban:
The useful environment variables for proxies are $HTTP_PROXY and $HTTPS_PROXY.

PR #247, opened 2015-03-19
schema/types: add escaped types.ACName helpers

vcaputo:
EscapedNewACName(string) (*ACName, error)
(*ACName) EscapedString() (string)

It's sometimes convenient to use an ACName as a name in the filesystem, but
the ACName permitting "/" prohibits direct use.  Rather than everyone
perform their own escaping, provide escape/unescape helpers for
filename-safe names.

This implementation simply uses "_" to represent "/".

### DIFF
--- a/schema/types/acname.go
+++ b/schema/types/acname.go
@@ -119,3 +119,29 @@ func SanitizeACName(s string) (string, error) {
 
 	return s, nil
 }
+
+// EscapedString produces an escaped string form of the name n safe for use as a posix filesystem name
+//
+// This is a helper function and its algorithm is not part of the spec.
+func (n *ACName) EscapedString() string {
+	return escape(n.String())
+}
+
+// EscapedNewACName
+// EscapedNewACName generates a new ACName from a string escaped using EscapedString().
+// If the unescaped string is not a valid ACName, nil and an error are returned.
+//
+// This is a helper function and its algorithm is not part of the spec.
+func EscapedNewACName(s string) (*ACName, error) {
+	return NewACName(unescape(s))
+}
+
+// The only thing unsafe in an ACName for use in a posix filesystem name is the forward slash,
+// simply substitute it with a posix-safe, shell-and-systemd-benign, and ACName-prohibited _.
+func escape(s string) string {
+	return strings.Replace(s, `/`, `_`, -1)
+}
+
+func unescape(s string) string {
+	return strings.Replace(s, `_`, `/`, -1)
+}


alban:
The patch looks good to me. Tests could be added in ``schema/types/acname_test.go``

schema/types/acname.go
```diff
@@ -119,3 +119,29 @@ func SanitizeACName(s string) (string, error) {
...
 }
+
+// EscapedString produces an escaped string form of the name n safe for use as a posix filesystem name
+//
+// This is a helper function and its algorithm is not part of the spec.
+func (n *ACName) EscapedString() string {
+	return escape(n.String())
+}
+
+// EscapedNewACName
```
philips: delete?

philips:
lgtm besides the comment nit. A tiny test would be nice to see too.

cdaylward:
One of the valuable principles behind the ACName is that it is a namespace. For instance, image names are part of the DNS namespace. The '/' in the name provides namespacing that maps directly into URLs and POSIX pathnames alike (also HDFS, etc). This change seems to undo this (as it is both not collision-free and not a reversible process). Use of '/' in the ACName does not prohibit direct use; the name is POSIX-safe. Arguably, it also encourages better general behavior in implementations. Namely, that a large collection of images are organized on disk in a way that reflects the namespace. I'm not sure why we want this included with the specification tooling as a matter of course.

jonboulle:
I agree; it seems awkward to put this in the schema code, and that if needed it would be better located in a helper package more local to its use. (I would argue that `SanitizeACName` is not a precedent because it produces names that conform to the spec).

philips:
Hrm, these arguments make sense to me. Perhaps we just have a utils package to help out with this stuff?

@cdaylward One problem however is you don't know how deep into the hierarchy you should go and things like example.com/foo and example.com/foo/bar both being images is troublesome if you don't flatten the image names in a hierarchical FS.

cdaylward:
@philips example.com/foo and example.com/foo/bar are pre-resolution app names. After discovery, or otherwise at storage/cache time, these are something like example.com/foo-1.0.0-linux-noarch.aci and example.com/foo/bar-1.0.0-linux-noarch.aci. I'd argue that is perhaps a poor namespace on the creator's part, but still completely valid and storable (collision free).

There are already many implicit implementation requirements. For instance, an implementation should probably detect _absurdly_ long application names and things like '..'. I'm not sure escaping to another valid character alleviates these types of concerns. It also introduces a new problem in that app names foo/bar and foo_bar would collide).

vcaputo:
I agree with everything being said here.

The result however is that I'm now questioning the choice of type used for RuntimeApp.Name.  Should this be something other than an ACName?  e.g. AppName having looser restrictions and no additional significance WRT name resolution and namespaces, instead serving only to uniquely identify an app within a pod?  The ACName for the image is readily available via RuntimeApp.Image.Name.

I propose we add AppName, loosen its character restrictions relative to ACName while still permitting it to contain an ACName, and give it convenient helper functions for (un)escaping filename/systemd service name usage.  Use this new type for RuntimeApp.Name.

This would serve to further  clarify the distinction between RuntimeApp.Name and RuntimeImage.Name.

@cdaylward @philips @jonboulle thoughts?

philips:
I think it makes sense to add some helpers to generate the filename and systemd naming but how do you plan to keep the uniqueness constraint for all apps in the pod using these helpers?

For example you get:

```
example.com/foo_bar
example.com/foo/bar
```

I would assume you want a helper that is like:

```
# takes a pod and returns a map from app instance name to systemd unit file name
pod.SystemdNames() map[ACName]string
```

Thinking outloud of the use cases here.

@vcaputo Also, why is loosening the character restrictions necessary?

vcaputo:
@philips the substitution character used by this PR is `_` because `_` is disallowed for an ACName:
> $ bin/rkt run example.com/foo_bar=imgs/pauser.aci
run: error parsing apps: name "example.com/foo_bar" invalid in "example.com/foo_bar=imgs/pauser.aci": ACName must contain only lower case alphanumeric characters plus ".", "-", "/"

Loosening up the restriction is useful because `RuntimeAppName` is potentially just some contextual unique label assigned by the user to an app within a given pod, `ACName` restrictions are quite strict for this usage.

jonboulle:
Hmm, then it sounds like we have _three_ types: `ACName`, `RuntimeAppName`, and `EscapedAppName` - which seems to me to be overcomplicating things a bit.

I agree in principle that `ACName` isn't strictly necessary but I'm not sure I'm convinced that adding another type is preferable to just utilising helper functions in implementations (like the aforementioned hypothetical `SystemdNames()`). @vcaputo is your concern that it will be too difficult (or repetitive) for implementations to handle this conversion while maintaining uniqueness, and hence that it would be better to codify it upstream in the spec?

philips:
@jonboulle @vcaputo's concern is that users will want to specify names with underscores, colons, etc because those are common patterns for naming things.

philips:
One reason for sticking to ACName for this field is we want it to be a valid URL path for the metadata service too.

At this point I am OK with keeping the restrictions on this field to ACName even though it isn't a perfect match for the use case. Then @vcaputo can add a utility escaping method to `github.com/appc/spec/escape` or something. 

In the future if we need to expand the naming because of a user request we can. It seems like a distraction at this point to choose the perfect subset of characters to allow/disallow when we just need this basic escaping functionality to keep moving forward.

@vcaputo Can you choose an escaping algorithm that would allow for more loose restrictions in the future? Then we can add it to an escape package or something?

Seem OK?

/cc @robszumski Do you think it will be troublesome that users will only be able to name running apps things like `hello-world` or `myapp.com/hello-world` but not `hello_world` or `hello:world`?

vcaputo:
@jonboulle `ACName` makes sense for discovery resolution, in contexts where discovery is irrelevant `ACName` is unjustifiably restrictive.

As for providing escaping helpers it just seemed like functionality most ACE implementations are going to need, and if we can promote consistency and code reuse at the same time with a few convenient helper methods, sounds good to me.

Helpers or no helpers, it seems appropriate for the `RuntimeAppName` to be distinguished from the `ACName` and arguably loosened up in its restrictions.  The default single image case may automatically derive the `RuntimeAppName` from the image's `ACName`, but in the other cases where the `RuntimeAppName` is being set by the user I'd rather not have the ACE telling them their not allowed to use `_`, `:`, or `@`, to name a few common ones, for seemingly baseless/arbitrary reasons.

robszumski:
@philips I'm not super worried about `:` but not being able to use `_` seems like it would be a restriction that many people would hit.

philips:
I don't have a strong feeling, if @jonboulle is OK with loosening up the restrictions of the Pod app name field I am OK too. We will just have to be aware of URL encoding the name in the metadata service. (cc @eyakubovich)

@vcaputo what escaping are you going to use?

mpasternacki:
Hm. ACNames map nicely into URL fragments, but I'd be triple-wary about using them as filesystem paths without *heavy* sanitization or escaping:

1. Nesting problems: having example/foo and example/foo/bar can be easily solved by suffixing: ./example/foo.suffix, ./example/foo/bar.suffix. But what if it's example/foo and example/foo.suffix/bar (not necessarily with good intentions)? It needs more sanitizing than just slapping a suffix at the end.
2. Possible absurd path length (this bites me in particular because of [FreeBSD's silly zfs filesystem / mountpoint length limit of 88 characters](https://lists.freebsd.org/pipermail/freebsd-fs/2010-March/007964.html)); need to create intermediate nested ZFS filesystems or have ZFS filesystem name diverge from its mountpoint
3. Path traversal (`.` and `..` fragments), would need validation/sanitization/quoting again
4. Far-fetched, but still something to think of: this may make some symlink attacks possible, especially combined with the issues above and the fact that absolute symlinks *inside* an image will be seen as absolute symlinks to host path by the runtime environment.

FWIW, in Jetpack I only use fixed strings, UUIDs, and integers as pathnames (e.g. a pod's third app's rootfs would be at `pods/UUID/rootfs/3/`; images also have generated UUIDs, due to SHA-512 being too long – ridiculous name length limit again). Mapping these to pretty names is part of the UI, and the UI knows the manifest. There are simply too many things to worry about when using external input as a filename.

Side Note: check, what happens when unpacking an ACI that includes the following entries, in the following order (possibly including intermediate directories):

    rootfs/toorfs -> /
    rootfs/toorfs/root/.ssh/authorized_keys

vcaputo:
Let's abandon this for now and see where things head

PR #248, opened 2015-03-19
SPEC: promote Pods to top-level concept

jonboulle:
Following some of the discussion in #207, this is an attempt to restructure
the spec by promoting the Pod to a top-level concept (along with ACIs, image
discovery, and ACE). Decoupling them from the executor section makes it
clearer that producing/manipulating pods can be completely independent of
execution.


jonboulle:
@cdaylward first cut, could do with some more tweaking but should be a little easier to review than the previous.

philips:
lgtm

cdaylward:
Looks good as it is.

PR #249, opened 2015-03-19
schema: replace bool with *bool in types.Volume.

yifan-gu:
Fix https://github.com/coreos/rocket/issues/615

By default the `readOnly` field in Volume is `false`, this will cause
an app container runtime to always override the image manifest's volume
mounts setting. Change it to `*bool` gives us the ability to ignore this
field in the container runtime manifest.

### DIFF
--- a/schema/types/volume.go
+++ b/schema/types/volume.go
@@ -19,7 +19,7 @@ type Volume struct {
 	// currently used only by "host"
 	// TODO(jonboulle): factor out?
 	Source   string `json:"source,omitempty"`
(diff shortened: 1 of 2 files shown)


jonboulle:
Can you please add more context in the commit/PR message about why this is required? This shouldn't be (well, isn't) a Rocket-specific fix.

yifan-gu:
Sorry :( How does that look now? @jonboulle

jonboulle:
@yifan-gu looks perfect! Thanks!

jonboulle:
LGTM

PR #251, opened 2015-03-19
spec: enforce unique app names within a pod

### LINKED ISSUE #84: spec: clarify relationship between CRM and "apps"

jonboulle:
Since 9aafd97 landed, app references within a pod gained a name field that is
distinct from the name of the underlying image. However, the spec did not
address whether this field needed to be unique, which made it unclear as to
how to facilitate e.g. multiple apps within a pod using the same image.

This makes the name field a) required, and b) unique within a pod.

Accordingly, apps in a running pod can now be uniquely, globally identified by
the combination of Pod UUID + name.

Fixes #84

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -868,7 +868,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
 * **acVersion** (string, required) represents the version of the schema spec (must be in [semver](http://semver.org/) format)
 * **acKind** (string, required) must be set to "PodManifest"
 * **apps** (list of objects, required) list of apps that will execute inside of this pod. Each app object has the following set of key-value pairs:
-    * **name** (string, optional) name of the app (restricted to AC Name formatting)
+    * **name** (string, required) name of the app (restricted to AC Name formatting). This is used to identify an app within a pod, and hence MUST be unique within the list of apps. This may be different from the name of the underlying image (see below).
     * **image** (object, required) identifiers of the image providing this app
         * **id** (string, required) content hash of the image that this app will execute inside of (must be of the format "type-value", where "type" is "sha512" and value is the hex encoded string of the hash)
         * **name** (string, optional) name of the image (restricted to AC Name formatting)
--- a/SPEC.md
+++ b/SPEC.md
@@ -868,7 +868,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
 * **acVersion** (string, required) represents the version of the schema spec (must be in [semver](http://semver.org/) format)
 * **acKind** (string, required) must be set to "PodManifest"
 * **apps** (list of objects, required) list of apps that will execute inside of this pod. Each app object has the following set of key-value pairs:
-    * **name** (string, required) name of the app (restricted to AC Name formatting). This is used to identify an app within a pod, and hence MUST be unique within the list of apps. This may be different from the name of the underlying image (see below).
+    * **name** (string, required) name of the app (restricted to AC Name formatting). This is used to identify an app within a pod, and hence MUST be unique within the list of apps. This may be different from the name of the referenced image (see below); in this way, a pod can have multiple apps using the same underlying image.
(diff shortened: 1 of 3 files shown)


schema/pod.go
```diff
@@ -100,13 +135,13 @@ type RuntimeApp struct {
 	Name        types.ACName      `json:"name"`
 	Image       RuntimeImage      `json:"image"`
 	App         *types.App        `json:"app,omitempty"`
-	Mounts      []Mount           `json:"mounts"`
-	Annotations types.Annotations `json:"annotations"`
+	Mounts      []Mount           `json:"mounts,omitempty"`
```
jonboulle: We need to go through and audit all these fields to ensure that "optional" and "required" (from the spec) are implemented appropriately here in the schema code.

philips:
lgtm

vcaputo:
l g t m

SPEC.md
```diff
@@ -868,7 +868,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
 * **acVersion** (string, required) represents the version of the schema spec (must be in [semver](http://semver.org/) format)
 * **acKind** (string, required) must be set to "PodManifest"
 * **apps** (list of objects, required) list of apps that will execute inside of this pod. Each app object has the following set of key-value pairs:
-    * **name** (string, optional) name of the app (restricted to AC Name formatting)
+    * **name** (string, required) name of the app (restricted to AC Name formatting). This is used to identify an app within a pod, and hence MUST be unique within the list of apps. This may be different from the name of the underlying image (see below).
```
cdaylward: Is it clear to a new reader why this isn't just the image name? Add comment that you can run multiple instances (apps) using the same image in a pod?

jonboulle: https://github.com/appc/spec/commit/e106602e7646d16b5141715ad464a4b864aa9891

jonboulle:
@cdaylward Landing this. Let me know if you think the wording should be tweaked further..

cdaylward:
@jonboulle No, that looks good.

PR #258, opened 2015-03-23
actool: new subcommands: cat-manifest and patch-manifest

alban:
The manifest can be printed as is: 
```
  $ actool cat-manifest file.aci
```

Or it can be parsed and printed in a pretty form:
```
  $ actool cat-manifest --pretty-print file.aci
```

The ACI can be copied with an alternative manifest:
```
  $ actool patch-manifest --manifest=./manifest file.aci target.aci
  $ mv target.aci file.aci
```

Or this can be done in one operation with --replace:
```
  $ actool patch-manifest --manifest=./manifest --replace file.aci
```

The manifest could be patched without providing the whole manifest file:
```
  $ actool patch-manifest --name=example.com/app \
                          --capability=CAP_NET_ADMIN,CAP_SYS_ADMIN \
                          --replace \
                          file.aci
```

### DIFF
--- a/actool/manifest.go
+++ b/actool/manifest.go
@@ -1,0 +1,387 @@
+package main
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/appc/spec/aci"
+	"github.com/appc/spec/schema"
+	"github.com/appc/spec/schema/types"
+)
+
+var (
+	inputFile  string
+	outputFile string
+
+	patchNocompress   bool
+	patchOverwrite    bool
+	patchReplace      bool
+	patchManifestFile string
+	patchName         string
+	patchExec         string
+	patchUser         string
+	patchGroup        string
+	patchCaps         string
+	patchMounts       string
+
+	catPrettyPrint bool
+
+	cmdPatchManifest = &Command{
+		Name:        "patch-manifest",
+		Description: `Copy an ACI and patch its manifest.`,
+		Summary:     "Copy an ACI and patch its manifest (experimental)",
+		Usage: `
+		  [--manifest=MANIFEST_FILE]
+		  [--name=example.com/app]
+		  [--exec="/app --debug"]
+		  [--user=uid] [--group=gid]
+		  [--capability=CAP_SYS_ADMIN,CAP_NET_ADMIN]
+		  [--mounts=work,path=/opt,readOnly=true[:work2,...]]
+		  [--replace]
+		  INPUT_ACI_FILE
+		  [OUTPUT_ACI_FILE]`,
+		Run: runPatchManifest,
+	}
+	cmdCatManifest = &Command{
+		Name:        "cat-manifest",
+		Description: `Print the manifest from an ACI.`,
+		Summary:     "Print the manifest from an ACI",
+		Usage:       ` [--pretty-print] ACI_FILE`,
+		Run:         runCatManifest,
+	}
+)
+
+func init() {
+	cmdPatchManifest.Flags.BoolVar(&patchOverwrite, "overwrite", false, "Overwrite target file if it already exists")
+	cmdPatchManifest.Flags.BoolVar(&patchNocompress, "no-compression", false, "Do not gzip-compress the produced ACI")
+	cmdPatchManifest.Flags.BoolVar(&patchReplace, "replace", false, "Replace the input file")
+
+	cmdPatchManifest.Flags.StringVar(&patchManifestFile, "manifest", "", "Replace image manifest with this file. Incompatible with other replace options.")
+	cmdPatchManifest.Flags.StringVar(&patchName, "name", "", "Replace name")
+	cmdPatchManifest.Flags.StringVar(&patchExec, "exec", "", "Replace the command line to launch the executable")
+	cmdPatchManifest.Flags.StringVar(&patchUser, "user", "", "Replace user")
+	cmdPatchManifest.Flags.StringVar(&patchGroup, "group", "", "Replace group")
+	cmdPatchManifest.Flags.StringVar(&patchCaps, "capability", "", "Replace capability")
+	cmdPatchManifest.Flags.StringVar(&patchMounts, "mounts", "", "Replace mount points")
+
+	cmdCatManifest.Flags.BoolVar(&catPrettyPrint, "pretty-print", false, "Print with better style")
+}
+
+func getIsolatorStr(name, value string) string {
+	return fmt.Sprintf(`
+                {
+                    "name": "%s",
+                    "value": { %s }
+                }`, name, value)
+}
+
+func patchManifest(im *schema.ImageManifest) error {
+
+	if patchName != "" {
+		name, err := types.NewACName(patchName)
+		if err != nil {
+			return err
+		}
+		im.Name = *name
+	}
+
+	if patchExec != "" {
+		im.App.Exec = strings.Split(patchExec, " ")
+	}
+
+	if patchUser != "" {
+		im.App.User = patchUser
+	}
+	if patchGroup != "" {
+		im.App.Group = patchGroup
+	}
+
+	if patchCaps != "" {
+		app := im.App
+		if app == nil {
+			return fmt.Errorf("no app in the manifest")
+		}
+		isolator := app.Isolators.GetByName(types.LinuxCapabilitiesRetainSetName)
+		if isolator != nil {
+			return fmt.Errorf("isolator already exists")
+		}
+
+		// Instantiate a Isolator with the content specified by the --capability
+		// parameter.
+
+		// TODO: Instead of creating a JSON and then unmarshalling it, the isolator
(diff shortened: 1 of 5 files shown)


alban:
I updated my branch to also support
```
--mounts=tmp,path=/tmp:tmp2,path=/tmp2,readOnly=true:tmp3,path=/tmp3,readOnly=false
```

alban:
Strace reveals that the aci file is written by chunks of 64 bytes and it is quite slow for big images. I have not found how to increase the buffer of the gzip writer.

If increasing the write buffer is not enough, maybe pgzip can help: https://github.com/appc/spec/issues/259

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+		if exit != 0 && !patchOverwrite {
+			os.Remove(fh.Name())
+		}
+	}()
+
+	// Prepare input reader
+
+	input, err := os.Open(inputFile)
+	if err != nil {
+		stderr("patch-manifest: cannot open %s: v", inputFile, err)
```
jonboulle: s/v/%v/

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+		return 1
+	}
+	defer input.Close()
+
+	tr, err := aci.NewCompressedTarReader(input)
+	if err != nil {
+		stderr("patch-manifest: cannot extract %s: v", inputFile, err)
```
jonboulle: s/v/%v/
...etc

jonboulle:
@alban that's annoying. what about uncompressed?

alban:
@jonboulle When using ``actool patch-manifest --no-compression``, it takes 2 seconds. Quite fast for a big image. A lot of the write() calls write 32768 bytes at once, but still 80% of the writes are below 4k. So the small writes might not be the bottleneck.

With the default compression, it takes 32 seconds, >99% of the writes are 64 bytes and only 5 IOs are >16k. But ``actool`` takes all my CPU so the small IOs might not be the problem. ``perf top`` shows the following:
```
  36.42%  actool                              [.] compress/flate.(*compressor).findMatch
  12.73%  actool                              [.] compress/flate.(*compressor).deflate
```

alban:
When using the BestSpeed option, gzip is faster with my big image (14 seconds) but it is still 7x slower than no compression:
```
-               gw = gzip.NewWriter(fh)
+               gw, _ = gzip.NewWriterLevel(fh, gzip.BestSpeed)
```

/bin/gzip and pgzip both takes 13 seconds.

philips:
@alban we already need to patch gzip to do gzran, maybe just bite the bullet and do it there? https://github.com/coreos/gzran

alban:
@philips : Where is gzran used at the moment? I'm not sure how gzran can help: the slow part is compressing the archive, not uncompressing.

@jonboulle : can the gzip optimizations be done after this gets merged?

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+			if err != nil {
+				return fmt.Errorf("cannot parse mount point %q", m)
+			}
+			app.MountPoints = append(app.MountPoints, *mountPoint)
+		}
+	}
+	return nil
+}
+
+// iterate iterates over the tar reader and optionally write in the tar writer if not nil
```
jonboulle: can you call this extractManifest or something and update the docstring pls? "iterate" is a bit vague/misleading

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+					return err
+				}
+			} else if tw != nil {
+				err := tw.WriteHeader(hdr)
+				if err != nil {
+					return err
+				}
+				_, err = io.Copy(tw, tr)
+			}
+			if err != nil {
```
jonboulle: inconsistency nit, weird to handle error here for the second branch but on line 188 for the first branch

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+				return err
+			}
+		default:
+			return fmt.Errorf("error reading tarball: %v", err)
+		}
+	}
+
+	return nil
+
```
jonboulle: remove

schema/types/mountpoint.go
```diff
@@ -1,7 +1,57 @@
...
 	ReadOnly bool   `json:"readOnly,omitempty"`
 }
+
+// MountPointFromString takes a command line mountpoint parameter and returns a mountpoint
+//
+// It is useful for actool patch-manifest --mounts
+//
+// Example mountpoint parameters:
+// 	database,path=/tmp,readOnly=true
+func MountPointFromString(mp string) (*MountPoint, error) {
```
jonboulle: please add a test table for this? I haven't been hardline enough about people adding tests with schema code..

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+	catPrettyPrint bool
+
+	cmdPatchManifest = &Command{
+		Name:        "patch-manifest",
+		Description: `Copy an ACI and patch its manifest.`,
+		Summary:     "Copy an ACI and patch its manifest (experimental)",
+		Usage: `
+		  [--manifest=MANIFEST_FILE]
+		  [--name=example.com/app]
+                  [--capability=CAP_SYS_ADMIN,CAP_NET_ADMIN]
```
jonboulle: indentation is slightly off

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+			return fmt.Errorf("no app in the manifest")
+		}
+		isolator := app.Isolators.GetByName(types.LinuxCapabilitiesRetainSetName)
+		if isolator != nil {
+			return fmt.Errorf("isolator already exists")
+		}
+
+		capsList := strings.Split(patchCaps, ",")
+		caps := fmt.Sprintf(`"set": ["%s"]`,
+			strings.Join(capsList, `", "`))
```
jonboulle: we don't normally linewrap this aggressively do we?

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+		isolatorStr := getIsolatorStr(types.LinuxCapabilitiesRetainSetName,
+			caps)
```
jonboulle: likewise

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+		isolator = &types.Isolator{}
```
jonboulle: wait.. wuh? why are you creating a JSON and then unmarshalling, what about just instantiating the Isolator directly?

actool/manifest.go
```diff
@@ -0,0 +1,364 @@
...
+
+	tr, err := aci.NewCompressedTarReader(input)
+	if err != nil {
+		stderr("cat-manifest: cannot extract %s: %v", inputFile, err)
+		return 1
+	}
+
+	err = iterate(tr, nil, true, nil)
+	if err != nil {
+		stderr("cat-manifest: Unable to read %s: %v", inputFile, err)
```
jonboulle: nit, inconsistent capitalisation, please standardise it across the output
`cat-manifest: cannot extract`
`cat-manifest: Unable to read`

actool/manifest.go
alban: The ``Isolator`` struct is defined in a different package ("types") and the ``IsolatorValue`` field is not exported. The only method initializing the private field is ``UnmarshalJSON``.

The ``IsolatorValue`` for ``LinuxCapabilitiesRetainSet`` also contain a private field without method or constructor to initialize it, except ``UnmarshalJSON``.

I added https://github.com/appc/spec/issues/268 and I will put a TODO comment with a link to it.

schema/types/mountpoint.go
alban: Added in ``schema/types/mountpoint_test.go``

alban:
@jonboulle : thanks for the review! I updated my branch.

alban:
I added --user and --group. That should help for testing https://github.com/coreos/rocket/issues/539

actool/manifest.go
jonboulle: egads, sorry about that.

actool/manifest.go
```diff
@@ -0,0 +1,380 @@
...
+		Run:         runCatManifest,
+	}
+)
+
+func init() {
+	cmdPatchManifest.Flags.BoolVar(&patchOverwrite, "overwrite", false, "Overwrite target file if it already exists")
+	cmdPatchManifest.Flags.BoolVar(&patchNocompress, "no-compression", false, "Do not gzip-compress the produced ACI")
+	cmdPatchManifest.Flags.BoolVar(&patchReplace, "replace", false, "Replace the input file")
+
+	cmdPatchManifest.Flags.StringVar(&patchManifestFile, "manifest", "", "Manifest file")
```
jonboulle: a bit ambiguous, prefer something like `Replace image manifest with this file. Incompatible with other replace options.`

jonboulle:
one tiny thing but LGTM after that

actool/manifest.go
alban: I agree, thanks for the phrasing.

alban:
Branch updated.

alban:
I added `--exec` (/cc @krnowak ) and rebased the branch.

jonboulle:
LGTM, thanks, this is great!

PR #265, opened 2015-03-25
Use types.ACName in parse.App labels name and add a LabelsFromMap function.

sgotti:
The first patch changes the discovery.App labels name from a string to a types.ACNAme to be consistent with types.Label.

The second patch will add an utility method to convert the labels map in discovery.App to a types.Labels slice.

This is useful for future rocket patches to avoid adding additional converting code to it.


jonboulle:
seems reasonable. Thanks!

PR #266, opened 2015-03-25
schema/pod: remove UUID field

jonboulle:
The arrival of pods in #207 removed the UUID field from the PodManifest
(instead requiring that the metadata service expose it), but this was sadly
not reflected in the actual schema code.

/cc @eyakubovich @philips @cdaylward

### DIFF
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -13,7 +13,6 @@ const PodManifestKind = types.ACKind("PodManifest")
 type PodManifest struct {
 	ACVersion   types.SemVer        `json:"acVersion"`
 	ACKind      types.ACKind        `json:"acKind"`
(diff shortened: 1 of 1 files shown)


eyakubovich:
should types.UUID remain? is it used anywhere else?

vcaputo:
lgtm

jonboulle:
@eyakubovich yes, it is used by implementations (e.g. Rocket)

vcaputo:
@jonboulle it might be a bit early in the existence of the spec to be retaining cruft on behalf of immature implementations.  Why not just migrate the uuid type stuff to rkt?

jonboulle:
@vcaputo actually, this _is_ needed in the spec code - https://github.com/appc/spec/pull/267/files#r27176061

PR #274, opened 2015-03-31
types/acname: add MustACName function

iaguis:
This function panics when the given string is not a valid ACName. It's
useful for cases where we're sure the ACName we pass is valid.


jonboulle:
I actually had thought it would be implemented in users of the schema, but I can see it being very widely used so I think it's reasonable to have here.

jonboulle:
LGTM, thanks.

PR #275, opened 2015-03-31
acirenderer: optimize PathWhitelist handling

iaguis:
The current implementation of PathWhitelist is slow when the list of
files is big.

This commit sets the current PathWhitelist outside of the Walk loop to
avoid calling pwlToMap on every file. It also removes a
filepath.Clean() call since the following filepath.Join will call Clean
anyway.

This helps with appc/docker2aci#34


jonboulle:
LGTM, gracias!

Issue #284, opened 2015-04-05
actool: limited to architectures amd64 and i386, cannot build image on and for armhf/armv7l


erikano:
Hello,

Background: Having earlier today tried out `docker` on my ARM single board computer and found it unsatisfactory, I decided to give `rkt` a go. After building `rkt` from [coreos/rkt](https://github.com/coreos/rkt), I built the app container tool `actool` found in this repository (the other tools were not built but that is a separate issue and not one that concerns me at the moment).

Preparing an app container of one of my current projects, when I ran `~/go/src/github.com/appc/spec/bin/actool build django-timelog/ django-timelog.aci`, I got:

    build: Layout failed validation: image manifest validation failed: bad arch "armhf" for linux (must be one of: [amd64 i386])

Is there any practical reason I should be aware of why this restriction is made? In other words, is there anything aside from a hard coded rule on acceptable architectures that would keep me from being able to build the app container on armhf (specifically, armv7l)?

I am running Lubuntu 14.04.2 on the [ODROID-XU3](http://www.hardkernel.com/main/products/prdt_info.php?g_code=G140448267127).

Regards,

Erik Nordstroem

PR #285, opened 2015-04-06
Resolves appc/spec#284 as per comment by philips in coreos/rkt#730.

erikano:


erikano:
I forgot to update the tests. Will do that as well.

erikano:
This second commit only tells the tests to not fail arch "arm" on linux.

As mentioned in #284, the other tools aside from actool don't build.

Let me know what you think -- do you want to merge this or would you require everything to work? Note that if the case is the latter, I may not be able to help since such a task would be too time consuming for me to perform right now.

jonboulle:
let's address in #287

PR #287, opened 2015-04-06
schema: add arm linux as valid arch

philips:
appc/spec#284 as per comment by philips in coreos/rkt#730.

### DIFF
--- a/schema/types/labels.go
+++ b/schema/types/labels.go
@@ -7,7 +7,7 @@ import (
 )
 
 var ValidOSArch = map[string][]string{
-	"linux":   {"amd64", "i386"},
+	"linux":   {"amd64", "i386", "arm"},
 	"freebsd": {"amd64", "i386", "arm"},
 	"darwin":  {"x86_64", "i386"},
 }
--- a/schema/types/labels_test.go
+++ b/schema/types/labels_test.go
@@ -24,9 +24,10 @@ func TestLabels(t *testing.T) {
 			`bad os "OS/360"`,
 		},
 		{
-			`[{"name": "os", "value": "linux"}, {"name": "arch", "value": "arm"}]`,
-			`bad arch "arm" for linux`,
+			`[{"name": "os", "value": "linux"}, {"name": "arch", "value": "pdp11"}]`,
+			`bad arch "pdp11" for linux`,
 		},
+
 		{
 			`[{"name": "name"}]`,
 			`invalid label name: "name"`,


philips:
Deprecates #285

jonboulle:
historical discussion here: https://github.com/appc/spec/issues/6#issuecomment-66525116

philips:
We are going to wait on this one until we can get more guidance on what types of ARM strings we should have here. It gets a little tricky because the ARM architecture has a number of types such as "armel", "armhf" and whatever arm64. We want to make sure we have a list and docs that understand these differences. If anyone knows of an authoritative list please let us know.

erikano:
@philips That is true. In the original issues I opened (appc/spec#284 and coreos/rkt#730), you may note that I was talking about "armhf" and "armv7l". The former ("armhf") is the label Ubuntu and Debian at least are using in their packaging systems. The latter ("armv7l") is what is reported by `uname -m` on the same system -- this ties into the issue raised by @mpasternacki and linked to by @jonboulle above in that it is probably too specific a name to be useful.

Would you be willing to add just "armhf" initially? Or would that be a misnomer in the context of Go, you think?

I might also add that on my ARM computer, I have run the "armhf" Debian in a chroot on "armhf" Ubuntu, so at least to me, "armhf" seems to be a useful label to adopt.

Likewise, on an old Android phone I used to have, "armel" Debian was able to run in a chroot.

Furthermore, as `sudo docker search armhf-` will show you, at least some people are using the label "armhf" for their docker images and while this is not docker, it may be of use to note that.

philips:
@erikano Yea, I want to be able to document our decision. I don't know where the canonical source is for these strings.

cnelson:
Looks like they are just made up by the devs working on the port

https://wiki.debian.org/ArmHardFloatPort

See notes on arrmel vs arm and naming debate 

-- 
cnelson

philips:
We are trying to get a more authoritative answer offline with some folks involved in these projects, stay tuned.

jonboulle:
@philips I think we established that for now we can at least add aarch64,
right? That one seems to be clearly defined and standardised (even if not
yet widely implemented)

philips:
On Thu, Apr 9, 2015 at 6:34 PM, Jonathan Boulle <notifications@github.com>
wrote:

> @philips I think we established that for now we can at least add aarch64,
> right?
>

Yes, we can add aarch64. Maybe even aarch32?!

jonboulle:
deprecated (again) by #304

PR #297, opened 2015-04-16
schema: add linux/aarch64 as valid os/arch

jonboulle:
Following on from discussions in #284 and #287, this adds the `aarch64`
architecture as a valid value for the architecture label when `os` is set to
`linux`.


crawford:
LGTM

philips:
ship it

PR #302, opened 2015-04-21
SPEC.md: "container" env variable

vbatts:
In likeness to other existing container technologies, an enviroment
variable of `container` set to the name of the runtime tool. Prior tool
examples include "lxc", "systemd-nspawn", "lxc-libvirt".

May require a little conversation, but it is a change I think will be ultimately accommodating to folks developing to the the container space.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -212,6 +212,7 @@ The following environment variables MUST be set for each application's main proc
 * **SHELL** login shell of the user
 * **AC_APP_NAME** name of the application, as defined in the image manifest
 * **AC_METADATA_URL** URL where the [metadata service](#app-container-metadata-service) for this pod can be found.
+* **container** name of the [App Container Executor](#app-container-executor) as a concise indicator the environment is in a container (free-form string)
 
 An executor MAY set additional environment variables for the application processes.
 


jonboulle:
@vbatts I think this is useful but underspecified. Can we add more colour around what exactly runtimes should set this to, and how it might be useful?

vbatts:
@jonboulle definitely under-specified. "container=lxc" was for long, what some tools had used as a standard to look for, for whether it was running in a contained environment. There was no specification around this, just normative behavior that folks got used to.
systemd has carried that with their [containerInterface](http://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/). And [detects for contained runtimes](https://github.com/systemd/systemd/blob/master/src/shared/virt.c#L366) to modify behavior specific to be inside containers.
This was in docker for a while, but removed despite folks wanting it to stick around.

vbatts:
rebased

jonboulle:
> This was in docker for a while, but removed despite folks wanting it to stick around.

hrm, link?

vbatts:
first https://github.com/docker/docker/pull/4326, then https://github.com/docker/docker/commit/431d510cae85bc1265c861028dd9751ae95088b2

vbatts:
conversation here too https://github.com/docker/docker/pull/2939

vbatts:
Relevant lxc code point https://github.com/lxc/lxc/blob/master/src/lxc/start.c#L748
And the most common reference point is in their man pages https://linuxcontainers.org/lxc/manpages/man1/lxc-attach.1.html under `--clear-env` and `--keep-env`

jonboulle:
@vbatts I am ambivalent about this for a couple of reasons: 1) it's not immediately apparent what utility this provides or what a sensible value would be for a runtime to set it to (and by implication, the most we could do in ACE validator is check it's non-empty?); 2) in #273 we are considering shrinking the set of prescribed environment variables, and this goes in the opposite direction

vbatts:
@jonboulle I agree with reduction of defaults, but as for a standard that will be implemented to, by container runtimes, and by applications needing to behave differently inside container runtimes, it seems it has a place. Rather than the current scenario "to see if it's lxc check this, and docker check that, and if its XXX check for this other thing". 
For application developers, some may only need `if (strlen(getenv("container")) > 0) {...`, but others may see it fit to react to the value as well.

alban:
I don't have a strong opinion on this, I don't mind having this in the spec.

While it is not in the spec, apps using rkt could use `rkt run --set-env=container=something` or they could check `$AC_APP_NAME`.

If `$container` gets added in the spec, what should rkt do when `--set-env=container=something` is used?

The fact that systemd in stage1 benefits from `$container` being set seems to be an implementation detail of `rkt` using systemd in stage1. In fact, systemd-nspawn already set [$container=systemd-nspawn](http://cgit.freedesktop.org/systemd/systemd/tree/src/nspawn/nspawn.c#n3959) for systemd in stage1. This environment variable is not passed the apps.

Systemd tries to [minimize the use of `$container` to change its behaviour](http://lists.freedesktop.org/archives/systemd-devel/2015-April/031243.html):
> we'll try to avoid conditionalizing on
> the fact wether things are in a container or not, but much rather on
> the actual feature that is missing/different in a container, for
> example a missing capability or such

I think that's a good policy to follow for apps as well.

philips:
I just want to be clear that we define what users who see this container variable should use it for. After talking to Lennart Poettering and @vbatts it seems that we should say the value of the variable should be only for informational purposes and will be a free-form string with possible spaces, etc. The important thing is that it will be set so an application can know if it is being executed inside of a container.

@alban If a user explicitly sets the container env variable from their ACI or at runtime I say it overrides.

jonboulle:
@vbatts ping, want to add more detail as mentioned by philips?

vbatts:
Doing it now

vbatts:
@jonboulle @philips updated. is this more along the lines?

jonboulle:
Seems OK. @philips ?

SPEC.md
```diff
@@ -212,6 +212,7 @@ The following environment variables MUST be set for each application's main proc
 * **SHELL** login shell of the user
 * **AC_APP_NAME** name of the application, as defined in the image manifest
 * **AC_METADATA_URL** URL where the [metadata service](#app-container-metadata-service) for this pod can be found.
+* **container** name of the [App Container Executor](#app-container-executor) as a concise indicator the enviroment is in a container (free-form string)
```
jonboulle: environment

vbatts: fixed

jonboulle:
LGTM

PR #312, opened 2015-04-23
spec: should a default value be specified for readOnly?

lairdshaw:
This is so pedantic I'm not even sure it's worth raising, but the spec currently reads: "readOnly (boolean, optional if kind is "host")". Is it worth stipulating that a default of false applies when readOnly is not present?


jonboulle:
:+1: lgtm, thank you sir

PR #309, opened 2015-04-23
spec: vagueness around empty pathWhitelist directories

lairdshaw:
Beside "pathWhitelist" under the heading, "Image Manifest Schema", the spec currently includes the wording: "Paths that end in slash will ensure the directory is present but empty". This seems a little vague to me. Perhaps it could be reworded to "If a path ends in a slash, then the ACE must check that the directory is empty. If it is not empty, then the ACE must [do something appropriate, perhaps emit a warning message]".


jonboulle:
lgtm, keep them coming! :-)

Issue #278, opened 2015-04-01
spec: target particular devices for resource isolators


jonboulle:
Currently most of the resources have a `default` parameter but mandate that it must be set to true, causing the isolator to apply to all devices of that class (e.g. all network interfaces, all block devices).

The spec should provide a way to target specific devices.

philips:
@jonboulle the problem is I couldn't figure out how the heck the container would know what the device is named in many cases. On top of that the kernel doesn't let you pick and choose for some things. Then after talking with @kelseyhightower and other I decided the user cases were few and punted on the whole thing. To that end I don't think this is necessary for v1.0.

vbatts:
but isolators may not always be device-centric. Is seems they will be file based (cgroups, quotas) and capability based.

thockin:
The way we did this in borg is that the user asks for X MiB of disk space and [Y, Z] devices (spindles).  We then make a scheduling decision and bind that that to concrete device names (I forget if that is Borglet or Borgmaster making that decision). 

For kubernetes we punted on it for now.  I'd suggest the same here.  It's not important enough to get wrong, yet.

PR #307, opened 2015-04-23
spec: devicePaths parameter missing from examples

lairdshaw:
One of the parameters under both "resource/block-bandwidth" and "resource/block-iops" is "devicePaths", however it is not present in the examples, and instead the parameter "default" is present. Perhaps "default" should be replaced with "devicePaths"?

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -289,7 +289,7 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
 * Scope: app/pod
 
 **Parameters:**
-* **devicePaths** the block devices that this limit will be placed on
+* **default** must be set to true and means that this limit applies to all block devices by default
 * **limit** read/write bytes per second
 
 ```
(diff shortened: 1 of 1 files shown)


lairdshaw:
Makes sense, how's this?

jonboulle:
@lairdshaw Technically correct although now it just highlights the awkwardness of this field :smile: (what is the utility of having a required field that must be set to one particular value?). I wonder if we should add a note to clarify that this is in anticipation of devicePaths being introduced at a future date, or similar. Thoughts?

lairdshaw:
Yes, they're awkward. If devicePaths is later introduced, then you have the problem of clashes - what are the semantics if devicePaths is set to a single device *and* default is set true? Or would the spec prohibit that combination? Perhaps it would be better not to have default at all and to signal that semantic through devicePaths (if/when it is introduced) - perhaps via either an empty value or an asterisk. Then we could do away with default altogether here.

jonboulle:
I'm going to land this since it fixes/codifies the situation today, we can follow up in #278

PR #315, opened 2015-04-24
SPEC.md: proposal for numeric UID/GID in App user/group fields

### LINKED ISSUE #231: Specify how to handle usernames and uids in ACIs (and groupnames & gids)
The spec should talk about user ids and group ids in containers.

ACIs contain ownership information about the files in two different forms:
- uid and gid: numbers (mandatory fields)
- username and groupname: strings (optional fields)

When the tarball is extracted, the mapping between uid/gid and username/groupname is lost. The spec does not mention whether to use uid/gid or username/groupname.

An ACI might not contain any /etc/passwd file.

If a container contains one mysql app and one apache app, they might want to run as users mysql and www-data respectively. The files in the apache aci might belong to user ID 33 but unless apache.aci contains /etc/passwd, uid=33 does not mean anything.

This issue was first mentioned in:
- https://github.com/appc/spec/pull/207#issuecomment-77173130
- https://github.com/coreos/rocket/issues/539#issuecomment-77167768

krobertson:
This proposes a clarification of the handling of the "user" and "group" fields
and how the values should be handled within the execution environment.

Existing convention was that a user/group value of a numerical string would be
treated as the UID/GID. POSIX standard allows for a username to be all
numbers.

To account for this, the recommendation is that the user/group fields should
always be resolved with /etc/passwd or /etc/group first. If a valid entry is
not found, then it should treat the user/group string as the UID/GID if the
string is all numerical.

Fixes #231

References:

POSIX standard on usernames:
  http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_431
  http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_278


jonboulle:
@krobertson awesome write-up
/cc @thockin @alban

SPEC.md
```diff
@@ -693,7 +693,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
-    * **user**, **group** (string, required) indicates either the UID/GID or the username/group name the app should run as (freeform string). If the user or group field begins with a "/", the owner and group of the file found at that absolute path inside the rootfs is used as the UID/GID of the process.
+    * **user**, **group** (string, required) indicates either the UID/GID or the username/group name the app should run as (freeform string). If the user or group field begins with a "=" contains all numbers after the equal sign, then the number shall be treated as the numerical UID/GID to use for the process. If the user or group field begins with a "/", the owner and group of the file found at that absolute path inside the rootfs is used as the UID/GID of the process.  This conforms to POSIX standards as well as Windows standards, where usernames are not allowed to contain a slash or equal sign. If the string is a username or group, the uid/gid should be resolved using the containers own `/etc/passwd` or `/etc/group` only if the files exist. Example values for the fields include `root`, `=1000`, or `/usr/bin/ping`.
```
alban: > the uid/gid should be resolved using the containers own `/etc/passwd` or `/etc/group`

If a pod containing two ACIs is executed and each ACI has `/etc/passwd`, should we check `/etc/passwd` only in the ACI related to this manifest, or should we check for `/etc/passwd` in all ACIs of the pod? In case of conflicting information in several `/etc/passwd`, should we just refuse to start the pod?

alban:
Do we care about backward compatibility with old ACIs containing "0" without the "="?
```
  "user": "0",
  "group": "0"
```

SPEC.md
thockin: Why should it matter if it is different across apps in a pod?  I think that, in this regard, containers should be self-contained.  If I want consistent values I will either use numerics or I will ensure /etc/passwd is in sync.

thockin: As an alternative, if you don't mind getting fancy, you can define this as a string-or-int.  If string, look it up in passwd, if int, don't.  I very much dislike ad hoc syntax.

krobertson: My understanding of multiple ACIs in the same pod, they're still isolated at the filesystem level, so should only introspect their own files. If there is any kind of conflict, then that is up the person who deployed or created those ACIs. They're the only one with enough knowledge to properly resolve it.

krobertson:
The backwards compatibility is one thing... @jonboulle have an opinion?

I know our implementation still looks for a simple number, I haven't updated our code yet. But what there is today isn't fully portable.

In ours, I know we could key off the ACI version in the manifest to support backwards handling.

SPEC.md
jonboulle: Yes, `/etc/passwd` should be specific per-app within a pod. I'm happy to leave conflicts across the apps undefined/caveat utilitor for now.

The elephant in the room here is volumes shared between apps...

jonboulle: > As an alternative, if you don't mind getting fancy, you can define this as a string-or-int. If string, look it up in passwd, if int, don't. I very much dislike ad hoc syntax.

you prefer ad hoc semantics? :-). I am also not thrilled about it, although it's a clever solution to supporting the different platform requirements. 

My alternative proposal would be for the spec to be more advisory/restrictive rather than explicitly supporting all these alternatives: i.e. these fields _must_ be alphanumeric (probably following the string-or-int logic), _should_ be platform-suitable (i.e. if you use a numeric on Windows, YMMV).

In the absence of more concrete use cases I am starting to question the utility of keeping this provision; it's convenient but arguably unnecessary:
> If the user or group field begins with a "/", the owner and group of the file found at that absolute path inside the rootfs is used as the UID/GID of the process.

jonboulle:
I am not majorly concerned about backwards compatibility if it means we're implementing the right solution =). As Ken mentioned I think we could address this reasonably in implementations for now.

SPEC.md
krobertson: A username of all integers is possible, but not really probable. Difficult to weigh the trade-offs.

thockin: I imagine that an all numeric username would break lots of things, but a quick test with chown works...

I prefer well-defined, type-centric semantics to stringly-type fields, every day of the week.  Storing numbers in strings has never been a good idea.

Or just do what chown does - look up the string in /etc/passwd, if it exists, use it.  If it does not exist && it parses as an int, use it as an int.

thockin: root@kubernetes-minion-bkez:/home/thockin# grep 1234 /etc/passwd
12345:x:8675:8675::/tmp:/sbin/nologin

root@kubernetes-minion-bkez:/home/thockin# touch f

root@kubernetes-minion-bkez:/home/thockin# chown 12345 f

root@kubernetes-minion-bkez:/home/thockin# ls -ln f
-rw-r--r-- 1 8675 0 0 Apr 28 21:48 f

root@kubernetes-minion-bkez:/home/thockin# chown 1234 f

root@kubernetes-minion-bkez:/home/thockin# ls -ln f
-rw-r--r-- 1 1234 0 0 Apr 28 21:48 f

krobertson: I would be fine with the spec being to attempt from /etc/passwd and if not found, to attempt to parse it as an int.  @alban @jonboulle ?

jonboulle: Sounds reasonable to me.

krobertson:
@thockin @alban @jonboulle 

Updated the PR based on feedback. If it looks good, I can squash and use the message from the 2nd commit, as the nature of this changed.

SPEC.md
```diff
@@ -693,7 +693,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
-    * **user**, **group** (string, required) indicates either the UID/GID or the username/group name the app should run as (freeform string). If the user or group field begins with a "/", the owner and group of the file found at that absolute path inside the rootfs is used as the UID/GID of the process.
+    * **user**, **group** (string, required) indicates either the username/group name or the UID/GID the app should run as (freeform string). The user and group values may be all numbers to indicate a UID/GID, however it is possible on some systems (POSIX) to have usernames that are all numerical. The user and group values should first be resolved using the image's own `/etc/passwd` or `/etc/group`. If no valid matches are found, then if the string is all numerical, it should be converted to an integer and used as the UID/GID. If the user or group field begins with a "/", the owner and group of the file found at that absolute path inside the rootfs is used as the UID/GID of the process. Example values for the fields include `root`, `1000`, or `/usr/bin/ping`.
```
thockin: s/should/will or shall ?

thockin: is the "file found at that absolute path" requirements based or speculative?  That's a weird feature

krobertson: That was one that was already in the spec, but think in an earlier commit, it was mentioned it was kind of speculative.

philips: The absolute path thing is a fairly elegant hack that the systemd [team came up with for systemd.sysusers](http://www.freedesktop.org/software/systemd/man/sysusers.d.html). The idea is that you could change all of the files in a container that are owned by uid 100 to uid 200 and then the system would sort itself out. It is more speculative in appc here than useful so I am OK removing it for now.

philips: A practical use case would be say running two ACIs that want to share a volume and need to be the same uid/gid to do that. So, you could change all of the files of the executable they intend to run to match. Again speculative.

krobertson: There is value in it... its already there, I'd say leave it unless there was a reason to not.

alban: > Why should it matter if it [uid mapping in /etc/passwd] is different across apps in a pod? I think that, in this regard, containers should be self-contained.

@jonboulle mentionned shared volumes between the apps.

Also, communication between apps using Unix sockets (either on a socket file on a shared volume, or an abstract socket since apps in a pod share the same net namespace) can use SCM_CREDENTIALS to share their uid/gid.

Processes in different apps on the same pod are in the same pid namespace, so whether they can kill (or signal) each other depends on whether they have the same uid (on design or coincidentally).

heavenlyhash: What happened to the `=` discussed earlier...?  I like the idea of that as a totally unambiguous path to force uids regardless of host system state.  ("state" being a dirty word, and all.)

jonboulle: @heavenlyhash can you explain further, I don't understand how this relates
to the host system state

krobertson: Agree, no host system state is involved with this. The updates include calling out it should perform look ups with the image's own `/etc/passwd` and `/etc/group`.

heavenlyhash: Sorry, I misspoke and misread!  Drop the word "host" from my comment :)

It still seems a little like "spooky action at a distance" that the filesystem contents can impact the meaning of the config, though.  It "makes sense" in this particular context, but it seems like a good precedent to me to say that the filesystem is off limits for propagating information back to the config.  (The copy attributes from `/usr/bin/ping` example is slightly different; that at least has a clear behavior *branch* specified by the config alone, if that makes sense as an explanation.)

One might want to build config parsers and validators and not need to do other things with the filesystem component, for example.

krobertson: When writing the image manifest, the filesystem should pretty much already be known. If you start command is `/usr/bin/ping`, then you know there is a folder `/usr/bin` and a file `ping`.

You are also specifying who to run the process as. If the image author is specifying to run a process as `1234` and doesn't realize that the filesystem they're referencing in their image has a user named `1234` in its `/etc/passwd`, then they aren't very aware of the contents of the image they're generating.

heavenlyhash: Say I want to write a validator like http://yaml.travis-ci.org/ .  Should I have to upload the *entire* filesystem image for it to parse or explain to config to me?

Here's what I'd like such an example validator to tell me:

- if "uid" contains "/usr/bin/ping", it tells me: "the uid will be set according to the uid of this file on the filesystem!"
- if "uid" contains "username", it tells me: "the uid will be looked up from the username in /etc/passwd on the filesystem"
- if "uid" contains "=1000", it tells me: "the uid will be forcibly set to 1000"

If "uid" contains "1000", it tells me... what, exactly?  "the uid will either be forcibly set to 1000, or it will be looked up"?  I think we have a much stronger, simpler, more comprehensible system if we spec to eliminate "either" in a situation like this.

Ideally (if plain numerics as uids are retained for backcompat), I'd say a validator like this example should see "1000", tell you that it will have *either* one of two behaviors, and flag that yellow as ambiguous, and give you a recommendation to switch to the newer formats that are more explicit.

krobertson: If you are talking about a system to validate a container, then I would say yes, you should upload the entire container, including the filesystem. Otherwise you are describing it validating only on partial state. Say your start command is `start`, but in your filesystem, the file is `start.sh`. It can't verify the start command without the filesystem. Or, say your `start` command isn't in the `$PATH`... it can't verify that without the filesystem.

jonboulle: > s/should/will or shall ?

Still needs this, but other than that SGTM

jonboulle:
@thockin if this LGTY would you like to land it?

krobertson:
Updated and squashed down.

vbatts:
good good. I like this clarity. It matches known behavior, `chown(1)`. 

LGTM

vbatts:
@krobertson thanks and congrats

thockin:
Sorry, I have been OOO last three days (still am, but i was too).

PR #320, opened 2015-04-24
pkg: export Makedev helpers

iaguis:
Move them from tarheader to device package and export go functions. This
allows the use of these functions by other packages.

Add makedev function.

This will be used by the CopyTree function I talk about in [coreos/rkt#693 (comment)](https://github.com/coreos/rkt/issues/693#issuecomment-95625786).

### DIFF
--- a/pkg/device/device.go
+++ b/pkg/device/device.go
@@ -1,0 +1,30 @@
+package device
+
+/*
+#define _BSD_SOURCE
+#define _DEFAULT_SOURCE
+#include <sys/types.h>
+
+unsigned int
+my_major(dev_t dev)
+{
+  return major(dev);
+}
+
+unsigned int
+my_minor(dev_t dev)
+{
+  return minor(dev);
+}
+*/
+import "C"
+
+func Major(rdev uint64) uint {
+	major := C.my_major(C.dev_t(rdev))
+	return uint(major)
+}
+
+func Minor(rdev uint64) uint {
+	minor := C.my_minor(C.dev_t(rdev))
+	return uint(minor)
+}
--- a/pkg/device/device.go
+++ b/pkg/device/device.go
@@ -16,6 +16,12 @@ my_minor(dev_t dev)
 {
   return minor(dev);
 }
+
+dev_t
+my_makedev(unsigned int maj, unsigned int min)
+{
+       return makedev(maj, min);
+}
 */
 import "C"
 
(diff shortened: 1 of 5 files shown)


iaguis:
If I copy `my_{major,minor}` functions to rkt I get a compile error because I'm redefining functions:

```
/var/tmp/go-link-5wNQlM/000003.o: In function `gnu_dev_minor':
/usr/include/sys/sysmacros.h:47: multiple definition of `my_minor'
/var/tmp/go-link-5wNQlM/000000.o:/usr/include/sys/sysmacros.h:47: first defined here
/var/tmp/go-link-5wNQlM/000003.o: In function `gnu_dev_major':
/usr/include/sys/sysmacros.h:41: multiple definition of `my_major'
/var/tmp/go-link-5wNQlM/000000.o:/usr/include/sys/sysmacros.h:41: first defined here
collect2: error: ld returned 1 exit status
/usr/lib/go/pkg/tool/linux_amd64/6l: running gcc failed: unsuccessful exit status 0x100
```

This is the only way I found so I can use them both in spec and in rkt :confused:

iaguis:
Silly me (thanks @alban), I didn't think about simply renaming the C functions to `my_rkt_minor` or something like that, that could be a solution.

Still, I'd like to share the code but I don't know if it belongs here.

pkg/device/device.go
```diff
@@ -0,0 +1,41 @@
+package device
```
vbatts: is this linux/posix only? might need to make build tags for unix and non-unix here

vbatts: further, the C macro only does bit-math. Something eventually easier to just do natively, rather than importing from C...

iaguis: You're right, this is linux only (the manpage says they're "not specified in POSIX.1, but are present on many other systems"). I saw you added major and minor in the first place, did you change your mind from [#164 (comment)](https://github.com/appc/spec/pull/164#issuecomment-72647937) and now think we should implement them natively?

vbatts: I still think it's cleaner to use the macros. Then you won't be responsible if it _ever_ changes. Though this macro is a simple one. 

I would just add build tags for now.

iaguis:
I added build tags, PTAL

pkg/device/device.go
jonboulle: If they are not posix surely they shouldn't be in a _posix file?

jonboulle: ping @vbatts

philips:
lgtm

PR #314, opened 2015-04-23
SPEC.md: reference the spec version where mentioned

vbatts:
closes #313

Issue #313, opened 2015-04-23
versioning of the spec itself


vbatts:
While the spirit of the spec feels like maintaining a forward-compatible spec, it seems there is a need for versioning of the spec itself.

Like actool to check correctness, there will be a need for loose coupling of a new change and the compliance of ACIs to this change. There is no field for this (yet), but could be introduced, and infer that manifests without an `AC_VERSION` (or similar) are less than the initial version that includes versioning.

With a version of the spec then implementors can more clearly pin and implement. and staged their release criteria (as opposed to a moving target)

/cc @jonboulle @philips

jonboulle:
@vbatts what is wrong with

> acVersion (string, required) represents the version of the schema specification that the manifest implements (string, must be in semver format)

in both image and pod manifests?

vbatts:
ohman. I read that as an arbitrary version of the aci. Don't know what the annotation of "version" didn't flag in my brain for that.

jonboulle:
@vbatts can we clarify?

vbatts:
absolutely. I can start a section for it, in addition to the loose ref link
PR I just opened

vbatts:
thinking further on this, it needs its own section, because there should be
a stance on version compatibility. e.g. "I'm building pod manifest from
version 0.x.y and it includes an app that is from from 0.x.y-1, is that
okay?", etc.

jonboulle:
Yes, fair point..

vbatts:
I updated my PR, but will have to address compatibility in a bit

Issue #298, opened 2015-04-20
actool: check/update manifest acVersion when building an ACI


iaguis:
Since the [ImageManifest](https://github.com/appc/spec/blob/master/SPEC.md#image-manifest-schema) includes the spec version, a lot of ACI build scripts will use the current version and forget to update it when the spec gets updated. We could avoid this version rot by being smarter when we build ACIs.

Proposal from [coreos/fleet#1191](https://github.com/coreos/fleet/pull/1191/files#r28578245):

> Maybe `actool build` can check the acVersion in the manifest, change it to the last version and:
>
> * either update other fields if there are incompatibilities between the last version and the specified version
> * or complain (or print a warning?) that the manifest is too old: "acVersion is too told. Please fix your manifest."
> * `actool validate` can print this warning/error

jonboulle:
relates to #313

jonboulle:
@iaguis want to take a stab at this?

iaguis:
Sure. Thinking more about it:

1. We can blindly change the version when building the ACI and hope everything works. Not a fan.
2. We can error out when the manifest has an *old* version (that is, older than actool's). This means we're forcing the user to use the latest version in their manifest, maybe it's too strict. We could just print a warning and build the ACI anyway[1].
3. We can create an add update functionality. This will require additional work every time we update the scheme version and change the manifest format.

The ideal solution would be the last one but is it worth the trouble? Maybe after 1.0.

For now I suggest option 2 with a warning so we encourage aci builders to update their manifests but we don't refuse building the ACI.

[1]: If we do a destructive update to the manifest layout in the future, we'd a way to generate ACIs for old versions of the spec.

iaguis:
I've implemented option 2 in #322

jonboulle:
Fixed by #322

PR #322, opened 2015-04-27
actool: warn if manifest's ACVersion is too old

iaguis:
We print a warning if the ACVersion of our input manifest is older than
the current version. This encourages ACI builders to update their
manifests but still allows them to build ACIs with an old ACVersion.

### DIFF
--- a/schema/types/semver.go
+++ b/schema/types/semver.go
@@ -30,6 +30,21 @@ func NewSemVer(s string) (*SemVer, error) {
 	return &v, nil
 }
 
+func (sv SemVer) LessThanMajor(versionB SemVer) bool {
+	majorA := semver.Version(sv).Major
+	majorB := semver.Version(versionB).Major
+	if majorA < majorB {
+		return true
+	}
+	return false
+}
+
+func (sv SemVer) LessThanExact(versionB SemVer) bool {
+	vA := semver.Version(sv)
+	vB := semver.Version(versionB)
+	return vA.LessThan(vB)
+}
+
 func (sv SemVer) String() string {
 	s := semver.Version(sv)
 	return s.String()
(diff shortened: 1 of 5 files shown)


jonboulle:
tests are failing
/cc @vbatts on this and #298 since it relates to #313 / #314

schema/types/semver.go
```diff
@@ -30,6 +30,12 @@ func NewSemVer(s string) (*SemVer, error) {
 	return &v, nil
 }
 
+func (sv SemVer) LessThan(versionB SemVer) bool {
```
vbatts: Perhaps the default would check against something like `LessThanMajor()`, and have an option for strict checking and have this function called `LessThanExact()` or similar.

jonboulle: Yes seems reasonable

jonboulle:
@iaguis mind coming back to this?

iaguis:
I updated the PR.

Now that we only check the major version perhaps it makes more sense to error out instead of just warn when the version is too old.

jonboulle:
LGTM, thanks

jonboulle:
@vbatts does this tickle your fancy?

vbatts:
So an error if there is a newer major version? I would keep it a warning until the time this tool supports validating more than one major version. I could foresee slow-moving groups having an image that is valid, but not the latest major version, unless this is not a use case you ever expect to occur.

vbatts:
bump

iaguis:
> So an error if there is a newer major version? I would keep it a warning until the time this tool supports validating more than one major version. I could foresee slow-moving groups having an image that is valid, but not the latest major version, unless this is not a use case you ever expect to occur.

Well, it is a warning, I was just asking about erroring out but what's implemented here is only a warning.

vbatts:
alrighty.

jonboulle:
@vbatts ?

vbatts:
LGTM

iaguis:
Should we merge this then?

PR #326, opened 2015-04-28
SPEC: Make UUID generation more vague.

### LINKED ISSUE #324: spec: Clarify issues regarding UUIDs of pods
The spec says:

> UUIDs assigned to pods MUST be unique for a given instance of a metadata service. Hence, 
> implementations of a metadata service shared among executors SHOULD either provide centralized
> allocation of UUIDs or partition the UUID space between individual executors.

Presumably this is because the spec calls out that:

> The executor must assign an RFC4122 UUID to each pod. This UUID is exposed to the pod through 
> the Metadata Service.

However, I didn't see any other place aside from the metadata service that the UUID is exposed.   Why not remove that the executor must generate the UUID, and then you can just say:

> UUIDs assigned to pods MUST be unique for a given instance of a metadata service.

The implementation can then have the metadata service itself generate the UUID if desired.

If I am missing a technical reason to have this generated in the executor -- maybe call it out in the spec?

schancel:
Make spec more vague around the issue of POD UUIDs.  This
allows for generation of UUIDs to occur on the metadata
service itself if desired by the implementation.

Fixes #324

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -159,7 +159,10 @@ This example pod will use a set of three apps:
 
 #### Pod UUID
 
-The executor must assign an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt) to each pod.
+Each pod much be assigned an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt). 
+The UUID serves as a canonical reference to a pod within a given administrative domain.
+In this context, an administrative domain is linked to the scope of the associated metadata service.
+For example, given a metadata service that is federated across a geographical cluster of systems, the pod UUID is uniquely scoped to the same cluster.
 This UUID is exposed to the pod through the [Metadata Service](#app-container-metadata-service).
 
 #### Filesystem Setup
@@ -491,8 +494,7 @@ The ACE must provide a Metadata service on the address given to the applications
 
 Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
 
-[UUIDs](#pod-uuid) assigned to pods MUST be unique for a given instance of a metadata service.
-Hence, implementations of a metadata service shared among executors SHOULD either provide centralized allocation of UUIDs or partition the UUID space between individual executors.
+[UUIDs](#pod-uuid) assigned to pods MUST be unique for the administrative domain of the metadata service.
 
 ### Pod Metadata
 


SPEC.md
```diff
@@ -159,8 +159,7 @@ This example pod will use a set of three apps:
 
 #### Pod UUID
 
-The executor must assign an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt) to each pod.
-This UUID is exposed to the pod through the [Metadata Service](#app-container-metadata-service).
+Each pod much be assigned an [RFC4122 UUID](http://www.ietf.org/rfc/rfc4122.txt). This UUID is exposed to the pod through the [Metadata Service](#app-container-metadata-service).
```
jonboulle: Please retain the sentence-per-line format for now (we prefer it for reviewability).

jonboulle: I think we should err on the side of describing this more verbosely, if not necessarily prescriptively (I am anticipating more questions around the purpose of the UUID) - maybe something like:
"The UUID serves as a canonical reference to a pod within a given administrative domain.
In this context, an administrative domain is linked to the scope of the associated metadata service.
For example, given a metadata service that is federated across a geographical cluster of systems, the pod UUID is uniquely scoped to the same cluster."

..am I making things better or worse?

SPEC.md
```diff
@@ -492,7 +491,6 @@ The ACE must provide a Metadata service on the address given to the applications
 Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
 
 [UUIDs](#pod-uuid) assigned to pods MUST be unique for a given instance of a metadata service.
-Hence, implementations of a metadata service shared among executors SHOULD either provide centralized allocation of UUIDs or partition the UUID space between individual executors.
```
jonboulle: See previous comment about verbosity - I think we should provide examples because questions around the scope/scale of the metadata service have come up before.

schancel:
I agree with your comments -- I was not sure how descriptive to be in technical writing like this. I will make those changes.

jonboulle:
@schancel as a general comment based on my experience so far shepherding the spec - it's definitely better to stray on the side of being _too_ verbose and illustrative (with examples, etc), as when we've stuck to purely technical language people haven't always grasped the subtleties/use cases/etc. We can start to clean this up as we approach 1.0 and move towards splitting the spec out into different sections, adopting a more formal structure, etc etc.

jonboulle:
ping on this

schancel:
I liked your wording, so I used that under the UUID definition.  I didn't want to restate much under the Metadata Service section, so there's only a short sentence and link to the UUID section.

jonboulle:
:+1: thanks!

PR #327, opened 2015-04-28
spec: use bearer token for metadata service auth

### LINKED ISSUE #153: SPEC: propose not requiring private net for metadata-svc use
Having private networking (container having its own IP) is certainly useful for both ensuring port space isolation and ability to firewall traffic between containers. At the same time it brings a lot of complexity to existing infrastructure. For many use cases, the benefits don't outweigh the costs and it's easier to get many of the benefits of containerization while sharing the network stack with the host.

The current spec effectively disallows sharing the host network as it makes it very difficult for metadata service to identify the calling container. I propose the use of bearer token passed in via Authorization header to securely identify a container. The bearer token would be passed in to application via environment variable AC_METADATA_TOKEN.

The spec already requires the use of a custom HTTP header (Metadata-Flavor) so the burden on the users should not greatly increase.

eyakubovich:
Using container's IP for identify is problematic because:
1. It places a burden on the network impl to protect against IP spoofing
2. It prohibits the container from sharing the network with the host.
3. It makes it impossible to have a central metadata service if the
nodes are behind a NAT.

The use of bearer removes these constraints.
The token is passed in via AC_METADATA_TOKEN env var.

Fixes #153

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -495,6 +495,10 @@ The ACE must provide a Metadata service on the address given to the applications
 
 Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
 
+**NOTE**: Implementations can insert an authorization token into AC_METADATA_URL.
+For example, AC_METADATA_URL can be set to `https://10.0.0.1:8888/Y4vFeVZzKM2T9rwkpWHfqXuGsNjS6O5c` with the path portion acting as a token to uniquely and securely identify a pod.
+Such a token SHOULD have no fewer than 128 bits of entropy (i.e. size of UUID).
+
 [UUIDs](#pod-uuid) assigned to pods MUST be unique for the administrative domain of the metadata service.
 
 ### Pod Metadata


schancel:
Token is going to have to be generated by the XE and passed off to the metadata service somehow.  The uniqueness requirements for the Bearer token need to be greater or more than for the UUIDs of Pods.  Is this the RFC that would be relevant? https://tools.ietf.org/html/rfc6750

eyakubovich:
@schancel Why does the token need to be greater than the UUID? UUID, which is 128 bits, could be encrypted by AES (128 bit block) to make a token. The XE could then know the AES key and not have to pass the token to the metadata service. But passing the token is also fine. I'm not a security person so I am curious on why this is insufficient.

schancel:
@eyakubovich Sorry, typo -- fixed.  Just needs to be greater or equal since this is now the secret identity of the container to the metadata service.  128 Bits is definitely fine, but since the spec went to the trouble of calling out the uniqueness requirements of the UUID, I thought it should do the same for the Bearer token.   With all that said, 1 out of 1.469366e27 is probably an acceptable number for a collision :smile:

schancel:
@eyakubovich Also w/ regards to the AES encrypted UUID -- that is possible.  All I mean is there has to be some random generated bits on the XE which gets handed off to the metadata service.  Either UUID + some preshared key or the random bits of the token itself.

schancel:
:+1:

jonboulle:
LGTM. I am slightly wary of adding another environment variable mandate when we've talked about going in the other direction (and making the metadata service optional) but since URL was required anyway we can clear that up in a subsequent change.

eyakubovich:
This would be a separate PR but I'm all for making metadata service optional. So missing/empty AC_METADATA_URL var could indicate no metadata service available.

SPEC.md
```diff
@@ -490,6 +491,10 @@ The App Container specification defines an HTTP-based metadata service for provi
 The ACE must provide a Metadata service on the address given to the applications via the `AC_METADATA_URL` [environment variable](#execution-environment).
 
 Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
+Clients must also authenticate by use of a bearer token.
```
jonboulle: @eyakubovich would it be appropriate to reference https://tools.ietf.org/html/rfc6750 here?

eyakubovich: @jonboulle I think it'll be more confusing than helpful. This does not implement rfc6750. It's not even OAuth. It just uses the same Authorization header format.

jonboulle: @eyakubovich I like to reference existing standards as much as possible when calling out things like "a bearer token"; AIUI that concept is not exactly genericised yet beyond OAuth...

eyakubovich: @jonboulle Ok, I'll link to https://tools.ietf.org/html/rfc6750#section-2.1

eyakubovich: Linked, just few lines below.

SPEC.md
```diff
@@ -490,6 +491,10 @@ The App Container specification defines an HTTP-based metadata service for provi
...
+The bearer token MUST be passed in by the ACE via `AC_METADATA_TOKEN` environment variable.
+Since the bearer token is used to securely identify a pod, it SHOULD have no fewer than 128 bits (size of UUID) bits of entropy.
```
jonboulle: Just realised you repeated repeated a word.

Can we tweak it slightly to this:
> Since the bearer token is used to securely identify a pod, it SHOULD have no fewer than 128 bits of entropy (i.e. size of UUID).

jonboulle:
@eyakubovich one more fix then LGTM....

vbatts:
Might this tokenized auth be optional? if the `$AC_METADATA_URL` returns a 401, then expect to use the `$AC_METADATA_TOKEN`.

eyakubovich:
@vbatts I think it being optional is a good idea. But then how about saving a round trip and just using missing AC_METADATA_TOKEN as a sign that it's not needed.

vbatts:
@eyakubovich another thought (that I did not formalize before you opened this PR) and is not as strong as an oauth handshake. What if the `$AC_METADATA_URL` provided to the container already had a salt loaded into the path of the url? The metadata-service API would remain transparent to the container service, but from the service side, it would be a watermark on the requests from the container. Thoughts?

eyakubovich:
@vbatts I think that's a great idea. Then then spec can just point out that an implementation can use this technique but otherwise it doesn't have to change.

vbatts:
@eyakubovich that was the thought. Then we don't have to introduce more variables and let @jonboulle keep cleaning house ;-)

jonboulle:
well, shoot. I'm pretty sure someone suggested that earlier (maybe @schancel ?) but I guess we forgot about it. @eyakubovich do you plan to rework this then?

eyakubovich:
@jonboulle yeah, I'll rework.

eyakubovich:
@jonboulle PTAL

SPEC.md
```diff
@@ -495,6 +495,9 @@ The ACE must provide a Metadata service on the address given to the applications
 
 Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
 
+**NOTE**: Implementations can insert an authorization token into AC_METADATA_URL.
```
jonboulle: `AC_METADATA_URL`

SPEC.md
```diff
@@ -495,6 +495,9 @@ The ACE must provide a Metadata service on the address given to the applications
...
+For example, AC_METADATA_URL can be set to `https://10.0.0.1:8888/Y4vFeVZzKM2T9rwkpWHfqXuGsNjS6O5c` with the path portion acting as a token to uniquely and securely identify a pod.
```
jonboulle: `AC_METADATA_URL`

jonboulle: Don't you want to talk about the entropy as previously?

eyakubovich: I guess. To be honest, I still feel uncomfortable with specifying level of entropy. The level of security is something that the implementation should decide on its own. It's immaterial for interop and prescribing something that's not actually required always makes me uneasy.

jonboulle: That's what SHOULD is for, right?

On Thu, Apr 30, 2015 at 3:31 PM, Eugene Yakubovich <notifications@github.com

schancel:
@jonboulle Yeah this is what I was originally proposing on IRC.   Either way seems fine to me.

SPEC.md
jonboulle: @eyakubovich the point I was trying to make here is
```
`AC_METADATA_URL`
```
for consistency with above/below

jonboulle:
one more nit! then let's ship this!

jonboulle:
I will address this in a follow-up.

Issue #292, opened 2015-04-09
spec has bad gpg encryption command


hawicz:
In SPEC.md, the gpg command to encrypt the aci uses the same name for the output and input, which can't work because as soon as it opens and truncates the output file it looses the input data.
Instead of:
> gpg --output reduce-worker.aci --digest-algo sha256 --cipher-algo AES256 --symmetric reduce-worker.aci

It should be something like:
> gpg --output reduce-worker.aci.out --digest-algo sha256 --cipher-algo AES256 --passphrase foo --symmetric reduce-worker.aci
> mv reduce-worker.aci.out reduce-worker.aci

philips:
@hawicz Good call, would you like to send a patch to fix this?

jonboulle:
fixed by #328

PR #328, opened 2015-04-29
spec: fix gpg example and clarify ACI signing

jonboulle:
As reported in #292, the example gpg command was truncating output by re-using
the input file. Instead, we write to a separate temporary output file and then
overwrite (mv) the original.

This also clarifies the ordering of operations for images: signing should
happen *after* (optional) encryption, not just compression. This was
previously discussed in #19, #156, and #204.


jonboulle:
fixes #292

philips:
lgtm

PR #333, opened 2015-05-01
SPEC.md: make the fields consistent

vbatts:
/cc @jonboulle

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -543,6 +543,8 @@ Accessible at `$AC_METADATA_URL/acMetadata/v1/pod/hmac`
 
 ### AC Name Type
 
+Abbreviated reference as `acName`.
+
 An AC Name Type is restricted to lowercase characters accepted by the DNS [RFC1123](http://tools.ietf.org/html/rfc1123#page-13) and "/".
 An AC Name Type cannot be an empty string and must begin and end with an alphanumeric character.
 An AC Name Type will match the following [RE2](https://code.google.com/p/re2/wiki/Syntax) regular expression: `^[a-z0-9]+([-./][a-z0-9]+)*$`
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -543,6 +543,8 @@ Accessible at `$AC_METADATA_URL/acMetadata/v1/pod/hmac`
 
 ### AC Name Type
 
+Field `acName`
```
jonboulle: This isn't really correct, there's no such field acName...

vbatts: ohman. It's app/name. `acName` completed on my editor. hmm

jonboulle: Hrmmm but still, the AC Name type is used much more broadly than for the app name..

vbatts: what are you saying? just remove acName here? since it's used else where as a formatting strictness for field values

jonboulle: I guess I am unclear what you are trying to point out with these "field" references - in the case of acKind and acVersion it is a simple 1:1 between the type and the field name, but in the case of AC Name it's a more generic type used by a whole bunch of fields...

jonboulle: So maybe just removing it here makes sense.

vbatts: k

vbatts:
updated. PTAL

SPEC.md
```diff
@@ -543,6 +543,8 @@ Accessible at `$AC_METADATA_URL/acMetadata/v1/pod/hmac`
 
 ### AC Name Type
 
+Abbreviated reference as `acName`.
```
jonboulle: sorry, this doesn't make sense; `acName` isn't mentioned anywhere else in the spec, and the other kinds are not references but field names. IMHO just remove this and 565 / 578

vbatts: K

jonboulle: ping

jonboulle:
deprecated by #399

PR #334, opened 2015-05-03
SPEC.md: formatting of bullets and blocks

vbatts:
Some markdown parsers need a space before a list of bullets.

And the code type declarations on blocks can be pretty.


jonboulle:
Lgtm, thanks

PR #366, opened 2015-05-04
README: Add Kurma to list of implementations

krobertson:
Added Kurma to the list of implementations in the README. Looked like the list
was in alphabetical order.

@jonboulle @vbatts


jonboulle:
Lgtm

PR #367, opened 2015-05-04
Update SPEC.md

AXGKl:
Small typos

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -131,7 +131,7 @@ The shared context is defined as the conjunction of the following Linux namespac
 - IPC namespace
 - UTS namespace
 
-The context may also consist of one or more isolators.
+The context MAY also consist of one or more isolators.
 
 The definition of the **pod** - namely, the list of constituent apps, and any isolators that should apply to the entire pod - is codified in a [Pod Manifest](#pod-manifest-schema).
 Pod Manifests can serve the role of both _deployable template_ and _runtime manifest_: a template can be a candidate for a series of transformations before execution.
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -215,6 +215,7 @@ The following environment variables MUST be set for each application's main proc
 * **AC_METADATA_URL** URL where the [metadata service](#app-container-metadata-service) for this pod can be found.
 
 An executor MAY set additional environment variables for the application processes.
+`$PATH` MAY contain more directories than the ones listed.
```
jonboulle: This is actually a behavioural change; what's your rationale?

jonboulle:
Thanks for the patch and apologies for getting to this out of order - I think you will need to rebase against master as some of these have been resolved.

SPEC.md
AXGKl: Hi Jonathan, 

sorry for delayed answer - second baby arrived last week ;-)

Regarding $PATH: Maybe I misunderstood sth but I thought it makes a lot of sense to be able to optionally allow additional directories already in the execution environment.
Why?
For example when an image contains different versions of an executable in different directory trees (say, /.../stable/myapp vs. /.../devel/myapp, from, say repo checkouts) - then you could in a PaaS like surrounding framework allow the user to simply switch the version of what is executed, globally and simple. Sure you could invent indirections via other environment variables, checked by wrapper scripts you could add inside - but I find $PATH to be the most direct and most transparent way for user to control what is started in such a setup.

Maybe I misunderstand the philosophy of appc and you want to get as close to freshly "booted" VMs as possible but since you allow explicitly "additional environment variables" in the principal exec env I would say the $PATH should also be allowed to be widened.

philips:
I am going to rebase and merge this myself now.

philips:
Merged most of the typos at 9b7889b. I am fine discussing the [PATH thing](https://github.com/appc/spec/pull/367#discussion_r30067853) but it should be an explicit PR about it.

philips:
@AXGKl Would you mind starting a new PR about this PATH thing?

PR #362, opened 2015-05-05
Metadata: specious argument

jonboulle:
"Writing files leads to assumptions like a libc environment attempting parse /etc/hosts"

I don't buy that this is a problem.  If you want to change something so fundamental to the UNIX way of operating, you can't just plonk it on people's heads and demand they change.  I find this to be a weak argument.  The rest seem OK, though.


jonboulle:
@thockin I agree, this is a wart; I imagine this section/justification will change more significantly soon, but posted a simple change for now, ptal.

PR #353, opened 2015-05-05
Clarify pathWhitelist

jonboulle:
We should clarify the difference between an empty list (which implies NOTHING is allowed - is that right?) vs a non-existing key/value which you spec as "all files".


jonboulle:
@thockin actually we had intended that they were equivalent - posted patch to make this clearer. (I can't see any reason to support a no-files case, since there's then no point in having any dependencies and the archive should just be empty)

thockin:
OK.  It's a special case but at least it is documented.

LGTM

PR #350, opened 2015-05-05
Image manifest unclear sentence

jonboulle:
"used as a human readable index to the App Container Image" doesn't mean anything to me.  Needs a better description.


jonboulle:
@thockin is this sufficient?

SPEC.md
```diff
@@ -732,7 +732,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
 
 * **acKind** (string, required) must be an [AC Kind](#ac-kind-type) of value "ImageManifest"
 * **acVersion** (string, required) represents the version of the schema specification [AC Version Type](#ac-version-type)
-* **name** (string, required) used as a human readable index to the App Container Image. (string, restricted to the [AC Name](#ac-name-type) formatting)
+* **name** (string, required) a human-readable name for this App Container Image (string, restricted to the [AC Name](#ac-name-type) formatting). This is not expected to be unique (see the **version** label) but SHOULD have a URL-like structure to facilitate **[App Container Image Discovery](#app-container-image-discovery)**.
```
thockin: Is it assumed that this name, when resolved through the discovery process, will result in this same image?  Might be worth saying something like that.

jonboulle: We do describe that [here](https://github.com/appc/spec/blob/master/SPEC.md#validation):
> Implementations must ensure that the name in the Image Manifest in the retrieved ACI matches the initial name used for discovery.

This is one of those awkward moments where we have erred on the side of brevity and not duplicated information between the manifest schema and the body of the spec, but referencing it again here would arguably be useful too... any strong feelings?

thockin: I tend to err on the side of being too explicit.  In this case I think it
is an important point.

jonboulle:
PTAL

thockin:
LGTM

PR #347, opened 2015-05-05
Pod schema on isolators is unclear

jonboulle:
It says isolators "apply to all apps".  Does it mean they apply to each app or to the set of apps in total?  I assume in total, but it's not clear.


jonboulle:
@thockin tweaked wording, IMHO this is an adequate explanation in this context because the authoritative definition is here, lmk if you think otherwise - https://github.com/appc/spec/blob/b420d1ec0839c728c8f82bbb9410a2da09956e5a/SPEC.md#isolators

thockin:
LGTM.  How did you convert an issue into a PR?

jonboulle:
@thockin deprecated (but still functional, so I'm using it while it lasts) feature of [hub](https://github.com/github/hub/blob/master/commands/pull_request.go#L30)

PR #363, opened 2015-05-05
Metadata clarification - tokens

jonboulle:
Maybe worth clarifying that auth tokens must not be easily discoverable (e.g. don't just use the pod UUID).  Containers will eventually attack hosts or other containers.  This should be tight enough to at least say that containers should have a VERY hard time guessing each others tokens.


jonboulle:
@thockin ptal
/cc @vbatts @schancel @eyakubovich

thockin:
LGTM

PR #355, opened 2015-05-05
Dependency matching: claims 3 criteria, lists 2

jonboulle:
"if it meets one of three criteria" - then you define only two criteria.


jonboulle:
@thockin attached

thockin:
LGTM

PR #346, opened 2015-05-05
Spec unclear on isolators

jonboulle:
Pod schema says "value can be a freeform string"

Image schema says just list of objects and nothing more

Isolators section says value "can be an arbitrary JSON value"


jonboulle:
Attached.

thockin:
LGTM

Issue #357, opened 2015-05-04
ACE: network clarifications


thockin:
"should be configured with an IPv4/IPv6 address" -- How many interfaces are required?  v4 AND v6, v4 OR v6, or v4 XOR v6?

Is loopback part of the spec?  Who sets up the loopback device? What guarantees does it have for v4 vs v6?

jonboulle:
Hrm, we sort of implicitly touch on this in the resource/network-bandwidth isolator:

>   **NOTE**: Network limits MUST NOT apply to localhost communication between apps in a pod.

I don't see a good reason to not be more explicit in the networking section about providing a loopback interface - can you take a pass at wording this?

Re: IPv4/IPv6, OR/AND seems reasonable to me.

/cc @eyakubovich

thockin:
I would if I knew the intentions, though it might be a bit - I have a lot to get to in the coming weeks.

jonboulle:
fixed by #368

PR #368, opened 2015-05-05
Clarify net interfaces

thockin:
Open question: do you really require routable IPs at this layer of the stack?  This implies that there is no way to do a machine-private container...

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -190,8 +190,8 @@ In this case, the executor will bind mount the host's `/opt/tenant1/work` direct
 
 #### Network Setup
 
-A Pod must have a [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interface, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
-The network interface MUST be configured with an IPv4/IPv6 address that is reachable from other pods.
+A Pod must have a loopback network interface and zero or more [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interfaces, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
+Each network interface MUST be configured with one or more IPv4 and/or IPv6 addresses.
 
 #### Logging
 


thockin:
Fixes #357

jonboulle:
I don't feel really strongly about this. This is a really fundamental part of the networking model we are trying to encourage; there are obvious use cases for machine-private containers but IMHO it's not the end of the world if they're not considered to implement the entire spec in that mode. I guess we could just change MUST to SHOULD.

SPEC.md
```diff
@@ -190,8 +190,8 @@ In this case, the executor will bind mount the host's `/opt/tenant1/work` direct
 
 #### Network Setup
 
-A Pod must have a [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interface, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
-The network interface MUST be configured with an IPv4/IPv6 address that is reachable from other pods.
+A Pod must have a loopback network interface and one or more [layer 3](http://en.wikipedia.org/wiki/Network_layer) (commonly called the IP layer) network interfaces, which can be instantiated in any number of ways (e.g. veth, macvlan, ipvlan, device pass-through).
```
eyakubovich: I would actually require zero or more non-loopback interfaces. It provides a possibility to have a pod be "off the network". This is problematic today as it would make it impossible to reach metadata service but since we want to make it optional, it should not be an issue.

SPEC.md
```diff
@@ -190,8 +190,8 @@ In this case, the executor will bind mount the host's `/opt/tenant1/work` direct
...
+Each network interface MUST be configured with one or more IPv4 and/or IPv6 addresses.  Non-loopback interfaces must have addresses that are reachable from other pods.
```
eyakubovich: The requirement to have non-loopback interfaces have IP's that are reachable from other pods is both too vague and too strong. It's vague since it does not specify which other pods. appc spec does not talk about clusters so it's unclear about the locus of pods here. It's too strong since we may want to have a pod with two non-loopback interfaces. One of them to talk to other pods and another to talk to the Internet. The second interface would violate the proposed language.

SPEC.md
thockin: good point

SPEC.md
thockin: proposed fix?  Maybe just drop the last sentence?

thockin:
New push

eyakubovich:
LGTM

PR #343, opened 2015-05-05
SPEC unclear: sha512 only supported form or just one of?

jonboulle:
'must be of the format "type-value", where "type" is "sha512"'

Does this mean that sha512 is the only supported type?  Earlier in the spec it read as if other common formats were valid, but not default.


jonboulle:
howzat?

thockin:
LGTM

PR #370, opened 2015-05-05
Delete the "Metadata-Flavor:" header requirement

mpasternacki:
The requirement has no clear purpose, and makes it harder to use some
tools. In particular, FreeBSD base system includes `/usr/bin/fetch`
that could be used to query the metadata service, but which doesn't
support setting custom headers. Removing this requirement makes it
possible to use FreeBSD base system without any additional software
installed.

An alternative approach, if there is any reason for keeping this requirement, would be to clearly specify the repercussions (and lower the requirement to SHOULD), e.g.:

> Clients querying any of these endpoints SHOULD specify the `Metadata-Flavor: AppContainer` header. ACE implementations MAY reject metadata service requests that do not contain such header, responding with a "400 Bad Request" HTTP status.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -497,8 +497,6 @@ The App Container specification defines an HTTP-based metadata service for provi
 
 The ACE must provide a Metadata service on the address given to the applications via the `AC_METADATA_URL` [environment variable](#execution-environment).
 
-Clients querying any of these endpoints MUST specify the `Metadata-Flavor: AppContainer` header.
-
 ACE implementations SHOULD embed an authorization token in `AC_METADATA_URL`, which provides a means for the metadata service to uniquely and securely identify a pod.
 For example, `AC_METADATA_URL` passed to a pod could be set to `https://10.0.0.1:8888/Y4vFeVZzKM2T9rwkpWHfqXuGsNjS6O5c` with the path portion acting as a token.
 Since the token is used by the Metadata Service to authenticate the pod's identity, it SHOULD have no fewer than 128 bits of entropy (i.e. size of UUID), and SHOULD NOT be easily guessable (e.g. the pod UUID should not be used).


schancel:
I was going to raise the same issue.  I don't see the purpose of this header being required.

jonboulle:
I believe the inspiration here was Google's metadata service: https://cloud.google.com/compute/docs/metadata#querying 

> This header indicates that the request was sent with the intention of retrieving metadata values, rather than unintentionally from an insecure source, and allows the metadata server to return the data you requested. If you do not provide this header, the metadata server denies your request.

In the generic appc case, the utility of this is highly questionable, ssince a) clients making requests need to specify the full path to metadata endpoints, which clearly signals their intent; and b) we now recommend the authorization token, which is a much more robust means of identifying "secure sources".

So, eyakubovich and I are fine with scrapping this - but would be good to have @philips weigh in as he added this originally.

mpasternacki:
I understand the threat that this could work around: IF a container is running an application that may accept untrusted input, AND that application can be tricked into performing an HTTP request (that either triggers an administrative action, or its results can be returned back to the attacker), AND the URL of the metadata service is known, then attacker has indirect access to the metadata service. In appc case, this means that attacker may get access to some sensitive information present in manifest or annotations. This may have greater impact if user creatively utilizes signing by the metadata service without isolating it from end user: tricking mds into signing a token chosen by attacker, and returning the signature to the attacker, which may be used to escalate privileges. Requiring a custom header means that the metadata service will only serve clients that explicitly show intent to reach the metadata service.

A hard-to-guess, unique, verified URL (the URL token) seems to be more robust protection against this kind of scenario. It adds a (much harder) requirement that the attacker needs to see pod's environment variables, and image can protect itself by erasing `AC_METADATA_URL` from environment before executing the world-facing process. Still, requiring a custom header provides an additional layer of security against unintended access.

I don't feel qualified to make that call now. We can say that URL token + possibility of cleaning the environment (which would be responsibility of image author: not a good idea most of the time) is secure enough and we want to keep the spec simple. On the other hand, we can go for extra security at a price of extra complexity by either keeping the requirement as it is (possibly adding a note explaining why it is important), or making it optional (e.g. my alternative SHOULD/MAY wording).

philips:
We currently recommend the use of an authorization token so I suggest we remove the Flavor header: https://github.com/appc/spec/blob/master/SPEC.md#metadata-service

philips:
LGTM can I get another maintainer to ack? @cdaylward @jonboulle @thockin @vbatts

jonboulle:
> We currently recommend the use of an authorization token so I suggest we remove the Flavor header: https://github.com/appc/spec/blob/master/SPEC.md#metadata-service

huh? that's exactly what we're doing here!

I acked earlier, so merging this.

PR #371, opened 2015-05-05
spec: tweak pod definition

### LINKED ISSUE #318: spec: pod and mount namespace
The spec states that the shared context for a pod includes the mount namespace. I am not sure what exactly it means given that each application runs in its own chroot and mount points can only shared using explicit bind statements. Shouldn't the spec drop the mount namespace from the shared context?

Also, it is not clear what is the exact behavior of a pod volume with `kind=empty` that is mounted for several applications. The specs states that *writes go to the app's chroot* which I read as implying that each app gets its own volume and these are not shared. Or is the intention that such volume is shared but its exact location on the host is implementation-dependent?

jonboulle:
- remove mount namespace requirement and instead switch to "shared
 volumes" language. As raised previously elsewhere [1][1] [2][2], this
 is somewhat moot since applications start chrooted into their
 respective filesystems anyway.
- Add illustrative explanation of the significance of the namespaces

[1]:
https://github.com/GoogleCloudPlatform/kubernetes/pull/4701#discussion_r25125656
[2]: https://github.com/appc/spec/issues/318

Fixes #318

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -122,16 +122,16 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
 ## App Container Pods (pods)
 
 The deployable, executable unit in the App Container specification is the **pod**.
-A **pod** is a list of apps that will be launched together inside a shared context.
-The shared context is defined as the conjunction of the following Linux namespaces (or equivalents on other operating systems):
+A **pod** is a list of apps that will be launched together inside a shared execution context.
+The execution context can be defined as the conjunction of several Linux namespaces (or equivalents on other operating systems):
 
-- PID namespace
-- network namespace
-- mount namespace
-- IPC namespace
-- UTS namespace
+- PID namespace (apps within the pod can see each other's processes)
+- network namespace (apps within the pod have access to the same IP and port space)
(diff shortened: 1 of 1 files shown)


jonboulle:
I spent a bit of time agnosing over wording here before deciding that I couldn't really improve on what I went with in https://github.com/GoogleCloudPlatform/kubernetes/pull/4701, so this is cribbed liberally from there.

/cc @thockin

SPEC.md
```diff
@@ -122,16 +122,16 @@ An optional *path whitelist* can be provided, in which case all non-specified fi
...
-The shared context is defined as the conjunction of the following Linux namespaces (or equivalents on other operating systems):
+A **pod** is a list of apps that will be launched together inside a shared execution context.
+The execution context can be defined as the conjunction of several Linux namespaces (or equivalents on other operating systems):
 
-- PID namespace
-- network namespace
-- mount namespace
-- IPC namespace
-- UTS namespace
+- PID namespace (apps within the pod can see each other's processes)
```
thockin: I'd say "see and signal"

jonboulle: Done.

jonboulle:
oops, I thought this landed already. @thockin ?

PR #358, opened 2015-05-05
Isolators suffixes

jonboulle:
'the following represent roughly the same value: 128974848, "129e6", "129M" , "123Mi"'

To my eyes these are NOT the same at all, and comparing them makes me nervous.  129e6 and 129M are exactly 129000000.  123Mi is exactly 128974848.  This is VERY IMPORTANT.  Please clarify the intentions here.

For example, kubernetes is very strict that the resource encoding is exact and not lossy at all, except in one way (that we only handle 3 decimal places).  We spec that numbers are *never* actually stored as floating point values (it's lossy).  It's easy to screw this up in subtle ways that make the whole thing become trouble.  E.g. you can easily find a set of numbers which, after processing, add up to a feasible set in one order and not in another order.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -282,10 +282,10 @@ If **request** is omitted, it defaults to the value of **limit**.
 - **limit** is the maximum amount of a resource available to the app/pod.
 If the app/pod consumes a resource in excess of its limit, it must be terminated or throttled to no more than the limit.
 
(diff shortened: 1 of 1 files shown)


jonboulle:
@thockin PTAL
Also fixes #360

thockin:
LGTM

PR #375, opened 2015-05-06
README.md: correctly show shell prompt.

alextoombs:
Very minor change, just noticed on a readthrough.


jonboulle:
The usual argument against this is that it makes things more easily copy pastaable. But we have more examples the other way, so this lgtm for consistency. Thanks.

PR #273, opened 2015-05-07
spec: reconsider SHELL/HOME environment variables

jonboulle:
As of c8559a283567c1c6840258c6833f5f832fc66cb6 the environment variables mandated by the spec are [as follows]([https://github.com/appc/spec/blob/c8559a283567c1c6840258c6833f5f832fc66cb6/SPEC.md#execution-environment]):
1. `PATH`
2. `USER`
3. `LOGNAME` (same as `USER`)
4. `HOME`
5. `SHELL`
6. `AC_APP_NAME`
7. `AC_METADATA_URL` 

1, 2 and 3 seem relatively uncontroversial - it's reasonable to assume every app executing on a modern OS has these concepts (inherently associated with the process). 6 and 7 are similarly straightforward expectations defined clearly in the spec itself. 

4 and 5 are slightly more problematic because, although part of the POSIX `passwd`, their significance and relevance for applications executing in containers is murky. They are conceptually really associated with an interactive user (--> login session), and it seems to me it should be possible to completely implement the execution side of the appc spec without introducing such a notion. Also, these are traditionally looked up from a database like `/etc/passwd` and it is unclear how this should be discussed (if at all) in the spec. Related: #231

Throwing out a few ideas for clarifying this:
- Declare only that `HOME` and `SHELL` should have "sensible" values but do not have any particular significance or requirements in the spec itself
- Remove `SHELL` entirely from the list of required environment variables ([now that we have relaxed the restrictions on what can be set](https://github.com/appc/spec/pull/228))
- Make `SHELL` optional and mandate (or recommend?) that it be used iff interactive sessions are to be supported
- Stipulate that `HOME` must reference a directory in the rootfs that is writeable by the user (i.e. the `app.User` field)

### DIFF
--- a/OS-SPEC.md
+++ b/OS-SPEC.md
@@ -9,10 +9,12 @@ ACIs that define the label `os=linux` can expect this environment by default.
 
 ## Devices and File Systems
 
+The following devices and filesystems MUST be made available in each application's filesystem
+
 |     Path     |  Type  |  Notes  |
 | ------------ | ------ | ------- |
-| /proc        | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)   | |
-| /sys         | [sysfs](https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt)     | |
+| /proc        | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)    | |
+| /sys         | [sysfs](https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt)    | |
 | /dev/null    | [device](http://man7.org/linux/man-pages/man4/null.4.html)                 | |
 | /dev/zero    | [device](http://man7.org/linux/man-pages/man4/zero.4.html)                 | |
 | /dev/full    | [device](http://man7.org/linux/man-pages/man4/full.4.html)                 | |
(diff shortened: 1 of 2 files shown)


jonboulle:
@thockin @eyakubovich @vcaputo took a swing at this, ptal

OS-SPEC.md
```diff
@@ -7,12 +7,23 @@ This document attempts to define a minimal set of things that must exist for mos
 
 ACIs that define the label `os=linux` can expect this environment by default.
 
+## Execution Environment
+
+It is RECOMMENDED that the following environment variables be set for each application's main process and any lifecycle processes:
```
thockin: Is there some real requirement driving the desire for these?  I feel like this is soft enough that it doesn't matter at all and should just be stricken entirely.

jonboulle: I was anticipating that there is still a nontrivial amount of software that does depend on these things and so it would be worth gently nudging implementers in that direction. But I don't feel strongly enough about it to keep fighting this :-)

thockin: Nothing says an implementation CAN'T do it, I just don't know that this
adds much to the spec.

jonboulle:
updated

thockin:
beautiful

jonboulle:
@thockin you need to use more gifs to express yourself IMO

thockin:
http://giphy.com/gifs/reactiongifs-qKNfbNbEN5ylW

jonboulle:
very good

PR #397, opened 2015-05-07
spec: Rename app name to image name.

yifan-gu:
Use `app name` in this section seems a little bit out of tune to me. How about changing it to `image name`?

/cc @jonboulle @philips

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -383,10 +383,12 @@ Sub-units (e.g. decimals, "0.3", or milli-units, "300m") are NOT permissible.
 
 ## App Container Image Discovery
 
-An app name has a URL-like structure, for example `example.com/reduce-worker`.
-However, there is no scheme on this app name, so it cannot be directly resolved to an App Container Image URL.
-Furthermore, attributes other than the name may be required to unambiguously identify an app (version, OS and architecture).
-App Container Image Discovery prescribes a discovery process to retrieve an image based on the app name and these attributes.
+An App Container Image name has a URL-like structure, for example `example.com/reduce-worker`.
+However, there is no scheme on this name, so it cannot be directly resolved to an App Container Image URL.
+Furthermore, attributes other than the name may be required to unambiguously identify an image (version, OS and architecture).
+These attributes are expressed in the **labels** field of the [Image Manifest](#image-manifest-schema).
+App Container Image Discovery prescribes a discovery process to retrieve an image based on the name and these attributes.
+
 Image Discovery is inspired by Go's [remote import paths](https://golang.org/cmd/go/#hdr-Remote_import_paths).
 
 There are three URLs types:
@@ -407,7 +409,7 @@ The simple discovery template is:
 
 First, try to fetch the App Container Image by rendering the above template (with `{ext}` rendered to `aci`) and directly retrieving the resulting URL.
 
-For example, given the app `{name}`: `example.com/reduce-worker`, with `{version}`: `1.0.0`, `{arch}`: `amd64`, and `{os}`: `linux`, try to retrieve:
+For example, given the image `{name}`: `example.com/reduce-worker`, with `{version}`: `1.0.0`, `{arch}`: `amd64`, and `{os}`: `linux`, try to retrieve:
 
     https://example.com/reduce-worker-1.0.0-linux-amd64.aci
 
@@ -420,7 +422,7 @@ Simple discovery does not provide a way to discover Public Keys.
 
 ### Meta Discovery
 
-If simple discovery fails, then we use HTTPS+HTML `meta` tags retrieved from a "discovery URL" to resolve an app name to downloadable URLs.
+If simple discovery fails, then we use HTTPS+HTML `meta` tags retrieved from a "discovery URL" to resolve an image name to downloadable URLs.
 
 The template for the discovery URL is:
 
@@ -452,10 +454,10 @@ When evaluating `ac-discovery` tags, the client MUST first ensure that the prefi
 
 ```bash
 urltmpl="https://{name}-{version}-{os}-{arch}.{ext}"
-curl $(echo "$urltmpl" | sed -e "s/{name}/$appname/" -e "s/{version}/$version/" -e "s/{os}/$os/" -e "s/{arch}/$arch/" -e "s/{ext}/$ext/")
+curl $(echo "$urltmpl" | sed -e "s/{name}/$name/" -e "s/{version}/$version/" -e "s/{os}/$os/" -e "s/{arch}/$arch/" -e "s/{ext}/$ext/")
 ```
 
-where _appname_, _version_, _os_, and _arch_ are set to their respective values for the application, and _ext_ is either `aci` or `aci.asc` for retrieving an App Container Image or signature respectively.
+where _name_, _version_, _os_, and _arch_ are set to their respective values for the image, and _ext_ is either `aci` or `aci.asc` for retrieving an App Container Image or signature respectively.
 
 Note that multiple `ac-discovery` tags MAY be returned for a given prefix-match (for example, with different scheme names representing different transport mechanisms).
 In this case, the client implementation MAY choose which to use at its own discretion.
@@ -703,7 +705,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
     },
     "dependencies": [
         {
-            "app": "example.com/reduce-worker-base",
+            "imageName": "example.com/reduce-worker-base",
             "imageID": "sha512-...",
             "labels": [
                 {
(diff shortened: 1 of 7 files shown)


jonboulle:
I think by and large this makes sense. One critical omission here: the "apps" field in the "dependencies" list.

yifan-gu:
Updated.

SPEC.md
```diff
@@ -447,10 +447,10 @@ When evaluating `ac-discovery` tags, the client MUST first ensure that the prefi
 
 ```bash
 urltmpl="https://{name}-{version}-{os}-{arch}.{ext}"
-curl $(echo "$urltmpl" | sed -e "s/{name}/$appname/" -e "s/{version}/$version/" -e "s/{os}/$os/" -e "s/{arch}/$arch/" -e "s/{ext}/$ext/")
+curl $(echo "$urltmpl" | sed -e "s/{name}/$imgname/" -e "s/{version}/$version/" -e "s/{os}/$os/" -e "s/{arch}/$arch/" -e "s/{ext}/$ext/")
```
jonboulle: This should just be name, I think, not sure why it was appname in the first place?

SPEC.md
```diff
@@ -378,10 +378,10 @@ Sub-units (e.g. decimals, "0.3", or milli-units, "300m") are NOT permissible.
 
 ## App Container Image Discovery
 
-An app name has a URL-like structure, for example `example.com/reduce-worker`.
-However, there is no scheme on this app name, so it cannot be directly resolved to an App Container Image URL.
-Furthermore, attributes other than the name may be required to unambiguously identify an app (version, OS and architecture).
-App Container Image Discovery prescribes a discovery process to retrieve an image based on the app name and these attributes.
+An app container image name has a URL-like structure, for example `example.com/reduce-worker`.
```
jonboulle: s/app container image/App Container Image/

SPEC.md
```diff
@@ -378,10 +378,10 @@ Sub-units (e.g. decimals, "0.3", or milli-units, "300m") are NOT permissible.
...
+However, there is no scheme on this image name, so it cannot be directly resolved to an App Container Image URL.
```
jonboulle: s/image name/name/

SPEC.md
```diff
@@ -378,10 +378,10 @@ Sub-units (e.g. decimals, "0.3", or milli-units, "300m") are NOT permissible.
...
+Furthermore, attributes other than the name may be required to unambiguously identify an image (version, OS and architecture).
+App Container Image Discovery prescribes a discovery process to retrieve an image based on the image name and these attributes.
```
jonboulle: Can you please add a line: "These attributes are expressed in the *labels* field of the ImageManifest"

SPEC.md
```diff
@@ -762,8 +762,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
 * **dependencies** (list of objects, optional) dependent application images that need to be placed down into the rootfs before the files from this image (if any). The ordering is significant. See [Dependency Matching](#dependency-matching) for how dependencies are retrieved.
-    * **app** (string, required) name of the dependent App Container Image.
-    * **imageID** (string of type [Image ID](#image-id-type), optional) content hash of the dependency. If provided, the retrieved dependency must match the hash. This can be used to produce deterministic, repeatable builds of an App Image that has dependencies.
+    * **name** (string of type [AC Name](#ac-name-type), required) name of the dependent App Container Image.
+    * **id** (string of type [Image ID](#image-id-type), optional) content hash of the dependency. If provided, the retrieved dependency must match the hash. This can be used to produce deterministic, repeatable builds of an App Image that has dependencies.
```
jonboulle: s/App Image/App Container Image/

jonboulle:
@yifan-gu sorry, while you're in there..

yifan-gu:
@jonboulle NW. Updated again.

SPEC.md
```diff
@@ -378,10 +378,13 @@ Sub-units (e.g. decimals, "0.3", or milli-units, "300m") are NOT permissible.
...
-However, there is no scheme on this app name, so it cannot be directly resolved to an App Container Image URL.
-Furthermore, attributes other than the name may be required to unambiguously identify an app (version, OS and architecture).
-App Container Image Discovery prescribes a discovery process to retrieve an image based on the app name and these attributes.
+An App Container Image name has a URL-like structure, for example `example.com/reduce-worker`.
+However, there is no scheme on this name, so it cannot be directly resolved to an App Container Image URL.
+Furthermore, attributes other than the name may be required to unambiguously identify an image (version, OS and architecture).
+These attributes are expressed in the **labels** field of the [Image Manifest](#image-manifest-schema).
+App Container Image Discovery prescribes a discovery process to retrieve an image based on the name and these attributes.
+
+
```
jonboulle: superfluous newline

jonboulle:
@yifan-gu Spec change lgtm. Please update the schema alongside this, sorry I forgot to mention that earlier...

yifan-gu:
@jonboulle Then I found we need to change a bunch of places as well.... I have changed all the appearance of `ImageID`.  PTAL.

krobertson:
LGTM, make more sense than "app name".

SPEC.md
```diff
@@ -762,8 +764,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
 * **dependencies** (list of objects, optional) dependent application images that need to be placed down into the rootfs before the files from this image (if any). The ordering is significant. See [Dependency Matching](#dependency-matching) for how dependencies are retrieved.
-    * **app** (string, required) name of the dependent App Container Image.
-    * **imageID** (string of type [Image ID](#image-id-type), optional) content hash of the dependency. If provided, the retrieved dependency must match the hash. This can be used to produce deterministic, repeatable builds of an App Image that has dependencies.
+    * **name** (string of type [AC Name](#ac-name-type), required) name of the dependent App Container Image.
```
cdaylward: Would prefer to see this as "image". "name" seems to always be confusing because it is often ambiguous as to whether it is a reference to something else, or the name of the object itself (in this case, the name of the dependency).

jonboulle: @cdaylward my initial argument in favour of `name` would be that it _is_ a key reference. But on thinking about it further it seems like:
a) this field should be optional; one might desire a deterministic set of dependencies (i.e. only referencing by ID) and the name field is not then strictly necessary (provided of course the implementation supports lookups based on ID alone)
b) a qualified `imageName` / `imageID` might be preferable since these are not actually image objects (rather, dependency objects)

@yifan-gu thoughts?

yifan-gu: @jonboulle 
For a), if the user also specify the labels, then it can identify the image deterministically, not? My previous understanding on this is that sha512 is more for machine than for human.

For b), I agree that they are not image objects by definition, but they are still describing images. If that's pretty clear to user, or if images are the dependency object, then I think removing the 'image' part is OK. Furthermore, maybe we can call it "required images" instead of dependencies?

jonboulle: @yifan-gu There is nothing to guarantee that labels are deterministic, no. They _should_ be, but we can't mandate it without having a centralised repository.

philips:
lgtm, overall. I will let @jonboulle resolve the image/name discussion with @cdaylward and @yifan-gu

yifan-gu:
@jonboulle I am ok with imageName, imageID, (do we need to change to imageLabels as well)?

SPEC.md
```diff
@@ -700,8 +702,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
     },
     "dependencies": [
         {
-            "app": "example.com/reduce-worker-base",
-            "imageID": "sha512-...",
+            "name": "example.com/reduce-worker-base",
+            "id": "sha512-...",
```
vbatts: perhaps this ID need a more generic name. It is defined as [Image ID](https://github.com/appc/spec/blob/master/SPEC.md#image-id), but could likely be CAS ID or similar

jonboulle: @vbatts I don't really follow, how does this relate to the discussion below?

SPEC.md
vbatts: lets see, the content addressed iD presently implies a tar archive, which is an ACI archive. 
"Dependencies" could be clarified a bit as "Required Images". This is a common pattern picked up in other packaging formats like RPM ("Requires").

Is there any other type of dependency _besides_ images?

SPEC.md
vbatts: dropping this to ID here seems too terse. UUID is an ID as well, but that is not what is meant here. Sorry if it's off-topic, it just seems a little too terse. Like the next step would be to loosen up the requirement to be Image ID formatted.

SPEC.md
jonboulle: @vbatts 
> Is there any other type of dependency besides images?
Nope.
I'm sympathetic to "required images".

SPEC.md
jonboulle: Let's revert this to "imageID". Right now it does mean something rather strict. If we want to loosen the requirement we can address that separately.

yifan-gu: OK

jonboulle: sorry, I already started down this path but forgot to put up a PR - https://github.com/jonboulle/spec/commit/e343fa9f41ab0822001fe02bfde469872e8de6de

yifan-gu: @jonboulle That looks good. You can send as a PR so I can close this :)

yifan-gu: NV, Let me just pull from you commit

yifan-gu:
Keeps the imageID, and change app to imageName.

jonboulle:
@yifan-gu I don't think you just pulled my commit - the build is broken now :/

yifan-gu:
@jonboulle I did, and squashed the commits. For some reason we forgot this line :(
https://github.com/jonboulle/spec/commit/e343fa9f41ab0822001fe02bfde469872e8de6de#diff-dfef21140f3e6e95b72421f8887971d0R866

jonboulle:
LGTM. Thanks.

Issue #342, opened 2015-05-04
AC Name formatting is potentially problematic


thockin:
I see the desire to have one single name format, but I find it worrisome.

The reason kubernetes uses DNS formatting rather than "DNS plus" as this does is so that these names could conceivably be used in DNS.  The AC Name format precludes that assumption.  Consider port names - if they are literally DNS compatible (specifically DNS labels and not subdomains), we can generate SRV records for named ports.  If they can contain `/` and `.` that is not practical.

Using AC Name for things that have global namespaces makes sense (label keys, image names).  Using something more restrictive for things with smaller scope is better, I think.  Apps in a pod, mounts, volumes, ports, etc.  They are all relative to their pod anyway.

Discuss.

jonboulle:
Having a more restrictive type for some of these fields makes sense to me; arguably we got lazy using AC Name for many more fields than its original purpose...

thockin:
I'm a fan of DNS RFC 1132 (which adds onto RFC 1035 and RFC 952) labels.  I could even see just RFC 1035 (disallows leading digits) if there is worry, but I think it's fine.

yifan-gu:
>
Using AC Name for things that have global namespaces makes sense (label keys, image names). Using something more restrictive for things with smaller scope is better, I think. Apps in a pod, mounts, volumes, ports, etc. They are all relative to their pod anyway.

I think this is a good point. Using DNS_LABEL in app.Name sounds like a good answer to https://github.com/appc/spec/pull/247 IMO

jonboulle:
Fixed by #398

Issue #351, opened 2015-05-04
Labels: consider namespacing well-known labels


thockin:
Consider explicitly namespacing well-known labels.  e.g. `appc-spec.org/version` instead of `version`.  In kubernetes we say that unqualified label keys are *explicitly* reserved for users.  Full stop, no exceptions.  without this, you can basically never add another "special" label after you go 1.0, lest you stomp on user-defined meanings.

jonboulle:
I'm definitely starting to lean towards this idea (ideally using `appc.io` for brevity). I think one of my few remaining reservations is that any user-friendly serialisation formats (like we suggest for [discovery](https://github.com/appc/spec/blob/master/discovery/parse.go#L30)) get preposterously awkward/unwieldy (`example.com/foo/bar,appc.io/version=latest,appc.io/arch=i386,appc.io/os=windows`)

@thockin what do you think of the compromise of namespacing them in the manifest but reserving the keywords at the top level for these well-known ones so at least we can continue assuming implementations can translate `version=latest` -> `appc.io/version=latest` ?

thockin:
You can claim reserved words now, but once people start using it you're
done :)

What iff you claim that some top-level prefix is reserved, e.g.
aci-version, aci-arch, aci-* and then claim the appc.io prefix for
less-critical labels.

Or you could just make these critical labels into fields :)

PR #398, opened 2015-05-08
spec: Propose AC Identifier Type. And redefine AC Name Type.

yifan-gu:
AC Identifier Type's format is the same as the old AC Name Type,
which contains "/" and ".", it is used for fields that have global
namespace, such as image name and image labels.

AC Name Type's removes "/" and "." from the AC Identifier Type.
This type is used for other fields that can be identified relatively
within a pod, image, or app context, such as app name, port name,
volume name.


yifan-gu:
First, I am bad at naming.
I have thought about an alternative that create `AC Label Type` as for names that do not have `/` or `.`.
But then we will not be able to use `AC Label Type` for labels, so I gave up. Instead, I think maybe renaming the `AC Name` to something more `global` is more reasonable ?

yifan-gu:
This tries to fix #342 #247 and partially #351
/cc @thockin @jonboulle @philips

yifan-gu:
This tries to fix #342 #247 and partially #351
/cc @thockin @jonboulle @philips

jonboulle:
How does this address #351?

jonboulle:
Not sure on the naming front, will have to mull on it...

yifan-gu:
@jonboulle I think as we now have two types, the decision of which type to use for labels implicitly answers #351  No?

yifan-gu:
OK, its' not answering #351 in whole, it just marks we go with the direction of namespacing the labels or not.

jonboulle:
@yifan-gu what do you think of altering this to cover #406?

yifan-gu:
@jonboulle Yes, I think we just need to update the `AC Identifier` here

thockin:
This looks like a good step to me.

yifan-gu:
Updated:
AC Identifier: `^[a-z0-9]+([-._~/][a-z0-9]+)*$` (lowercase alphanumeric plus `- . _ ~ /`, must start with alphanumeric)
AC Name: `^[a-z0-9]+([-][a-z0-9]+)*$` (lowercase alphanumeric plus `- `, must start with aplhanumeric)

AC Identifier will be used for image names and label keys(see #351).
AC Name will be used in all other places.

yifan-gu:
I will change the implementation code if this looks good to us for now :)

SPEC.md
```diff
@@ -763,8 +779,8 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
-* **labels** (list of objects, optional) used during image discovery and dependency resolution. The listed objects must have two key-value pairs: *name* is restricted to the [AC Name](#ac-name-type) formatting and *value* is an arbitrary string. Label names must be unique within the list, and (to avoid confusion with the image's name) cannot be "name". Several well-known labels are defined:
+* **name** (string, required) a human-readable name for this App Container Image (string, restricted to the [AC Identifier](#ac-identifier-type) formatting). This is not expected to be unique (see the **version** label) but SHOULD have a URL-like structure to facilitate **[App Container Image Discovery](#app-container-image-discovery)**. If this image is resolved through the discovery process, this field MUST match the name used for discovery.
```
jonboulle: This is supposed to stay AC Name surely?

jonboulle: er, ignore me, reading backwards

jzelinskie:
sgtm, idk how this effects the discovery protocols though (i.e. escaping/encoding/decoding in order to include symbols like `-._~/`).

jonboulle:
@jzelinskie why would we need to escape/decode any of those symbols?

jonboulle:
@yifan-gu Sorry for the delay - this LGTM so I'd say go ahead with the schema changes.

jonboulle:
cc @krobertson

SPEC.md
```diff
@@ -552,19 +552,35 @@ Accessible at `$AC_METADATA_URL/acMetadata/v1/pod/hmac`
...
 
-### AC Name Type
+### AC Identifier Type
 
-An AC Name Type is restricted to lowercase characters accepted by the DNS [RFC1123](http://tools.ietf.org/html/rfc1123#page-13) and "/".
-An AC Name Type cannot be an empty string and must begin and end with an alphanumeric character.
-An AC Name Type will match the following [RE2](https://code.google.com/p/re2/wiki/Syntax) regular expression: `^[a-z0-9]+([-./][a-z0-9]+)*$`
+An AC Identifier Type is restricted to lowercase URI unreserved characters defined in [RFC3986](http://tools.ietf.org/html/rfc3986#section-2.3)
+An AC Identifier Type cannot be an empty string and must begin and end with an alphanumeric character.
+An AC Identifier Type will match the following [RE2](https://code.google.com/p/re2/wiki/Syntax) regular expression: `^[a-z0-9]+([-._~/][a-z0-9]+)*$`
```
thockin: so `v~-_._-~v` is a valid identifier.  Interesting.  We specifically said that identifiers are a relatively loose string (like you spec) preceeded by an optional DNS name.  Specifically, the existence of a / at all implies that the first part is a DNS subdomain.  It's a little tighter than this - maybe this is OK.  Just go through and think about the most ridiculous valid string a user could input.

jzelinskie:
@jonboulle I think it was simply confusion on my behalf. The current description for AC Identifier references the DNS RFC for its schema and I was under the impression that DNS was desired to be the "lowest common denominator" in terms of character sets such that a discovery protocol could be implemented on top of it.

SPEC.md
yifan-gu: @thockin Yeah, we need to address this... Thanks for catching!

yifan-gu:
Another problem I hit during the change is for the name of isolators.
e.g. `resource/block-bandwidth`, `os/linux/capabilities-retain-set`, which container `/` in the name, so it will not satisfy the `ACName`, and we have to use `ACIdentifier`. How do you guys think about this?

SPEC.md
jonboulle: @thockin do you have any ideas for a sensible prescription here?

jonboulle:
@yifan-gu sorry, you will need to rebase

`ACIdentifier` is fine for isolators since those are already hierarchically namespaced.

yifan-gu:
@jonboulle I have rebased. Take a look?

yifan-gu:
Ehh, I see, the spec is breaked into different sections..

yifan-gu:
Now rebased.

jonboulle:
LGTM, another pair of eyes would be good though.
@vbatts @krobertson

jonboulle:
or @mpasternacki ?

spec/types.md
```diff
@@ -2,19 +2,35 @@
...
+An AC Identifier Type is restricted to lowercase URI unreserved characters defined in [RFC3986](http://tools.ietf.org/html/rfc3986#section-2.3)
+An AC Identifier Type cannot be an empty string and must begin and end with an alphanumeric character.
+An AC Identifier Type will match the following [RE2](https://code.google.com/p/re2/wiki/Syntax) regular expression: `^[a-z0-9]+([-._~/][a-z0-9]+)*$`
 
 Examples:
 
 * database
 * example.com/database
 * example.com/ourapp
+* example.com/~user/app_v1
```
vbatts: nice.

vbatts:
So, this deprecates ACName, but it's still to be left as a type?

vbatts:
no. The annotations will still use ac-name-type, and pod names.

I'm not opposed to this identifier, there is just enough overlap that having both seems confusing.

yifan-gu:
@vbatts The original reason for having two types is that we want more strict names for those app names, port names etc. (in particular, we want to eliminate `/` `.` in these names #342)

mpasternacki:
LGTM

mpasternacki:
Though I'd nitpick about the name: I'd use new name for new type, especially that I'd expect "identifier" to be shorter than "name" – like "id" vs "full name". `ACIdentifier` is similar to current `ACName`, and proposed `ACName` is essentially a "short id" kind of type. I'd feel better about it if the names were reversed, but this is a cosmetic issue, so I won't argue about it.

vbatts:
LGTM

Issue #388, opened 2015-05-06
capabilities: consider dropping 'remove' method


globalcitizen:
Two clashing definitions are given for the list of linux kernel capabilities that should be made available to the container: one explicit, one implicit.

Presumably this is because the `..._remove` method is the traditional approach within LXC and therefore is the format expected within docker and other image types that `appc` is assumed to import from (and ultimately deprecate :question:). However, this is bad because it is non-deterministic.

A neater approach would be to convert `..._remove` sets to retain sets on import.

philips:
I understand the concern about the remove set and I am not a huge fan but I don't think we can convert the remove set to retain as the kernel will add capabilities overtime.

globalcitizen:
The very fact that the kernel will add capabilities over time is why an explicit `...retain` approach should be strenuously encouraged and `..._remove` forcibly deprecated.

Scenario:
 * Kernel version 1 has capabilities A, B and C.
 * Kernel version 2 adds capability D.

With a `..._retain` approach, a service explicitly specifying that it requires A and B capabilities will survive the upgrade to kernel version 2 without magically acquiring the new capability.

With a `..._remove` approach, the same service explicitly specifying that it does not require C capability will magically acquire capability D upon kernel upgrade.

It is known correct practice to "deny all, permit explicitly" within a security context. I hope this is clear.

Issue #336, opened 2015-05-04
OS-SPEC not specific enough about what apps will experience in-container


thockin:
OS-SPEC says nothing about /etc/hostname, /etc/hosts, /etc/mtab, or /etc/resolv.conf.  These are pretty important parts of most Linux systems, and hard to do some very fundamental things without.  More generally, this spec does not define enough (IMO) of what the app will experience when it runs.

- What will be reported for kernel version?  I assume simply the host's.
- What will be reported as hostname?
- Can it rely on /tmp? (assuming that comes from the image)
- Are cgroups exposed into the container (assuming not)?
- What assumptions can I make from /proc/self/cgroup?
- What about ulimits and sysctls?

There are many facets of running on linux that (IMO) we should pin down.

jonboulle:
Care to make some proposals in these areas?

thockin:
I am a bit bandwidth constrained at the moment, so I will not have time to kick off a sub-spec.  I know this is a place where @vmarmol and @rjnagal have already put some more detailed thought, not sure if any artifacts remain.

I do think that we need to say something about basic DNS - for example, it is a cornerstone of kubernetes service discovery to be able to inject DNS config into containers at startup time.  Similarly we should say SOMETHING about hostname or else make it a configurable.  Ulimits and sysctls are a really thorny problem - would be nice to drive some canonical list of what works and doesn't for namespaces.  Maybe loop in some of the Joyent world for a cross check on "runs linux containers but isn't linux".

PR #400, opened 2015-05-08
SPEC: clarify language around the capabilities flags

philips:
This essentially offers two options:

- Remove certain capabilities and include all others
- Retain certain capabilities and removea all others

Fix everything up to be consistent with the remove-set and retain-set
language.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -244,7 +244,7 @@ The first example is "capabilities" but this will be expanded to include things 
 
 **Parameters:**
 
-* **set** list of capabilities that will be removed from the process's capabilities bounding set
+* **set** list of capabilities that will be removed from the process's capabilities bounding set, all others will be included.
 
 ```json
 "name": "os/linux/capabilities-remove-set",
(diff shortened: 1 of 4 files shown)


SPEC.md
```diff
@@ -261,7 +261,7 @@ The first example is "capabilities" but this will be expanded to include things
 
 **Parameters:**
 
-* **set** list of capabilities that will be retained in the process's capabilities bounding set
+* **set** list of capabilities that will be retained in the process's capabilities bounding set, all others will be removed
```
jonboulle: Add a trailing period for consistency with the other.

jonboulle:
This is okay, but still underspecified. I would argue that _both_ methods are still somewhat implicit/non-deterministic; we need to make some kind of reference to what the "default" set would be.  I wouldn't mind addressing this via #336

philips:
@jonboulle Setting up a default set is just going to get us into a world of trouble as the Kernel adds more overtime. I am fine with saying the image must explicitly give us the CAPS it wants and remove the remove-set.

jonboulle:
@philips I didn't actually mean codifying a specific set in the spec, but at least saying that it would be in reference to the host kernel's default for example. Anyway I am still sympathetic to #388 , let's continue there.

jonboulle:
Merging this as it fixes the schema/spec divergence, will follow up to fix capabilities more better.

PR #402, opened 2015-05-10
Clarify encryption method and add a note about semantic security

namn:
Make it clear in the spec that the encryption method is based on OpenPGP symmetric AES cipher with 256-bit key. Also add a short note about encryption after compression and a recommendation for implementations to allow disablement of compression.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -79,6 +79,8 @@ This set of formats makes it easy to build, host and secure an ACI using technol
 - Image archives SHOULD be signed using PGP, the format MUST be ascii-armored detached signature mode.
 - Image signatures MUST be named with the suffix `.aci.asc`.
 
+**Security Note**: Compressing the image will often change the image size significantly. When encryption is applied after compression, it could lead to information leakage, that would not have been revealed without compression, due to such observable difference. Implementations supporting compression SHOULD include an option to disable compression.
+
 The following example demonstrates the creation of a simple ACI using common command-line tools.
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -75,10 +75,12 @@ This set of formats makes it easy to build, host and secure an ACI using technol
 - Image archives MUST have only two top-level pathnames, `manifest` (a regular file) and `rootfs` (a directory). Image archives with additional files outside of `rootfs` are not valid.
 - All files in the image MUST maintain all of their original properties, including timestamps, Unix modes, and extended attributes (xattrs).
 - Image archives MAY be compressed with `gzip`, `bzip2`, or `xz`.
-- Image archives MAY be encrypted with AES symmetric encryption, after (optional) compression/encryption.
+- Image archives MAY be encrypted using PGP symmetric encryption with 256-bit key AES cipher, after optional compression.
```
jonboulle: Why do we need to be so prescriptive here? I would rather leave it to the operator - since it's symmetric encryption it's going to be controlled at both ends so we should be less opinionated, imho.

namn: I assume you are referring to the "256-bit key" part. My initial intention was to spell out that the image is encrypted in OpenPGP format, as evidenced by the `gpg` command below this. The "256-bit key" part was added because that same command used it and I had wrongly assumed it to be authoritative. I can remove that part.

jonboulle:
Thanks! LGTM

Issue #348, opened 2015-05-04
Pod Schema ports is confusing


thockin:
It says that name is "name of the port in the image manifest" but the Ports block is not attached to an app, it is attached to the pod.  What does name reference?

jonboulle:
Urgh, looks like half of the port definition in the image manifest schema went AWOL. Bedtime, will fix in the morning..

jonboulle:
Fixed by #405

PR #405, opened 2015-05-12
spec: fix ports definition in schema

jonboulle:
- Add definitions of name, port, protocol fields to objects in the ports
 section of the ImageManifest
- Clarify informational purpose of the ports field
- Clarify in PodManifest that the port name is a reference to
 ImageManifest ports fields


jonboulle:
fixes #348

SPEC.md
```diff
@@ -748,9 +748,12 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
+    * **ports** (list of objects, optional) ports that this app will be listening on once started. This field is informational: example uses include helping users to discover the listening ports of the application, or indicating to executors ports that should be exposed on the host. This information could also optionally be used to limit the inbound connections to the container via firewall rules to only ports that are explicitly exposed. Each object can represent either a single port or a port range (contiguous set of ports).
+        * **name** (string, required, restricted to the [AC Name](#ac-name-type) formatting) descriptive name for this port; for example, "http" or "database". This field is used as a key in the [Pod Manifest](#pod-manifest-schema) when specifying ports to be forwarded from the host.
+        * **protocol** (string, required) transport layer (Layer 4) protocol that will be used. Should be one of "tcp" or "udp".
```
eyakubovich: I don't know if we want to restrict it to just "tcp" or "udp". Such a restrictive list will make it easier to know that a runtime can support the whole set but what if an application wants to take advantage of SCTP? I think it should list some well known L4 protocols ("tcp", "udp", "sctp", "dccp") as well as specify that it can be something else. If the runtime doesn't know what to do with it, it can error out (same as with `socketActivated`).

thockin: YAGNI

When we have a concrete case we can extend it.  The obvious case I see is to specify higher level protocols - HTTP, for example.  Saying that a port is HTTP is *far* more informative than saying it is TCP.  We don't (yet) do this in kubernetes but I think it will come post 1.0.

SPEC.md
```diff
@@ -924,6 +927,6 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
...
 * **annotations** (list of objects, optional) arbitrary metadata the executor will make available to applications via the metadata service. Objects must contain two key-value pairs: **name** is restricted to the [AC Name](#ac-name-type) formatting and **value** is an arbitrary string). Annotation names must be unique within the list.
-* **ports** (list of objects, optional) list of ports that will be exposed on the host.
-    * **name** (string, required) name of the port in the image manifest that will be exposed on the host (restricted to the [AC Name](#ac-name-type) formatting).
-    * **hostPort** (integer, required) port number on the host that will be mapped to the container port.
+* **ports** (list of objects, optional) list of ports that SHOULD be exposed on the host.
+    * **name** (string, required, restricted to the [AC Name](#ac-name-type) formatting) name of the port to be exposed on the host. This field is a key referencing by name ports specified in the Image Manifest(s) of the app(s) within this Pod Manifest; consequently, port names MUST be unique among apps within a pod.
```
thockin: I consider this a real shortcoming of kubernetes, frankly, and I wanted to fix it but it was hard, so we punted.  We may never get to fix it.

Specifically, you want containers to be arbitrarily composable but you have introduced an a priori unresolvable conflict.  If two containers both expose a port "http", your pod is invalid.  It's historically OK in kubernetes because Docker doesn't name ports, so it's the pod spec that names them, which is in the control of the pod deployer.  If you embed port names into images, we lose that control.

Options:
1a. always reference port names as appname.portname
1b. allow appname.portname but allow bare portname if there's no conflict
2a. demand that pod authors expose and name app ports in the pod spec
2b. allow pod authors to re-name app ports in the pod spec but allow pass-through if there's no conflict

others?

SPEC.md
eyakubovich: Re HTTP: I think it's a rat's nest to try to specify application level protocol. At least at transport level, there's still only a handful. We can argue if tcp and udp should be extended with a few more or not, but it's still a small set. There's a humongous number of application level protocols. There won't be anything reasonable that an ACE can do with this info -- it will be purely a comment.

eyakubovich: We can extend it later, that's not a problem. I don't feel that strongly about it but I think there are users of something other than tcp+udp out there and it would be nice not to cut them out.

jonboulle: I intentionally used **should** not **must** here. I will add a note to be clear that it can be something else and is up to the runtime to support/reject.

krobertson: Not really a port, but how would ICMP/ping fit in? With the description talking about executors potentially able to use it for firewalling, what if an image wishes to be pingable?

thockin: I don't think that is an image's decision.

SPEC.md
```diff
@@ -748,9 +748,12 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
+    * **ports** (list of objects, optional) ports that this app will be listening on once started. This field is informational: example uses include helping users to discover the listening ports of the application, or indicating to executors ports that should be exposed on the host. This information could also optionally be used to limit the inbound connections to the container via firewall rules to only ports that are explicitly exposed. Each object can represent either a single port or a port range (contiguous set of ports).
+        * **name** (string, required, restricted to the [AC Name](#ac-name-type) formatting) descriptive name for this port; for example, "http" or "database". This field is used as a key in the [Pod Manifest](#pod-manifest-schema) when specifying ports to be forwarded from the host.
+        * **protocol** (string, required) transport layer (Layer 4) protocol that will be used. Should be one of "tcp" or "udp".
+        * **port** (integer, required) port number that will be used. Must be >=0 and <=65535.
```
krobertson: Specifying 0 may be odd. If the process is using 0, then the executor will not have a sound way to represent it in the pod manifest, such as if another process wanted to connect to it. Should it maybe be >0? Having the executor pick the port may become a rats nest.

SPEC.md
krobertson: Maybe, but thought I'd pose the question.

SPEC.md
thockin: You're right that it should be > 0, unless we want auto-allocation, which I don't think was the intent?

SPEC.md
jonboulle: @thockin 2b is implicitly permitted today. Do you think that's sufficient?

SPEC.md
jonboulle: This was me lazily converting the existing [port schema validation](https://github.com/appc/spec/blob/master/schema/types/port.go#L50) without thinking. I do not see auto-allocation as being useful here. 

@eyakubovich you implemented that code, what's your thinking? :-)

SPEC.md
philips: @jonboulle shouldn't the tuple be appName and portName? Then everything is super clear.

jonboulle: @philips what tuple?

SPEC.md
eyakubovich: I'd like not to implement auto-allocation but I have to admit that it may be useful in some cases. For example, to expose multiple services to the host and then locate them through DNS+SRV records. Can we call out 0 as auto-allocation as MAYBE supported by the ACE?

eyakubovich: Strike out my previous comment -- this is ImageManifest, not PodManifest. `>0` is what we want here.

SPEC.md
jonboulle: ping @philips

jonboulle:
I am landing this as it fixes the current (broken) state, and filed https://github.com/appc/spec/issues/424 to follow up the thread from https://github.com/appc/spec/pull/405#discussion_r30172875

PR #408, opened 2015-05-17
acirenderer: correctly handle overridden directories.

sgotti:
If an upper image overrides with another file type a directory in a
parent image, do not add the parent image directory's contents to the
rendered image.

For example, given an image A that depends on B:

B:
drwxr-xr-x ./
drwxr-xr-x ./rootfs
drwxr-xr-x ./rootfs/a
-rw-r--r-- ./rootfs/a/file01
drwxr-xr-x ./rootfs/a/b
-rw-r--r-- ./rootfs/a/b/file02

A:
drwxr-xr-x ./
drwxr-xr-x ./rootfs
-rw-r--r-- ./rootfs/a

The files from image B (`./rootfs/a/b`, `./rootfs/a/file01`,
`./rootfs/a/b/file02`) shouldn't be added to the rendered image.

### DIFF
--- a/pkg/acirenderer/acirenderer.go
+++ b/pkg/acirenderer/acirenderer.go
@@ -90,7 +90,7 @@ func GetRenderedACIFromList(imgs Images, ap ACIProvider) (RenderedACI, error) {
 		return nil, fmt.Errorf("image list empty")
 	}
 
-	allFiles := make(map[string]struct{})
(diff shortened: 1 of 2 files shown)


iaguis:
LGTM

jonboulle:
lgtm, thanks!

PR #409, opened 2015-05-17
Fix twitter handle for Brandon Philips...

amouat:


jonboulle:
Thanks for the PR, but these are GH handles, not Twitter handles

amouat:
Ahh, oops. Sorry!

PR #410, opened 2015-05-18
SPEC.md: clarify timestamps

### LINKED ISSUE #330: spec: standard time string format
In the `annotations`, the value for key `created` is RFC3339.

Should we make a section for standard formats like this?

vbatts:
Closes #330


jonboulle:
LGTM. Some of the type references should be more consistent throughout the schema but that's something we can address more generally later. Thanks.

PR #411, opened 2015-05-18
*: v0.5.2

jonboulle:
It's been a while since the last tagged appc release and there are [a lot of
changes in here](https://github.com/appc/spec/compare/v0.5.1...master). However, it's almost exclusively wording tweaks and
clarifications and as far as I can tell there shouldn't be any breaking changes
in here. As such, I'm proposing we cut this as 0.5.2. 

This is in anticipation of several large changes under discussion or in flight ([e.g.](https://github.com/appc/spec/pull/397)) which will
definitely break compatibility and hence the next release will be 0.6.0; one of
the goals with this release is to minimise the diff there.


vbatts:
LGTM

PR #412, opened 2015-05-19
discovery: use a copy of app before adding defaultVersion label.

sgotti:
The discovery.App struct is passed by value to the doDiscover function
but the App.Labels map references the same underlying data structures of
the passed App.Labels map. So adding the "version" label will changes
the passed App.Labels.
Fix this using a copy of discovery.App.


jonboulle:
LGTM, thanks!

PR #413, opened 2015-05-20
discovery: Update the package to expose the http.Client

krobertson:
This changes the HTTP handling within the discovery package to make its
http.Client an exported field. This has a couple of use cases where consumers
of the package may want to customize the handling, such as:

* Adjusting the dial handler with custom timeouts
* Altering the http.Transport on the client for testing outside of the
  discovery package.
* Altering the proxy used for discovery requests without setting altering the
  proxy settings for the whole process.

This also updates the internal test handling to use an internal interface that
is used to wrap http.Client.Get(). With this, the way the existing tests were
mocking out requests is preserved.

### DIFF
--- a/discovery/http.go
+++ b/discovery/http.go
@@ -14,10 +14,19 @@ const (
 )
 
 var (
-	// httpGet is the function used by discovery to retrieve URLs; it is
+	// Client is the default http.Client used for discovery requests.
+	Client *http.Client
+
+	// httpGet is the internal object used by discovery to retrieve URLs; it is
 	// defined here so it can be overridden for testing
-	httpGet func(url string) (resp *http.Response, err error)
+	httpGet httpGetter
 )
+
+// httpGetter is an interface used to wrap http.Client for real requests and
+// allow easy mocking in local tests.
+type httpGetter interface {
+	Get(url string) (resp *http.Response, err error)
+}
 
 func init() {
 	t := &http.Transport{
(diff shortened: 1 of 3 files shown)


jonboulle:
@krobertson What if we just exposed HttpGet and allow that to be replaced instead? Then for your use case, people can just set it to the Get of their custom Client. Seems like it'd be simpler and alleviate one level of indirection.

krobertson:
It could. For some reason, I was thinking that exposing the Client was more idiomatic, but can't recall which other libraries had seen it in before. And then just wrapped the existing httpGet handling.

jonboulle:
Off the top of my head it doesn't seem more idiomatic, but I'd gladly be
shown otherwise - don't feel strongly about it.

krobertson:
One concern may be that to set `HttpGet`, you'd have to recreate a new client, so doing something like setting up a proxy would also clear the existing timeout handling. Setting the func point means you aren't just altering one thing.

discovery/http.go
```diff
@@ -14,11 +14,20 @@ const (
 )
 
 var (
-	// httpGet is the function used by discovery to retrieve URLs; it is
+	// Client is the default http.Client used by for discovery requests.
```
jonboulle: s/by for/for/

jonboulle:
OK, I'm happy enough to just accept this, but one typo to fix :-)

discovery/http.go
krobertson: Doh, fixed.

jonboulle:
lgtm, thanks

PR #414, opened 2015-05-22
Add arm64 big endian label and checks.

glevand:


jonboulle:
Whoops, we were told about this one but evidently I forgot to add it. Thanks!

PR #419, opened 2015-05-28
pkg/tarheader: Explicitly cast st.Rdev to uint64 to make it build on FreeBSD

mpasternacki:
FreeBSD's `st.Rdev` is `uint32`, adding explicit cast up to `uint64` to get signature working.


jonboulle:
LGTM, thanks.

PR #421, opened 2015-05-31
Allow implementations to extend discovery

justinsb:
Discovery can be extended arbitrary, but only in response to operator
configuration.  Implementations are encouraged to adopt mechanisms similar to
the existing meta-discovery mechanism, and implementations must support the
public mechanisms by default.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -479,6 +479,65 @@ All HTTP redirects MUST be followed when the discovery URL returns a `3xx` statu
 
 Discovery URLs that require interpolation are [RFC6570](https://tools.ietf.org/html/rfc6570) URI templates.
 
+### Extensible Discovery
+
+The above methods (simple discovery & meta discovery) must be supported by a
+conformant ACI implementation.  Some advanced use cases, such as private image
+repositories, or maintaining mirrors of image repositories, may require
+different mappings of image names to storage locations.  A conformant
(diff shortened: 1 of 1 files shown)


justinsb:
For discussion: allow implementations to define alternative discovery mechanisms that support use-cases such as private repositories and custom mirrors.  As long as these are opt-in, we need not define them in the spec, but we should probably do so as it becomes more clear exactly what they look like.

If we like the _idea_ here (if not the implementation!) then I will try to implement private-repositories on S3 for rkt, to verify that this is realistic (and see what problems are hit in practice)

krobertson:
How does something like the S3 use case differ from having and endpoint that returns a 3xx URL to the S3 location? Or using an http proxy to handle mirroring?

justinsb:
@krobertson A 3xx URL would still require maintaining a server with a valid SSL certificate.  (Though this is a great idea for offloading traffic from that server).  Also, for repositories that require auth, S3 has an authentication mechanism and it probably makes sense to just leverage that.

For public image repositories, I would imagine they would just use the existing discovery mechanism but specify the "awkward" S3 url as their backing-store i.e. `https://s3.amazonaws.com/mybucket/...`  So I don't think they need explicit S3 support.

An http proxy is a nice idea, though I think that the adaptation to support S3 is relatively minor and would mean you wouldn't need the extra moving-piece.

PR #423, opened 2015-06-04
spec: Updated links between different definations.

yifan-gu:


jonboulle:
lgtm, thanks muchly @yifan-gu

PR #422, opened 2015-06-09
spec: dependencies should have "size"/"length"

jonboulle:
Dependency references should have an optional "size" or "length" field to allow clients to prevent against DOS attacks  ([e.g.](https://github.com/theupdateframework/tuf/blob/d7fd61234e423a30b5ab10253f347fbd515180af/docs/tuf-spec.txt#L138)) when discovering and retrieiving them.


krobertson:
LGTM.

As an aside, it might be worth extending the discovery handling to cover doing HEAD requests to the ACI endpoint to look up the size as well. LMK, I could write that up.

jonboulle:
@krobertson sounds great!

PR #429, opened 2015-06-11
SPEC: typo

HuKeping:
Make it consistent with others.

Signed-off-by: Hu Keping <hukeping@huawei.com>


jonboulle:
Thanks for the patch, but this is grammatically slightly awkward; if anything I would rather go in the other direction of having the zeroth article for all of them. How does that sound?

HuKeping:
sure

Issue #364, opened 2015-05-04
Volumes: three-level volumes definitions are confusing


thockin:
The whole three-level volume thing is confusing to me (and I dreamed up the kubernetes two-level system, so ...).  I think this warrants some energy in simplifying, but I am guessing this ties into concerns about "what" to run and "how" to run it.

Specifically, the need to expresss mountpoints on an image: Does this mean that I can not do clever things with mount points and extant images?  E.g. bind mounting host stuff in to random places... This has been very powerful in Dockerland - e.g. injecting a binary via a volume, using that as the entrypoint, and slingshotting to the "real" entrypoint, all without the container image's participation.

I get the desire to say "I only accept volumes on path X" but I think that is not sufficient to prevent trouble, and I think it is over-constraining.  At the very least, it deserves some thorough explanation.  But having to thread volume info into three places is just too much for my poor brain.

jonboulle:
This issue (arbitrary mounts) emerged recently in our implementation:
https://github.com/coreos/rkt/issues/761
https://github.com/coreos/rkt/pull/853

jonboulle:
> Does this mean that I can not do clever things with mount points and extant images? E.g. bind mounting host stuff in to random places... This has been very powerful in Dockerland - e.g. injecting a binary via a volume, using that as the entrypoint, and slingshotting to the "real" entrypoint, all without the container image's participation.

Well, since everything is overridable by the executor via PodManifest, you can indeed do this and be compliant with the spec - the proposed `--inject-volumes` flag for rkt does exactly this: https://github.com/coreos/rkt/pull/853/files#diff-0321d59a96842ba88f277bbab5b32852R131

So the current spec definitely allows such arbitrary behaviour even if it's not immediately obvious. Would just calling this out more clearly seem sufficient? Or would you still prefer more drastic simplification?

thockin:
This is where my issue list gets fun.  I certainly don't intent to impose my worldview on everyone, but I *do* find the three levels to be awkward.  Is the goal to give ACI creators the ability to say "I expect a volume on /foo and if I don't have one, it is an error" ? That's how it reads.  If so, I'm not sure I buy the utility.  Volumes are not created equal, but I don't think you want ACI creators dictating "here I want a tmpfs".  I think that really becomes part of the HOW, not the WHAT.

Or is there some other problem that the extra indirection solves, in exchange for its cost?

ibukanov:
[...]

PR #431, opened 2015-06-11
spec: app filesystem clarifications

### LINKED ISSUE #425: spec: clarification needed for dependencies unpacking
> An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
> - The `rootfs` contained in the ACI is extracted
> - If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted, in the order in which they are listed
> - If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed

What should happen if one of dependencies include a path that is already present in the image? Straightforward implementation of the above process would overwrite the image's file with one from the dependency. On the other hand, common sense would suggest that this reading is not intentional; this definitely is not the behaviour I would expect.

I would expect that a dependency is similar to a "parent image" in Docker: a Ruby application ACI extends (and depends on) a Ruby runtime ACI, which in turn extends (and depends on) bare distribution ACI. Files from Ruby runtime ACI would overwrite files from the bare distribution, and files from the Ruby application ACI would overwrite files from both Ruby runtime and bare distribution.

Please clarify: is my understanding of the use case wrong here? In either case, more explicit phrasing in this part of the spec would be useful.

If I am wrong, and dependencies shouldn't be used to provide parent/child inheritance like in Docker, is there any good approach to distribute only the new in the ACI if the "parent" image is discoverable?

mpasternacki:
- Rephrase ACI dependencies unpacking (hopefully fixes #425)
 - Clarify behaviour of mount points: nonexistent mount points are created, existing files are removed, existing directories' contents are discarded
 - Specify that symlinks are not a valid volume source (to protect against attacks on host)
 - Add explicit reference to RFC2119 in SPEC.md

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -17,7 +17,7 @@ The core goals of the specification include:
 
 ### Requirements
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://tools.ietf.org/html/rfc2119).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this and other documents of the specification are to be interpreted as described in [RFC2119](http://tools.ietf.org/html/rfc2119).
 
 ## Sections
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -15,6 +15,10 @@ The core goals of the specification include:
 * Using common technologies for cryptography, archiving, compression and transport
 * Using the DNS namespace to name and discover images
 
+### Requirements
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://tools.ietf.org/html/rfc2119).
+
 ## Sections
 
 The specification consists of several key sections; the goal is that each can be implemented independently, but are composable with one another.
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -26,9 +26,11 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
-- The `rootfs` contained in the ACI is extracted
-- If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted, in the order in which they are listed
-- If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed
+1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the app's filesystem, in the order in which they are listed.
+2. The `rootfs` contained in the ACI is extracted into the app's filesystem
+3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed
+
+If during rootfs extraction a target path is already present in app's filesystem from an earlier dependency, the previously extracted path MUST be overwritten. If the existing path is a symbolic link to a directory, the link MUST NOT be followed and it MUST be removed and replaced with the new path.
 
 Every execution of an app MUST start from a clean copy of this rendered filesystem.
 
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -26,11 +26,12 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
-1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the app's filesystem, in the order in which they are listed.
-2. The `rootfs` contained in the ACI is extracted into the app's filesystem
-3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed
 
-If during rootfs extraction a target path is already present in app's filesystem from an earlier dependency, the previously extracted path MUST be overwritten. If the existing path is a symbolic link to a directory, the link MUST NOT be followed and it MUST be removed and replaced with the new path.
+1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the target directory, in the order in which they are listed.
+2. The `rootfs` contained in the ACI is extracted into the target directory
+3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed from the target directory
+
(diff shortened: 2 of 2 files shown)


SPEC.md
```diff
@@ -15,6 +15,10 @@ The core goals of the specification include:
 * Using common technologies for cryptography, archiving, compression and transport
 * Using the DNS namespace to name and discover images
 
+### Requirements
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://tools.ietf.org/html/rfc2119).
```
jonboulle: nit, maybe "in this and other documents of the specification" ?

mpasternacki: I thought of the spec to be a single logical document split into multiple files, but yes, telling this explicitly seems good. Changed.

spec/ace.md
```diff
@@ -26,9 +26,11 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
...
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
-- The `rootfs` contained in the ACI is extracted
-- If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted, in the order in which they are listed
-- If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed
+1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the app's filesystem, in the order in which they are listed.
+2. The `rootfs` contained in the ACI is extracted into the app's filesystem
+3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed
+
+If during rootfs extraction a target path is already present in app's filesystem from an earlier dependency, the previously extracted path MUST be overwritten. If the existing path is a symbolic link to a directory, the link MUST NOT be followed and it MUST be removed and replaced with the new path.
```
jonboulle: s/in app's filesystem/the filesystem/ or similar

General nit, would you mind following the one-sentence-per-line convention for now? (better for reviewability of subsequent changes)

spec/ace.md
```diff
@@ -49,6 +51,14 @@ Volumes that are specified in the Pod Manifest are mounted into each of the apps
 For example, say that the worker-backup and reduce-worker both have a `mountPoint` named "work".
 In this case, the executor will bind mount the host's `/opt/tenant1/work` directory into the `path` of each of the matching "work" `mountPoint`s of the two app filesystems.
 
+If the target `path` does not exist in the rendered filesystem, it SHOULD be created, including any missing parent directories. If the target `path` is a non-empty directory, its contents SHOULD be discarded (e.g. obscured by the bind mount). If the target `path` refers to a file, ACE SHOULD remove that file and create a directory in its place.
+
+Mount point `path` directories that ACE creates SHOULD be owned by UID 0 and GID 0, and have access mode `0755` (`rwxr-xr-xr-x`). ACE implementation MAY provide a method for administrator to specify different permissions on a per-pod basis.
+
+ACE MUST NOT create any paths in the host file system, and is REQUIRED to consider missing volume source paths as an error. If an underlying operating system supports it ACE MAY implement single-file volumes.
```
jonboulle: First stipulation here sounds a little overly prescriptive, what's your motivation?

spec/ace.md
```diff
@@ -49,6 +51,14 @@ Volumes that are specified in the Pod Manifest are mounted into each of the apps
 For example, say that the worker-backup and reduce-worker both have a `mountPoint` named "work".
 In this case, the executor will bind mount the host's `/opt/tenant1/work` directory into the `path` of each of the matching "work" `mountPoint`s of the two app filesystems.
 
+If the target `path` does not exist in the rendered filesystem, it SHOULD be created, including any missing parent directories. If the target `path` is a non-empty directory, its contents SHOULD be discarded (e.g. obscured by the bind mount). If the target `path` refers to a file, ACE SHOULD remove that file and create a directory in its place.
```
jonboulle: While we're at it perhaps we should talk about the case of multiple volumes having conflicting paths?

spec/ace.md
```diff
@@ -49,6 +51,14 @@ Volumes that are specified in the Pod Manifest are mounted into each of the apps
...
+
+Mount point `path` directories that ACE creates SHOULD be owned by UID 0 and GID 0, and have access mode `0755` (`rwxr-xr-xr-x`). ACE implementation MAY provide a method for administrator to specify different permissions on a per-pod basis.
```
jonboulle: Should this be inverted to suggest that the permissions would be at the executor's discretion but may be 0/0/0755 as a fall-back ?

spec/ace.md
mpasternacki: If ACE can be tricked into actively altering host filesystem, this may lead to security vulnerabilities. It seems to be common sense for implementation, but the security concerns are grave enough to mention it in the spec. A similar issue would be requiring ACE to refuse to use symbolic links as sources (or even to refuse to use any path that has a symbolic link in it as a volume source); would this be a good place to add that part, or should I go to pod.md where volume is described?

If you are concerned about the spec being too strict, I'm fine with downgrading this requirement to a SHOULD level, with an addition that if the implementation supports altering host filesystem, it should be possible to turn it off.

spec/ace.md
mpasternacki: I'm concerned about consistency: if we reverse the ordering, this means that the path can be created with any permissions, which can give different results on different implementations (or even in different installations of the same implementation, depending on GID or umask of ACE process).

Consistency of the results can be partially circumvented by just including empty directory in the ACI, but this won't help with mount points introduced by app override in pod manifest: there's no way to make these behave consistently without specifying suggested permissions.

We can also simply say that it's an error if the mount path doesn't exist, but I think that would be too strict: flexibility of mounting volumes anywhere via app overrides seems to be a feature rather than a bug.

spec/ace.md
mpasternacki: Fixed. Will keep sentences to separate paragraphs too.

spec/ace.md
mpasternacki: Yes, I didn't think of this one. I'd be inclined to just specify that nested volumes are invalid and should be rejected by ACE. Or should mount order be sorted so that volumes can overlap?

This is an issue I'm having right now: I have an ACI to build FreeBSD ports in a completely clean base system (to find dependency issues). I mount `/usr/ports` from base system read-only, and `/usr/ports/distfiles` (downloaded source tarballs) read-write. I mount distfiles as `/usr/ports.distfiles` in the pod and change the location in `/etc/make.conf`, but I'd be in a tough spot if this wasn't configurable.

On the other hand, allowing overlapping ports feels like opening a can of worms, tricky implementation, special cases, and possible vulnerabilities.

spec/ace.md
jonboulle: OK, I'm convinced.

jonboulle: > flexibility of mounting volumes anywhere via app overrides seems to be a feature rather than a bug. 

Related, would appreciate your input/proposal for https://github.com/appc/spec/issues/364

spec/ace.md
jonboulle: I agree that overlapping volumes is a can of worms and would prefer to avoid it. +1 to specifying that nested are invalid.

spec/ace.md
jonboulle: > If ACE can be tricked into actively altering host filesystem, this may lead to security vulnerabilities. 

I would prefer this be downgraded to a SHOULD because there are legitimate use cases for this behaviour and I don't think we should necessarily preclude them in the spec. I'm happy with the addition you suggest.

jonboulle:
@mpasternacki sorry for the delay, two things I'd like to see tweaked a bit then I'd say this looks good to go, thanks!

jonboulle:
@mpasternacki ping

jonboulle:
ping again!

jonboulle:
Moving this to #494

PR #433, opened 2015-06-11
Make build scripts work on FreeBSD

mpasternacki:
Changed build and test scripts to work on FreeBSD. Removed harcoded setting GOOS/GOARCH to Linux/amd64; if cross-compilation is needed, will add an argument to set the target arch.

Are there any instructions on how to use the validator ACIs? A validator pod manifest, maybe?

### DIFF
--- a/ace/build_aci
+++ b/ace/build_aci
@@ -1,7 +1,8 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash
 #
 # Builds an ACI containing a go implementation of an ACE validator
 #
+set -eu
 
 PREFIX="ace"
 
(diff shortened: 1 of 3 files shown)


jonboulle:
:+1: nice

jonboulle:
> Are there any instructions on how to use the validator ACIs? A validator pod manifest, maybe?

We have this - https://github.com/appc/spec/blob/master/README.md#validating-app-container-executors-aces - but it's loooong been on my todo list to split it out. Pod manifest is also a good idea.

mpasternacki:
The instruction just says that the validator images are there, doesn't say how to run them :( if you have a working pod manifest file you can publish or send me (`$FIRSTNAME@$LASTNAME.net`), I'll process that into documentation. I can also figure that out by trial and error, but this could take more time and some questions.

jonboulle:
urgh, sorry, I was confusing it with our rkt instructions - https://github.com/coreos/rkt/blob/master/Documentation/app-container.md#validating-rkt-as-an-appc-implementation 

It is a little implementation-dependent - I think we just need to capture that a) it expects/requires a metadata service (which is another TODO - that should be configurable), and b) it expects an empty volume mounted at "database".

the pod manifest that rkt generates from the above command looks like this:
```
{
    "acKind": "PodManifest",
    "acVersion": "0.5.4",
    "annotations": null,
    "apps": [
        {
            "annotations": [
                {
                    "name": "created",
                    "value": "2014-10-27T19:32:27.67021798Z"
                },
                {
                    "name": "authors",
                    "value": "Carly Container <carly@example.com>, Nat Network <nat@example.com>"
                },
                {
                    "name": "homepage",
                    "value": "https://github.com/containers/standard"
                },
                {
                    "name": "documentation",
                    "value": "https://github.com/containers/standard/blob/master/README.md"
                },
                {
                    "name": "lorem",
                    "value": "ipsum"
                }
            ],
            "image": {
                "id": "sha512-3930490f1ec99cfc49f2b19d6cec3d1ff7a92479a8313d6bded3d60de04e0d5b",
                "name": "coreos.com/ace-validator-main"
            },
            "name": "coreos.com/ace-validator-main"
        },
        {
            "image": {
                "id": "sha512-3ab70b05b72133592ca5d9034981f8546a13ab8363a3f5309398c30daa9c27cf",
                "name": "coreos.com/ace-validator-sidekick"
            },
            "name": "coreos.com/ace-validator-sidekick"
        }
    ],
    "isolators": null,
    "ports": [],
    "volumes": [
        {
            "kind": "host",
            "name": "database",
            "source": "/tmp"
        }
    ]
}
```

PR #439, opened 2015-06-13
.travis.yml: update the cmd/vet and cmd/cover discovery URL. Add go 1.4

yifan-gu:


jonboulle:
#438

jonboulle:
Yours is better.

PR #440, opened 2015-06-14
spec: Add schema 'armv6l' as valid arch for linux

### LINKED ISSUE #417: schema: add armv6l as valid arch for linux (Raspberry Pi 1)
Some arm architectures are already added by #304 for #284. But according to https://github.com/coreos/rkt/issues/730#issuecomment-105490008 this is not enough for Raspberry Pi 1:

> @alban, if someone would like to use rkt on a Raspberry Pi 1 (there are about 5 million devices out there), the minimum supported ARM version should be ARMv6. This is also the default when building a GOLANG binary with GOARCH=arm. For this reason I strongly recommend to support armv6l too! BTW, a armv6l binary is compatible and runs absolute fine on a `armv7l´ machine - but not the other way around.

/cc @DieterReuter

The current support is:
```
var ValidOSArch = map[string][]string{
        "linux":   {"amd64", "i386", "aarch64", "armv7l", "armv7b"},
        "freebsd": {"amd64", "i386", "arm"},
        "darwin":  {"x86_64", "i386"},
}
```

DieterReuter:
This PR brings support for a Raspberry Pi 1. In contrast a Raspberry 2 uses "armv7l" which is already supported in `spec`.

With some more detailed tests I had problems to import ARM Docker images with `docker2aci`. This was due to the fact that Docker uses only `"Os": "linux"` and "`Architecture": "arm"` (can be determined with `docker insect <imageid>`. These issues could be easily solved by adding "arm" as a valid arch.

Fixes #417

### DIFF
--- a/schema/types/labels.go
+++ b/schema/types/labels.go
@@ -7,7 +7,7 @@ import (
 )
 
 var ValidOSArch = map[string][]string{
-	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv7l", "armv7b"},
+	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv6l", "armv7l", "armv7b"},
 	"freebsd": {"amd64", "i386", "arm"},
 	"darwin":  {"x86_64", "i386"},
 }
--- a/schema/types/labels_test.go
+++ b/schema/types/labels_test.go
@@ -32,6 +32,14 @@ func TestLabels(t *testing.T) {
 			`bad arch "arm64_be" for linux`,
 		},
 		{
+			`[{"name": "os", "value": "linux"}, {"name": "arch", "value": "arm"}]`,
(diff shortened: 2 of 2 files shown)


jonboulle:
What does `"Architecture": "arm"` mean on Docker images? It's too broad to be useful and I'm -1 on adding it here, would rather address that in docker2aci.

DieterReuter:
@jonboulle OK, I agree - and it's better to keep the spec clean. I can change the PR to include `armv6l` only, or should I close this PR and file a new one?

jonboulle:
@DieterReuter Updating the PR and force pushing this branch is fine, thanks

DieterReuter:
@jonboulle Done.

jonboulle:
LGTM

yifan-gu:
LGTM

PR #442, opened 2015-06-15
spec/schema: change annotations to ACIdentifier

jonboulle:
https://github.com/appc/spec/commit/555e3e9b5e78781146eec8ff2cc30e5a27a0dafd
redefined ACName to a more restrictive syntax, and changed some fields
(like name, isolators, etc) to a new ACIdentifier type with lighter
restrictions similar to the previous ACName. Since annotations is a
field we expect to be namespaced it makes more sense for it to also use
the ACIdentifier type.


ace/validator.go
```diff
@@ -350,7 +350,7 @@ func validateAppAnnotations(metadataURL string, pm *schema.PodManifest, app *sch
 			r = append(r, err)
 		}
 
-		lbl, err := types.NewACName(key)
+		lbl, err := types.NewACIdentifier(key)
 		if err != nil {
 			r = append(r, fmt.Errorf("invalid annotation label: %v", err))
```
yifan-gu: Should we replace `label` with `name` here to stay consistent with the doc?

schema/types/annotations_test.go
```diff
@@ -199,8 +199,8 @@ func TestAnnotationsSet(t *testing.T) {
 
 	a.Set("dog", "woof")
 	w = Annotations{
-		Annotation{ACName("foo"), "bar"},
-		Annotation{ACName("dog"), "woof"},
+		Annotation{ACIdentifier("foo"), "bar"},
+		Annotation{ACIdentifier("dog"), "woof"},
```
yifan-gu: maybe add some test cases such as `example.org/foo/foo_bar`

jonboulle:
@yifan-gu updated

ace/validator.go
```diff
@@ -287,7 +287,7 @@ func validatePodAnnotations(metadataURL string, pm *schema.PodManifest) results
 			r = append(r, err)
 		}
 
-		name, err := types.NewACName(key)
+		name, err := types.NewACIdentifier(key)
 		if err != nil {
 			r = append(r, fmt.Errorf("invalid annotation label: %v", err))
```
yifan-gu: here s/label/name/ as well?

jonboulle: fixed

yifan-gu:
LGTM. Want another pair of eyes, but I think German team is :sleeping: this time...

iaguis:
Noch nicht! LGTM

jonboulle:
I would like to get another maintainer ack before landing.

schema/types/annotations_test.go
philips: +1

philips:
lgtm, I agree with having one more test as @yifan-gu pointed out

jonboulle:
I did add a case @ https://github.com/appc/spec/pull/442/files#diff-e0d5d10f9ae10b151ee373931cbaa0a3R214

PR #444, opened 2015-06-16
schema/pod: add GetAppByImageName function

iaguis:
This function retrieves an app from the Pod's AppList that has the
specified name in its image.

This function was added to rkt in https://github.com/coreos/rkt/pull/1049

### DIFF
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -112,6 +112,19 @@ func (al AppList) Get(name types.ACName) *RuntimeApp {
 	return nil
 }
 
+// GetAppByImageName retrieves an app from the AppList whose image has the
+// specified name; if there is no such app, nil is returned. The returned
+// *RuntimeApp MUST be considered read-only.
+func (al AppList) GetAppByImageName(name types.ACIdentifier) *RuntimeApp {
+	for _, a := range al {
+		if name.Equals(*a.Image.Name) {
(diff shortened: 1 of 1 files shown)


yifan-gu:
The problem is what if two apps have the same image? Though currently we don't support that yet, but we are not prohibiting it in the spec, and I assume after we replace the imageID with appName in rkt for app reference, then it's possible for user to have same image for multiple apps.

jonboulle:
@yifan-gu is right - this should really be `GetAppByName` not ImageName

yifan-gu:
@jonboulle We already have that, called `Get`. The only place for `GetAppByImageName` in the bump pr is https://github.com/coreos/rkt/pull/1049/files#diff-0a939c0d8e5b66c4a2e56cc2d890adb4R448 which can be avoided by getting the app name from the URL

yifan-gu:
So I think we don't have to have this to get 0.6.1

jonboulle:
sorry, misread!

jonboulle:
Still -1 on adding this to the spec tooling due to the issue yifan mentioned (no prohibition on duplicate images)

iaguis:
Let's close this then, rkt doesn't even need the function anymore and your concerns make sense. Thanks!

Issue #183, opened 2015-02-09
spec: clarify empty volume kind


jonboulle:
> kind (required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (writes go to container). "host" fulfills a mount point with a bind mount from a source.

"writes go to container" -> this is unclear as to whether this is shared between the apps in the container or not. I propose that empty means a (temporary)  directory that _is_ shared among the apps in the container. 

/cc @mpasternacki

philips:
@jonboulle The volume is shared if it is mapped into two apps.

cdaylward:
I presumed this would mean shared by all apps in the container (what I was aiming for). @jonboulle when you say temporary directory, do you mean a temporary directory (tmpnam()) on the host that is bind-mounted in?

jonboulle:
@philips The point is that the wording ("writes go to container") is ambiguous; since each app has its own rootfs, if we just allowed it to be a writeable/empty directory it would not be shared. 

@cdaylward yes, that is essentially what I mean

mpasternacki:
To clarify: a "temporary directory", as I understand it, would mean a directory whose lifetime is *at least* equal to lifetime of a container (i.e. survives container restart and host restart). Ideally, it could be "anonymously" passed from container to container, in a way similar to `--volumes-from` of Docker.

To clarify: Docker, if volume's source directory is not provided, creates a new unique directory and uses it as the volume source. When one creates a new container, a `--volumes-from=CONTAINER` can be provided, which reuses all host-bound and anonymous volumes from an existing container. If a container is removed, and it holds the last remaining reference to an anonymous volume, the volume is deleted (at least in theory). This allows, for example, for database upgrades without explicitly maintaining an index of data directories on the host. A simplified example:

```
docker run --name=postgres93 -d postgresql:9.3
# A /var/lib/postgresql volume is implicitly created, because the volume is declared in Dockerfile
docker stop postgresql93
docker run --name=postgresl94 --volumes-from=postgres93 postgresql:9.4
docker rm postgresql93
```

The idea of the created directory being "temporary" needs to be more precise: lifetime of the created directory needs to be known. I believe it should be at least equal to lifetime of the container, if not maintained separately and reference-count-garbage-collected as Docker does.

jonboulle:
Reference: https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/volumes.md#emptydir

thockin:
[...]

Issue #276, opened 2015-04-01
spec: define pod lifecycle


jonboulle:
We need to address several aspects of pod lifecycles in the spec:

- Garbage collection lifecycle of the pod filesystem including:
   - [ ] Format of app exit codes and signal
   - [ ] The refcounting plan for resources consumed by the ACE such as volumes
- Lifecycle of apps within the pod including:
   - [ ] All exit or first to exit
   - [ ] Start and stop order

jonboulle:
Another aspect to this, which has come up in various discussion channels, is the relationship between the pod's _lifecycle_ and _identity_. What _is_ a pod's identity, and to what extent should it be immutable?

Should a pod be considered a resource envelope (similar to the _alloc_ in Google's [Borg](https://research.google.com/pubs/pub43438.html)), which individual applications can be added to or removed from? Or is a pod's identity more tightly linked to its constituent applications and their respective lifetimes; i.e. if a single application dies within a pod, or it is desired to stop one of the applications within a pod, then it is necessary to tear the entire pod down and create it again? (Coming back to the old petss vs cattle metaphor - the pod/resource-envelope that can be mutated at-will is the pet, the pod which must be destroyed + recreated is the cattle.)

Opening the door to things like "updating an application within a pod" fundamentally throws into question the whole idea of identity (by updating or stopping/starting every application in the pod in turn, we can easily end up with an entirely new set of applications in the pod, so what is really left to link it back to its original identity?). 

I am personally in strong favour of leaning towards the cattle side of the equation and retaining the pod as the fundamental scheduling unit; it is inherently easier to reason about, implement, and we don't need to define a complicated "inter-pod" API/ABI (i.e. what operations are permissible for individual apps vs on the pod as a whole). But there are practical arguments for allowing at least some operations to performed at a greater level of granularity.

/cc @thockin who has plenty of thoughts on this, /cc @vbatts

kelseyhightower:
@jonboulle Will there be support for just containers without pods?

jonboulle:
@kelseyhightower please elaborate
Am 27.04.2015 11:04 schrieb "Kelsey Hightower" <notifications@github.com>:

vbatts:
per chat, I think as pods become like templates, there ought to be a bullet that covers pod discovery and signature. At this point, they are seeming like their own `*.acp` or similar. Which would have a minimum a `/pod` manifest.

jonboulle:
[...]

PR #449, opened 2015-06-18
spec pods: Clarify empty volume kind

kibra:
The spec for the "empty" volume kind is unclear. See https://github.com/appc/spec/issues/183 for the issue/discussion around this.

**Note:** My change proposes the rather strong spec of: `lifetime of the empty volume == lifetime of the pod` 
I feel that this makes the use case of "empty" more clear than specifying `lifetime of the empty volume >= lifetime of the pod` as was suggested in https://github.com/appc/spec/issues/183#issuecomment-98582304 However, I don't have a strong opinion around this, and would be happy to change what I wrote.

### DIFF
--- a/spec/pods.md
+++ b/spec/pods.md
@@ -163,9 +163,11 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
     * **annotations** (list of objects, optional) arbitrary metadata appended to the app. The annotation objects must have a *name* key that has a value that is restricted to the [AC Name](types.md#ac-name-type) formatting and *value* key that is an arbitrary string). Annotation names must be unique within the list. These will be merged with annotations provided by the image manifest when queried via the metadata service; values in this list take precedence over those in the image manifest.
 * **volumes** (list of objects, optional) list of volumes which will be mounted into each application's filesystem
     * **name** (string, required) used to map the volume to an app's mountPoint at runtime. (restricted to the [AC Name](types.md#ac-name-type) formatting)
-    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (i.e., writes go to the app's chroot). "host" fulfills a mount point with a bind mount from a **source**.
+    * **readOnly** (boolean, optional, defaults to "false" if unsupplied) whether or not the volume will be mounted read only.
+    * **kind** (string, required) either:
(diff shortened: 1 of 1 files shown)


kibra:
This clarification is needed to unblock some rkt work: https://github.com/coreos/rkt/issues/1040

mpasternacki:
:+1: Question of a volume lifetime seems to be dependent on #276 (define "pod exits"), so I'd say it's out of scope now.

jonboulle:
LGTM. @thockin ?

jonboulle:
@kibra can you rewrite your commit message please - Clearify --> Clarify

kibra:
@jonboulle Fixed, thanks for the catch.

kibra:
@jonboulle Ping. I'm not super happy with my language here, but I think it's a least a step forward. Let me know if there's something you think I should change.

spec/pods.md
```diff
@@ -163,9 +163,11 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
...
 * **volumes** (list of objects, optional) list of volumes which will be mounted into each application's filesystem
     * **name** (string, required) used to map the volume to an app's mountPoint at runtime. (restricted to the [AC Name](types.md#ac-name-type) formatting)
-    * **kind** (string, required) either "empty" or "host". "empty" fulfills a mount point by ensuring the path exists (i.e., writes go to the app's chroot). "host" fulfills a mount point with a bind mount from a **source**.
+    * **readOnly** (boolean, optional, defaults to "false" if unsupplied) whether or not the volume will be mounted read only.
+    * **kind** (string, required) either:
+        * **empty** - creates an empty directory on the host and bind mounts it into the container. All containers in the pod share the mount, and the lifetime of the volume is equal to the lifetime of the pod (i.e. the directory on the host machine is removed when the pod exits)
```
jonboulle: > (i.e. the directory on the host machine is removed when the pod exits)

This makes me a little uneasy. How about "the directory on the host machine is removed when the pod's filesystem is garbage collected"

jonboulle:
@kibra One more tweak then LGTM

spec/pods.md
kibra: @jonboulle Agreed. That makes much more sense.

kibra:
@jonboulle Thanks for the feed backup. I've updated my PR.

jonboulle:
thanks!

Issue #373, opened 2015-05-05
Metadata endpoints are too vague


schancel:
Should include content type and JSON schema references for the returned values.

E.g. is `/acMetadata/v1/pod/uuid` supposed to be a JSON object, raw bytes, base64 encoded, or a canonical UUID of the pod?

jonboulle:
@schancel @eyakubovich can one of you tackle this please?

jonboulle:
Ping @schancel

schancel:
Pong.

philips:
@schancel I think @jonboulle is asking if you could help out with a PR.

jonboulle:
yes, that's what I was asking :-)

schancel:
Yes, I was acknowledging that.  :)

schancel:
I think we closed this one up with my PR?

PR #451, opened 2015-07-03
spec/ace.md: Add specificity to metadata endpoint descriptions.

schancel:
Addresses https://github.com/appc/spec/issues/373

### DIFF
--- a/ace/validator.go
+++ b/ace/validator.go
@@ -226,7 +226,7 @@ func ValidateMountpoints(wmp map[string]types.MountPoint) results {
 	return r
 }
 
-func metadataRequest(req *http.Request) ([]byte, error) {
+func metadataRequest(req *http.Request, expectedContentType string) ([]byte, error) {
 	cli := http.Client{
 		Timeout: 100 * time.Millisecond,
 	}
@@ -242,6 +242,11 @@ func metadataRequest(req *http.Request) ([]byte, error) {
 		return nil, fmt.Errorf("Get %s failed with %v", req.URL, resp.StatusCode)
 	}
 
+	if respContentType := resp.Header.Get("Content-Type"); respContentType != expectedContentType {
+		return nil, fmt.Errorf("`%s` did not respond with expected Content-Type header.  Expected %s, received %s",
+			req.RequestURI, expectedContentType, respContentType)
+	}
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return nil, fmt.Errorf("Get %s failed on body read: %v", req.URL, err)
@@ -250,27 +255,28 @@ func metadataRequest(req *http.Request) ([]byte, error) {
 	return body, nil
 }
 
-func metadataGet(metadataURL, path string) ([]byte, error) {
+func metadataGet(metadataURL, path string, expectedContentType string) ([]byte, error) {
 	uri := metadataURL + metadataPathBase + path
 	req, err := http.NewRequest("GET", uri, nil)
 	if err != nil {
 		panic(err)
 	}
-	return metadataRequest(req)
+
+	return metadataRequest(req, expectedContentType)
 }
 
-func metadataPost(metadataURL, path string, body []byte) ([]byte, error) {
+func metadataPost(metadataURL, path string, body []byte, expectedContentType string) ([]byte, error) {
 	uri := metadataURL + metadataPathBase + path
 	req, err := http.NewRequest("POST", uri, bytes.NewBuffer(body))
 	if err != nil {
 		panic(err)
 	}
-	req.Header.Set("Content-Type", "text/plan")
+	req.Header.Set("Content-Type", "text/plain")
 
-	return metadataRequest(req)
+	return metadataRequest(req, expectedContentType)
 }
 
-func metadataPostForm(metadataURL, path string, data url.Values) ([]byte, error) {
+func metadataPostForm(metadataURL, path string, data url.Values, expectedContentType string) ([]byte, error) {
 	uri := metadataURL + metadataPathBase + path
 	req, err := http.NewRequest("POST", uri, strings.NewReader(data.Encode()))
 	if err != nil {
@@ -278,7 +284,7 @@ func metadataPostForm(metadataURL, path string, data url.Values) ([]byte, error)
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
-	return metadataRequest(req)
+	return metadataRequest(req, expectedContentType)
 }
 
 func validatePodAnnotations(metadataURL string, pm *schema.PodManifest) results {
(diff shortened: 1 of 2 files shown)


spec/ace.md
```diff
@@ -290,5 +291,5 @@ Accessible at `$AC_METADATA_URL/acMetadata/v1/pod/hmac`
 
 | Entry | Description |
 |-------|-------------|
-|sign   | POST a form with content=&lt;object to sign&gt; and retrieve a base64 hmac-sha512 signature as the response body. The metadata service holds onto the secret key as a sort of pod TPM. |
+|sign   | Client APPS must POST a form with content=&lt;object to sign&gt;.  The response must specify a `Content-Type` header of `text/plain; charset=us-ascii` and the body must be a base64 encoded hmac-sha512 signature based on an HMAC key maintained by the Metadata Service. |
```
jonboulle: not sure why APPS is capitalised

spec/ace.md
```diff
@@ -247,12 +247,13 @@ The App Container specification defines an HTTP-based metadata service for provi
 
 ### Metadata Service
 
-The ACE must provide a Metadata service on the address given to the applications via the `AC_METADATA_URL` [environment variable](#execution-environment).
+The ACE must provide a Metadata service on the address given to the applications via the `AC_METADATA_URL` [environment variable](#execution-environment).  This URL must reference an endpoint which conforms to the HTTP protocol with TLS enabled except in cases where the Metadata Service resides on the same physical host.
```
jonboulle: mind sticking to sentence-per-line for now?

spec/ace.md
schancel: Because I was half asleep.  Will fix.

spec/ace.md
schancel: Ah right, sorry

spec/ace.md
```diff
@@ -247,12 +247,13 @@ The App Container specification defines an HTTP-based metadata service for provi
...
 
 ACE implementations SHOULD embed an authorization token in `AC_METADATA_URL`, which provides a means for the metadata service to uniquely and securely identify a pod.
 For example, `AC_METADATA_URL` passed to a pod could be set to `https://10.0.0.1:8888/Y4vFeVZzKM2T9rwkpWHfqXuGsNjS6O5c` with the path portion acting as a token.
 Since the token is used by the Metadata Service to authenticate the pod's identity, it SHOULD have no fewer than 128 bits of entropy (i.e. size of UUID), and SHOULD NOT be easily guessable (e.g. the pod UUID should not be used).
 
+For the following endpoints, the media type of `application/json` must be specified and the body must conform to [RFC4627](http://www.ietf.org/rfc/rfc4627.txt) unless otherwise specified.
```
jonboulle: Perhaps this should say "unless otherwise specified" since below there are some countercases (text/plain)

jonboulle:
Thanks, looks good overall. Would you mind updating the validator too? (if it needs it)

jonboulle:
LGTM. @eyakubovich can you review?

spec/ace.md
```diff
@@ -248,11 +248,13 @@ The App Container specification defines an HTTP-based metadata service for provi
 ### Metadata Service
 
 The ACE must provide a Metadata service on the address given to the applications via the `AC_METADATA_URL` [environment variable](#execution-environment).
+This URL must reference an endpoint which conforms to the HTTP protocol with TLS enabled except in cases where the Metadata Service resides on the same physical host.
```
eyakubovich: I think it's worth writing this spec from the point of the application and the environment it needs to support and expect. In that sense, we should not mandate that a runtime must use TLS; only that the `AC_METADATA_URL` may use `http` or `https` scheme and that both must be supported. If the runtime has setup an `stunnel` or uses IPSec, it may not want to use TLS at the application level.

spec/ace.md
```diff
@@ -264,9 +266,9 @@ Retrievable at `$AC_METADATA_URL/acMetadata/v1/pod`
 
 | Entry       | Description |
 |-------------|-------------|
-|annotations/ | Top level annotations from Pod Manifest. |
+|annotations/ | Top level annotations from Pod Manifest. Response body should conform to the sub-schema of the annotations property from the Pod specification (e.g. ```[ { "name": "ip-address", "value": "10.1.2.3" } ]```). |
```
eyakubovich: That's not what happens now (at least as implemented by `rkt`). It returns `text/plain` with just the annotation names, separated by newlines. That is inline with AWS metadata service -- the inspiration for this metadata service. However if want to move all responses to `application/json`, I would be OK with it.

schancel: I think we should.   The apps need to support parsing JSON for the other endpoints -- I think it would be easier from the perspective of the application to make it consistent.

schancel:
I suspect the validator will need changes.  I'll look into that today and add some commits on here.

ace/validator.go
```diff
@@ -359,7 +371,7 @@ func validateAppAnnotations(metadataURL string, pm *schema.PodManifest, app *sch
 			continue
 		}
 
-		val, err := metadataGet(metadataURL, "/apps/"+string(app.Name)+"/annotations/"+key)
+		val, err := metadataGet(metadataURL, "/apps/"+string(app.Name)+"/annotations/"+key, "text/plain; charset=us-ascii")
```
schancel: Not sure how we should represent this in the spec.  I didn't notice that this was possible from reading the spec originally. @eyakubovich

ace/validator.go
```diff
@@ -278,30 +284,36 @@ func metadataPostForm(metadataURL, path string, data url.Values) ([]byte, error)
...
 }
 
 func validatePodAnnotations(metadataURL string, pm *schema.PodManifest) results {
 	r := results{}
 
 	var actualAnnots types.Annotations
+	var fetchedAnnots types.Annotations
 
-	annots, err := metadataGet(metadataURL, "/pod/annotations/")
+	annots, err := metadataGet(metadataURL, "/pod/annotations/", "application/json")
```
schancel: See next comment as well.

ace/validator.go
jonboulle: ping @eyakubovich ?

jonboulle: @philips ?

eyakubovich: @philips I think you wanted to keep parity with aws/gce style metadata services. But if you'd like, we can switch to all JSON. I think in containers people will be doing less bash scripting compared to VM so JSON is probably more appropriate.

philips: json content type only is OK

philips:
lgtm and +1 on moving to JSON only. Sorry for the delay on the review.

schancel:
Okay so should I just remove the reference to GET endpoints for specific keys?

jonboulle:
Yes, I think that makes sense

schancel:
Okay.  Sorry for the huge delay.  Does this still look reasonable?  I removed the validation of the original gets for the annotation keys, and I don't see where the spec referenced that was possible, so no update there.

jonboulle:
LGTM

PR #452, opened 2015-07-05
Default NO_SIGNATURE to empty string (fix #446)

### LINKED ISSUE #446: build_aci script fails if NO_SIGNATURE is defined
Because of strict mode, `if [ -z "$NO_SIGNATURE" ] ;` fails with `unbound variable` instead of executing the branch.

mpasternacki:


jonboulle:
thank you!

PR #454, opened 2015-07-17
schema/types: export IsValidOSArch function

iaguis:
This function can be reused by rkt to check images against its own valid
OS and architectures.

Useful for coreos/rkt#271


jonboulle:
LGTM

PR #460, opened 2015-07-31
ace: generate an armored gpg signature

iaguis:
We were generating an .asc signature file but we didn't specify --armor
to the gpg commad so it was a binary file instead of an ASCII armored
file.


vbatts:
LGTM

PR #461, opened 2015-08-02
aci: make NewCompressedReader an io.ReadCloser

sgotti:
now NewCompressedReader returns an io.Reader but gzip.NewReader returns an
io.ReadCloser that was leaked.

This patch makes NewCompressedReader return an io.ReadCloser and wraps others
implementing only io.Reader making them an io.ReadCloser with a fake Close
method.

Consequently NewCompressedTarReader must implement a Close method needed to
close the underlying CompressedReader.

In the meantime duplicated code from actool/validate.go is removed in favor of
the one provided by the aci package.

In another patch also XzReader will be changed to become an io.ReadCloser.

As this is an api change, at least rkt and docker2aci needs to be patched when
updating vendored appc/spec in rkt.

### DIFF
--- a/aci/file.go
+++ b/aci/file.go
@@ -132,6 +132,7 @@ func ManifestFromImage(rs io.ReadSeeker) (*schema.ImageManifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer tr.Close()
 
 	for {
 		hdr, err := tr.Next()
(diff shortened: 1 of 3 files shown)


vbatts:
There is a stdlib `ioutil.NopCloser` that you can use instead of making this fake closer.

sgotti:
@vbatts someone stole my idea! :stuck_out_tongue:  Thanks for the suggestion!

philips:
@jonboulle please review

sgotti:
Updated after @jonboulle review.

Issue #455, opened 2015-07-19
New XZ package in pure Go


philips:
We should use this instead! https://github.com/xi2/xz

xi2:
Hi. I'm the author of the xz package. Thank you for your interest.

I just wanted to mention that since my posting to golang-nuts I have added the small number of missing features required for it to handle any xz file, no matter how weird or wonderful. One bug has been found and fixed and I've now run several hundred thousand compressed files through it both with and without the x86 BCJ filter.

If you choose to evaluate it, any feedback (good or bad) is welcome.

vbatts:
ohman. very exciting.

Is it consistent to itself? (e.g. compressing the same payload will produce the same cryptographic checksum of the compressed content)

xi2:
@jonboulle, the package is only a decompressor unfortunately. I believe #14 is seeking a compressor?

vbatts:
ah, I see. I look forward to compress and decompress :-)

PR #462, opened 2015-08-02
aci: make XzReader implement io.ReadCloser

sgotti:
this depends on #461. The relevant patch is `aci: make XzReader implement io.ReadCloser`

this is needed as the XzReader, now calling an external process, needs to be
closed or the underlying process will continue to read from the passed file
descriptor, creating a race with subsequent methods calling another XzReader.

The Close method closes the output pipe, kills the process and then waits for
it to exit.

A possible fix after/substituting this will be to migrate to a pure go xz decompressor like the one proposed in #455

This is a part of the fix for coreos/rkt#1224 (missing the rkt part using the right functions)

### DIFF
--- a/aci/file.go
+++ b/aci/file.go
@@ -103,25 +103,43 @@ func DetectFileType(r io.Reader) (FileType, error) {
 	}
 }
 
-// XzReader shells out to a command line xz executable (if
+// XzReader is an io.ReadCloser which decompresses xz compressed data.
+type XzReader struct {
+	io.ReadCloser
+	cmd     *exec.Cmd
+	closech chan error
+}
+
+// NewXzReader shells out to a command line xz executable (if
 // available) to decompress the given io.Reader using the xz
-// compression format
-func XzReader(r io.Reader) io.ReadCloser {
+// compression format and returns an *XzReader.
+// It is the caller's responsibility to call Close on the XzReader when done.
+func NewXzReader(r io.Reader) (*XzReader, error) {
 	rpipe, wpipe := io.Pipe()
 	ex, err := exec.LookPath("xz")
 	if err != nil {
 		log.Fatalf("couldn't find xz executable: %v", err)
 	}
 	cmd := exec.Command(ex, "--decompress", "--stdout")
+
+	closech := make(chan error)
+
 	cmd.Stdin = r
 	cmd.Stdout = wpipe
 
 	go func() {
 		err := cmd.Run()
 		wpipe.CloseWithError(err)
+		closech <- err
 	}()
 
-	return rpipe
+	return &XzReader{rpipe, cmd, closech}, nil
+}
+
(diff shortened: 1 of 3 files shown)


philips:
@jonboulle can you give this a review?

philips:
lgtm but blocked on @jonboulle

aci/file.go
```diff
@@ -103,25 +103,41 @@ func DetectFileType(r io.Reader) (FileType, error) {
 	}
 }
 
+type XzReader struct {
+	io.ReadCloser
+	cmd     *exec.Cmd
+	closech chan error
+}
+
 // XzReader shells out to a command line xz executable (if
```
jonboulle: needs to be moved to the type

jonboulle: er, adjusted

actool/validate.go
```diff
@@ -176,34 +171,3 @@ func detectValType(file *os.File) (string, error) {
 		return "", nil
 	}
 }
-
-func maybeDecompress(rs io.ReadSeeker) (io.Reader, error) {
-	// TODO(jonboulle): this is a bit redundant with detectValType
```
jonboulle: thanks!

actool/manifest.go
```diff
@@ -422,7 +423,7 @@ func runPatchManifest(args []string) (exit int) {
 		}
 	}
 
-	err = extractManifest(tr, tw, false, newManifest)
+	err = extractManifest(tr.Reader, tw, false, newManifest)
```
jonboulle: out of interest is this actually necessary? I thought it would've done the right thing if you passed `tr`

aci/file.go
```diff
@@ -155,20 +172,31 @@ func ManifestFromImage(rs io.ReadSeeker) (*schema.ImageManifest, error) {
 	}
 }
 
+type CompressedTarReader struct {
+	*tar.Reader
+	cr io.ReadCloser
+}
+
 // NewCompressedTarReader creates a new tar.Reader reading from the given ACI image.
```
jonboulle: s/tar.Reader/CompressedTarReader/

aci/file.go
```diff
@@ -155,20 +172,31 @@ func ManifestFromImage(rs io.ReadSeeker) (*schema.ImageManifest, error) {
 	}
 }
 
+type CompressedTarReader struct {
```
jonboulle: CompressedTarReaderblabla

jonboulle:
A couple of tiny doc things but LGTM overall, thanks sgotti!

actool/manifest.go
sgotti: Unfortunately when a functions wants a *tar.Reader, passing the embedding struct doesn't work. I haven't found a better way to avoid this...

sgotti:
Updated after @jonboulle review on #461 and this one.

Issue #464, opened 2015-08-03
ace: validator fails to detect mount points


iaguis:
Currently, the ACE validator checks if `/db` is a mount point since [it's defined as a mount point in the manifest](https://github.com/appc/spec/blob/0d1edc37553152577a56f256279992200f142b48/ace/image_manifest_main.json#L37).

It checks it's a mount point by calling `statfs()` on the directory and the parent and then comparing the f_fsid field. This check doesn't detect if a directory is a bind mount so it will error out mistakenly when the mount point is in the same filesystem as the parent:

```
# mkdir a b
# mount --bind a b
# stat --file-system b
File: "b"
    ID: df4743518d77c185 Namelen: 255     Type: ext2/ext3
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 116580597  Free: 32241161   Available: 32237065
Inodes: Total: 29622272   Free: 28560905
# stat --file-system b/..
File: "b/.."
    ID: df4743518d77c185 Namelen: 255     Type: ext2/ext3
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 116580597  Free: 32241161   Available: 32237065
Inodes: Total: 29622272   Free: 28560905
```

We could use [`name_to_handle_at()`](http://man7.org/linux/man-pages/man2/open_by_handle_at.2.html) or the [`mnt_id` flag in `/proc/[pid]/fdinfo/[fd]`](http://man7.org/linux/man-pages/man5/proc.5.html) but the former is not included in all kernels and the latter was added in Linux 3.15.

mpasternacki:
I bumped into a similar issue on FreeBSD: the `Fsid` fields were zeroed. I'm not 100% sure whether it's because of bind mount, or because jail is not allowed to see the fields. I've got a monkey-patch in Jetpack for that (https://github.com/3ofcoins/jetpack/commit/b4b599e697572d851d82163be1117d20336d971b) and I want to make a proper pull request. Not sure whether this approach would work on Linux, though: it seems to me that the `Statfs_t` fields I'm using are not available on Linux, but this may provide a direction to work from.

iaguis:
It seems we can't use [`name_to_handle_at()`](http://man7.org/linux/man-pages/man2/open_by_handle_at.2.html) because it requires `CAP_DAC_READ_SEARCH`.

[...]

PR #466, opened 2015-08-03
ace/validator: deep compare Statfs_t on FreeBSD

mpasternacki:
Inside a jail, `Fsid` on bind mounts is `{0,0}` and fails the test.

### DIFF
--- a/ace/os_default.go
+++ b/ace/os_default.go
@@ -1,0 +1,9 @@
+// +build !freebsd
+
+package main
+
+import "syscall"
+
+func isSameFilesystem(a, b *syscall.Statfs_t) bool {
+	return a.Fsid == b.Fsid
+}
--- a/ace/os_freebsd.go
+++ b/ace/os_freebsd.go
@@ -1,0 +1,16 @@
+// +build freebsd
+
+package main
+
+import "syscall"
+
+func isSameFilesystem(a, b *syscall.Statfs_t) bool {
+	if a.Fsid != (syscall.Fsid{}) || b.Fsid != (syscall.Fsid{}) {
+		// If Fsid is not empty, we can just compare the IDs
+		return a.Fsid == b.Fsid
+	}
+	// Fsids are zero, this happens in jails, but we can compare the rest
+	return a.Fstypename == b.Fstypename &&
+		a.Mntfromname == b.Mntfromname &&
+		a.Mntonname == b.Mntonname
+}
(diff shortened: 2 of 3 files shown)


mpasternacki:
Related: #464

iaguis:
Following up from https://github.com/appc/spec/issues/464#issuecomment-127233744

I'd like the function to be called `func isSameMountpoint(pathA, pathB string) (bool, error)` and it will do the following.

* On Linux it will try to use the `mnt_id` solution and if it's not supported it will error out
* On FreeBSD it will do `fstat()` and, if `Fsid` is not empty, it will compare them and if it is, it will compare `Fstypename`, `Mntfromname` and `Mntonname`. This function would never return an error.

If we detect an error on the test, we just skip it.

mpasternacki:
Sounds good to me. Should the Linux version just `ioutil.ReadFile("/proc/self/fdinfo/X")` and parse it as text, or is there any library call available for that? Could you draft such a function? I'll adapt the PR to use the API you proposed.

I don't think we need to particularly worry about older kernels. The `/proc/$PID/fdinfo/` has been introduced in Linux 2.6.22 in 2007, while cgroups (according to Wikipedia) landed in 2.6.24. Worrying about kernels older than 8 years and not supporting features needed to implement containers (at least not without crazy patches) seems to be excessively conservative to me.

I'd leave three versions, then: Linux (as you described), FreeBSD (with deep compare on `Statfs_t`), and default version using current code as a fallback.

iaguis:
> The /proc/$PID/fdinfo/ has been introduced in Linux 2.6.22 in 2007

The directory is there since 2007 but [the `mnt_id` field was introduced in Linux 3.15](http://man7.org/linux/man-pages/man5/proc.5.html):

> The mnt_id field, present since Linux 3.15 [...]

iaguis:
> Could you draft such a function?

I'm working on it :)

mpasternacki:
Maybe we could use an older `/proc/PID/mountinfo` interface, then?

iaguis:
> Maybe we could use an older `/proc/PID/mountinfo` interface, then?

That would be quite difficult just for a validation test:

* Parsing that file is more difficult, you'd need to find under which mountpoint falls the file by going through the parents until you find something that's in `mountinfo`
* Symbolic links make things even more complicated

mpasternacki:
We could make this check higher-level, the question we need answered is "is `/db` a mount point?", and `mountinfo` answers this question easily. Comparing `Statfs` is just an implementation detail, we don't *need* to implement checking whether two paths are on the same FS.

iaguis:
> We could make this check higher-level, the question we need answered is "is /db a mount point?", and mountinfo answers this question easily. Comparing Statfs is just an implementation detail, we don't need to implement checking whether two paths are on the same FS.

Hmmm... you're right. I got lost in the low-level details.

I guess that could work then :)

iaguis:
Filed another PR: https://github.com/appc/spec/pull/467

mpasternacki:
Closing in favour of #467

PR #467, opened 2015-08-03
ace/validator: fix mountpoint checks

### LINKED ISSUE #464: ace: validator fails to detect mount points

iaguis:
This includes:

- A proper check for Linux. The previous one was just checking if the
  devices were the same, this is not true for bind-mounts within the
  same device

- A check for FreeBSD courtesy of Maciej Pasternacki
  <maciej@3ofcoins.net>. Jails seem to have f_fsid set to zero so the
  original check didn't work. Now we check r_fstypename, f_mntfromname
  and f_mntonname too.

- As a fallback (for other systems) we do the same as the original code:
  just check the f_fsid.

Fixes #464 
cc @mpasternacki

### DIFF
--- a/ace/os_linux.go
+++ b/ace/os_linux.go
@@ -1,0 +1,31 @@
+// +build linux
+
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+func checkMountImpl(d string, readonly bool) error {
+	mountinfoPath := fmt.Sprintf("/proc/self/mountinfo")
+	mi, err := os.Open(mountinfoPath)
+	if err != nil {
+		return err
+	}
+	defer mi.Close()
+
+	isMounted, ro, err := parseMountinfo(mi, d)
+	if err != nil {
+		return err
+	}
+	if !isMounted {
+		return fmt.Errorf("%q is not a mount point", d)
+	}
+
+	if ro == readonly {
+		return nil
+	} else {
+		return fmt.Errorf("%q mounted ro=%t, want %t", d, ro, readonly)
+	}
+}
--- a/ace/os_default.go
+++ b/ace/os_default.go
@@ -1,0 +1,12 @@
+// +build !linux
+// +build !freebsd
+
+package main
+
+import (
+	"syscall"
+)
+
+func isSameFilesystem(a, b *syscall.Statfs_t) bool {
+	return a.Fsid == b.Fsid
+}
--- a/ace/os_freebsd.go
+++ b/ace/os_freebsd.go
@@ -1,0 +1,16 @@
+// +build freebsd
+
+package main
+
+import "syscall"
+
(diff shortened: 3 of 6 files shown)


mpasternacki:
After adding the  missing `import "fmt"` in `os_shared.go`, works perfectly for me. :+1:

iaguis:
Ooops, thanks for catching that!

ace/os_linux.go
```diff
@@ -0,0 +1,63 @@
...
+	sc := bufio.NewScanner(mi)
+	for sc.Scan() {
+		var (
+			mountID        int
+			parentID       int
+			majorMinor     string
+			root           string
+			mountPoint     string
+			mountOptions   string
+			optionalFields string
```
alban: according to `proc(5)`:
> (7)  optional fields: zero or more fields of the form "tag[:value]".

Does the `Scanf` work if there are several fields? It would be good to add some unit tests about that.

alban:
Except the issue with parsing `optionalFields`, it looks good.

ace/os_linux.go
iaguis: It's probably fine because we only care about (5) but for that same reason we can only parse the first 5 fields (which are fixed) and avoid possible problems. I'll update the PR with that and some unit tests. Thanks!

iaguis:
Updated with only parsing the first (6) fields and added a test

jonboulle:
:+1: thanks

PR #469, opened 2015-08-04
spec: move labels matching part to discovery.

sgotti:
Now aci.md "Dependency Matching" chapter explains how dependency's name and
labels are used for discovery and then compared to the name and labels in the
Image Manifest of the retrieved image.

This labels matching should be done for every discovery.

This patch moves the labels matching part to discovery, and keeps in "Dependency
Matching" the additional ImageID check.

From the logical point of view of "Dependency Matching" this means that now labels
matching is done before ImageID matching since it's included in the discovery
logic.


jonboulle:
LGTM.. @vbatts?

philips:
lgtm

PR #475, opened 2015-08-09
spec: pod: mounts point to path instead of mountPoint

philips:
In order to support the use case of mounting at an arbitrary location
inside of the app rootfs we remove the "merge table" between app
mountPoint and volumes. Before, if someone wanted to simply mount over
say /var/log with a volume and that wasn't offered as a mountPoint by
the app the entire app section would need to be overridden.

This does not remove the mountPoint name or volume name so if a runtime
wishes to have this volume/mountPoint map it may still be implemented in
the runtime UX.

This concern was brought up by @thockin in #364.

### DIFF
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -126,19 +126,19 @@ func (al AppList) Get(name types.ACName) *RuntimeApp {
 	return nil
 }
 
-// Mount describes the mapping between a volume and an apps
-// MountPoint that will be fulfilled at runtime.
+// Mount describes the mapping between a volume and the path it is mounted
+// inside of an app's filesystem.
 type Mount struct {
-	Volume     types.ACName `json:"volume"`
-	MountPoint types.ACName `json:"mountPoint"`
+	Volume types.ACName `json:"volume"`
+	Path   string       `json:"path"`
 }
 
 func (r Mount) assertValid() error {
 	if r.Volume.Empty() {
 		return errors.New("volume must be set")
 	}
-	if r.MountPoint.Empty() {
-		return errors.New("mountPoint must be set")
+	if r.Path == "" {
+		return errors.New("path must be set")
 	}
 	return nil
 }
--- a/spec/pods.md
+++ b/spec/pods.md
@@ -59,7 +59,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
                 ]
             },
             "mounts": [
-                {"volume": "work", "mountPoint": "work"}
+                {"volume": "work", "path": "/mnt/foo"}
             ]
         },
         {
@@ -94,7 +94,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
                 ]
             },
             "mounts": [
-                {"volume": "work", "mountPoint": "backup"}
+                {"volume": "work", "path": "/mnt/bar"}
             ],
             "annotations": [
                 {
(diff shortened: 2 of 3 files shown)


philips:
An alternate solution to this was proposed by @ibukanov in https://github.com/appc/spec/issues/364#issuecomment-101536296. Any strong opinions either way?

spec/pods.md
```diff
@@ -159,7 +159,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
     * **app** (object, optional) substitute for the app object of the referred image's ImageManifest. See [Image Manifest Schema](aci.md#image-manifest-schema) for what the app object contains.
     * **mounts** (list of objects, optional) list of mounts mapping an app mountPoint to a volume. Each mount has the following set of key-value pairs:
       * **volume** (string, required) name of the volume that will fulfill this mount (restricted to the [AC Name](types.md#ac-name-type) formatting)
-      * **mountPoint** (string, required) name of the app mount point to place the volume on (restricted to the [AC Name](types.md#ac-name-type) formatting)
+      * **path** (string, required) path inside of the app filesystem to mount the volume; generally this will come from one of an apps mountPoint paths
```
jonboulle: > generally this will come from one of an apps mountPoint paths

I think it's unclear what exactly this would look like to a user or how an implementation would handle it. For the sake of me understanding, is this supposed to be a no-op with the current e.g. rkt UX? That the mountPoint names might be exposed to users and would be used when an executor reifies a podmanifest?

philips: Right, the basic observation is two fold:

1) We want to enable volumes to be mounted anywhere
2) We want to give applications the ability to have a descriptive name of the mount points so a user can guess/discover where to place a relevant volume

But, the spec doesn't need to enforce a mapping between volume and mount point as it did before.

spec/pods.md
```diff
@@ -159,7 +159,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
...
     * **annotations** (list of objects, optional) arbitrary metadata appended to the app. The annotation objects must have a *name* key that has a value that is restricted to the [AC Name](types.md#ac-name-type) formatting and *value* key that is an arbitrary string). Annotation names must be unique within the list. These will be merged with annotations provided by the image manifest when queried via the metadata service; values in this list take precedence over those in the image manifest.
 * **volumes** (list of objects, optional) list of volumes which will be mounted into each application's filesystem
     * **name** (string, required) used to map the volume to an app's mountPoint at runtime. (restricted to the [AC Name](types.md#ac-name-type) formatting)
```
jonboulle: This should be updated

philips:
@jonboulle updated that name line

schema/pod.go
```diff
@@ -129,16 +129,16 @@ func (al AppList) Get(name types.ACName) *RuntimeApp {
 // Mount describes the mapping between a volume and an apps
```
jonboulle: this too :|

philips: Fixed this

jonboulle:
tentative LGTM - I'd still like to see a bit of a more verbose explanation but am happy to follow up with that. However would be good to get some more eyes on this. @vbatts @krobertson ?

vbatts:
this is a breaking change, though the having the container specify where is expects a named mount makes more since.

vbatts:
only other piece would be, if there is a needed for a container to request a typed-mount (e.g. a `tmpfs` mount) at a particular location. Maybe that is just something to be done in the hooks...

philips:
@vbatts Yes, this is a breaking change, we have said we can break things pre-1.0.

I can see a use case for a container requesting that but that gets into how big can that tmpfs be, etc. Seems like a separate issue we need to tackle from this issue.

krobertson:
What is the correlation between the `mountPoints->path` in the image spec to the pod spec?  Typically look for the examples of the JSON to have consistency where the spec is describing consistency, so should `image.json` mention the same mount points and paths?

vbatts:
since it's ok to be a breaking change, and the `tmpfs` scenario would be handled elsewhere,
this LGTM

jonboulle:
Fixed via #495

PR #339, opened 2015-09-11
spec: support supplementary GIDs

jonboulle:
Image manifest allows me to spec what GID to run as, but does not allow me to add supplemental groups.  This is important for many reasons, not the least of which is disk quota management.


vbatts:
solid
LGTM

PR #485, opened 2015-09-11
ace validator: fix typo in the HTTP-POST options sent to metadata srv

alban:
When requesting the metadata service to verify a signature from another pod,
the uuid of the pod needs to be passed as an HTTP-POST option. According to
the spec, the name of the option is "uuid" and not "uid".


vbatts:
LGTM

PR #486, opened 2015-09-14
pkg/device: add darwin build flag

stufflebear:
this makes this package buildable on osx.


jonboulle:
I don't have access to OS X right now but I'll take your word for it - LGTM :-)

PR #495, opened 2015-09-18
spec: mounts using path instead of mountPoint

jonboulle:
Builds on #475, with some minor tweaks:
- clarifying the use of names as keys
- linking the example image and pod manifests (as suggested in 
  https://github.com/appc/spec/pull/475#issuecomment-132827287)


jonboulle:
Merging this based on the several LGTMs in #475

PR #506, opened 2015-09-28
actool/manifest.go: support patching app.SupplementaryGroups

vcaputo:
Adds --supplementary-groups flag to actool which takes a comma-separated
list of gids.

### DIFF
--- a/actool/manifest.go
+++ b/actool/manifest.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


vcaputo:
/cc @jonboulle 

In adding this I noticed the current patching implementation relying on non-empty strings as triggers prevents the user from specifying empty values for patches.  That should probably be fixed up at some point.

jonboulle:
LGTM, thanks

> In adding this I noticed the current patching implementation relying on non-empty strings as triggers prevents the user from specifying empty values for patches. That should probably be fixed up at some point.

yes, indeed

PR #508, opened 2015-09-29
last-ditch: Handle labels

krnowak:
This PR does some cleanups first and then adds labels to pod's runtime image and to image manifest.

This may be useful for rkt to extract the version of an image (and maybe arch and os if there is need) in error cases.

### DIFF
--- a/schema/lastditch/pod_test.go
+++ b/schema/lastditch/pod_test.go
@@ -17,14 +17,11 @@ package lastditch
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 )
 
 func TestInvalidPodManifest(t *testing.T) {
-	// empty image JSON
-	eImgJ := "{}"
-	// empty image instance
-	eImgI := rImgI("", "")
 	tests := []struct {
 		desc     string
 		json     string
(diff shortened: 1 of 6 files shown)


jonboulle:
Did you use a bad email address for these commits? it's not linking against your account

schema/lastditch/pod_test.go
```diff
@@ -32,47 +29,77 @@ func TestInvalidPodManifest(t *testing.T) {
...
+			desc:     "Check a pod manifest with an app based on an image containing labels with some extra fields",
+			json:     podJ(appsJ(appJ("a", rImgJ("i", "id", labsJ(labJ("n", "v", extJ("color"))), extJ("stuff")), extJ("trolls"))), extJ("goblins")),
+			expected: podI(appsI(appI("a", rImgI("i", "id", labsI(labI("n", "v")))))),
 		},
 	}
 	for _, tt := range tests {
 		got := PodManifest{}
 		if err := got.UnmarshalJSON([]byte(tt.json)); err != nil {
-			t.Errorf("%s: unexpected error during unmarshalling pod manifest: %v", tt.desc, err)
+			t.Errorf("%s: an unexpected error during unmarshalling a pod manifest: %v", tt.desc, err)
```
jonboulle: While not technically incorrect, in terms of style I find articles superfluous in these kinds of messages. The error is perfectly parseable and unambiguous without them.

krnowak: I blame @iaguis - he is always pointing out all the articles I forgot to write. :P

Will remove them.

krnowak:
> Did you use a bad email address for these commits? it's not linking against your account

Argh, yes. Will send a new batch with fixed email.

schema/lastditch/pod_test.go
iaguis: but not here :P

jonboulle:
I am a bit concerned that lastditch is basically expanding to just become the real schema without validation. Is there any way we can improve this situation to remove that duplication?

jonboulle:
This is fine for now, I'm going to file a follow up to address my concerns.

krnowak:
Sorry, I forgot to reply to this.

I added labels only because you could get an image version from them, which is a kind of basic information for error reporting and that's all. I don't plan adding new features there.

But yeah, I shortly had similar thoughts, especially when thinking about `--force` flag to `rkt rm` or `rkt image rm` to remove faulty pod/image from store - but currently that can be implemented by first trying to remove pod/image with normal PodManifest/ImageManifest and if that fails - with a `lastditch.PodManifest`/`lastditch.ImageManifest`. But that leads to a code duplication. I suppose that `PodManifest`/`ImageManifest` should be designed from the start to be a struct holding ultimately only strings, and `ACIdentifier`, `ACName` and others should have been an implementation detail. OTOH, that would complicate marshalling and unmarshalling. Anyway, probably that ship had sailed.

jonboulle:
> I suppose that PodManifest/ImageManifest should be designed from the start to be a struct holding ultimately only strings, and ACIdentifier, ACName and others should have been an implementation detail. 

I don't really know what you mean by this, could you provide a small example? 
I'm definitely open to any suggestions to clean up the models and validation code.

krnowak:
All I'm saying is that PodManifest in schema package should look like PodManifest in lastditch package.

With more details - I don't think that PodManifest/ImageManifest (or any type that those two manifests hold instances of) should have fields of restricting type like ACKind, ACIdenfifier, ACName or SemVer - these should be just plain strings. For example a Label struct could be simply:
```go
type Label struct {
	Name  string // string instead of ACIdentifier
	Value string
}
```

The validation could happen during serialization/deserialization, but with one catch - even if some string was invalid (say Name field in Label was not a valid ACIdentifier) we still would try to return a serialized/deserialized manifest. Something like:

```go
im := schema.blankImageManifest()
err := im.Deserialize(getSomeBytes())
if !schema.IsInvalidErr(err) {
	return err // badly structured JSON
}
if schema.IsInvalidErr(err) && !force {
	return err
}
stuff, err := getStuffWeNeed(im); err != nil {
	return err
}
```

jonboulle:
The nice part about using stronger types for the fields is that then they're self describing, self validating, and also the type system provides some amount of validation for free when working with the structs. But I am starting to think the downsides outweigh the niceties and something like your approach would be an improvement, despite the awkward code I think it will necessitate.

PR #509, opened 2015-09-29
aci build: add a callback to BuildWalker

### LINKED ISSUE #498: build aci: setting file ownership in the archive
When building an aci with [BuildWalker](https://github.com/appc/spec/blob/master/aci/build.go#L89), it adds files in the archive with the same ownership as they are on the filesystem.

It would be useful to be able to change the ownership of files in the aci, without needing to create the aci as root. It is needed for building the stage1 of rkt, see discussion on https://github.com/coreos/rkt/issues/1453#issuecomment-142314125.

At the moment, there is no parameters or callbacks to be able to set `hdr.Uid, hdr.Gid = 0, 0`.

/cc @jellonek

iaguis:
This callback allows users to modify the final tar headers or filter out
tar entries when building an ACI.

Also, add a --owner-root flag to acbuild so users can adjust the uids
and gids of all the files and directories in stage1 to 0:0. This is
needed by rkt: https://github.com/coreos/rkt/issues/1453

Fixes #498

### DIFF
--- a/aci/build.go
+++ b/aci/build.go
@@ -23,10 +23,18 @@ import (
 	"github.com/appc/spec/pkg/tarheader"
 )
 
+// TarHeaderWalkFunc is the type of the function which allows setting tar
+// headers or filtering out tar entries when building an ACI. It will be
+// applied to every entry in the tar file.
+//
+// If true is returned, the entry will be included in the final ACI; if false,
+// the entry will not be included.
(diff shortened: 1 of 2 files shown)


alban:
LGTM but I wrote a part of the patch, so that does not count much.

/cc @jonboulle @jellonek

jonboulle:
LGTM, thank you.

Issue #465, opened 2015-08-03
schema/types: volume doesn't allow source paths with special URL characters


iaguis:
We parse volume parameters with [`url.ParseQuery`](https://github.com/appc/spec/blob/0d1edc37553152577a56f256279992200f142b48/schema/types/volume.go#L101). This means that if the `source` field includes special URL characters they would be interpreted (e.g. `+` is converted to a space) making impossible to use those directories.

jonboulle:
Good catch, do you want to take a look at this? I suppose we need our own simpler impl of ParseQuery?

PR #514, opened 2015-10-01
discovery: URL-encode label values

krnowak:
Appc label values can be arbitrary strings, so they might need to be
escaped to become a valid URL label.

The problem arose when the "version" label has a value like
"0.8.1+git734734", where the "+" is a reserved char as noted in RFC
3986, section 2.2.


jonboulle:
related: #465

jonboulle:
@krnowak can you add a test please?

krnowak:
Added the tests, fixed some more bugs and added yet more tests.

jonboulle:
@sgotti what do you think of this?

jonboulle:
LGTM

PR #520, opened 2015-10-06
schema/types: escape values when parsing

### LINKED ISSUE #465: schema/types: volume doesn't allow source paths with special URL characters

iaguis:
To allow values with URL special characters like "+" or "&".

Fixes #465 

cc @krnowak

### DIFF
--- a/discovery/parse.go
+++ b/discovery/parse.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/appc/spec/schema/common"
 	"github.com/appc/spec/schema/types"
 )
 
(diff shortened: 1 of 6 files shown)


krnowak:
LFAD.

iaguis:
Ping?

jonboulle:
Sorry, this fell through the cracks. 
@iaguis please add a test table, then lgtm...

iaguis:
Added a test and fixed the examples in the function documentation to match the actual behavior :sweat_smile:

jonboulle:
:+1: thank you sir

PR #521, opened 2015-10-06
spec: relax the requirement of metadata service

eyakubovich:
Metadata service requires running a daemon which might
be burdensome for some implementations and use cases.


jonboulle:
LGTM, we intended to make this change a long time ago but I suppose it slipped through the cracks.

PR #523, opened 2015-10-08
schema: fix field names in errors

### LINKED ISSUE #493: error messages when asserting valid fields should reference actual json field names
When an app manifest doesn't have a user field the error message complains with 

    User is required

But the name of the json field isn't `User`, it's `user`.  That makes this error message confusing.  There are a few of this checks that are misleading like this.  See https://github.com/appc/spec/blob/785b1ee5/schema/types/app.go#L69-L76

jonboulle:
Fixes #493

### DIFF
--- a/schema/types/app.go
+++ b/schema/types/app.go
@@ -67,13 +67,13 @@ func (a *App) assertValid() error {
 		return err
 	}
 	if a.User == "" {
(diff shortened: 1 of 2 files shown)


iaguis:
LGTM

Issue #144, opened 2015-01-24
SPEC: volumes as single files


philips:
An example would be providing a volume that only has resolv.conf. The use case would be a volume for an automatically generated `resolv.conf`file. Then you could simply add a mountPoint stanza that says: `{source: magic-resolv-conf, destination: /etc/resolv.conf}`.

You could imagine the volume would look something like:
```
name: magic-resolv-conf
kind: host
source: /run/magic/resolv.conf
readOnly: true
```

It would be an interesting solution to the problem. Thoughts?

/cc @eyakubovich

cdaylward:
There are usually a half-dozen or so of these that usually get pulled into a container, resolve.conf, timezone, locale, etc (common when using a single authoritative image in disparate datacenter environments). Being able to bring in a single file is very useful.

Half-baked: It might also be worth considering a mechanism for "copy from host". A container's rootfs gets a copy of the file at the end of rootfs set up but before container instantiation. This would allow fine-grain control over what host state bleeds into a container and mitigate needing to produce new ACIs where simple container recycling would be preferable.

philips:
I was thinking about this today; perhaps we define some well-known OS volumes in the OS-SPEC. For example appc.io/linux/etc/resolv.confor something would provide a mountpoint for an appropriate resolv.conf and appc.io/linux/etc/hosts would provide an appropriate etc hosts file?

jonboulle:
question posed elsewhere: If we make this a read-only bind mount, how would we support runtime changes (e.g. if someone needs to modify /etc/hosts and /etc/resolv.conf while the pod is running)?

krobertson:
Single file volume mounts are likely useful outside of providing things like resolv.conf. I agree with @philips about an OS-SPEC specifying some defaults. It can make more sense to provide sane defaults for things like DNS and being able to resolve "localhost"... container authors might not think of these things up front, but I've seen apps bomb if they can't resolve localhost or its hostname.

The container's own hostname raises a concern on `/etc/hosts` though. We normally give containers a hostname, many frameworks need to be able to resolve their hostname, so we force writing a `/etc/hosts` per container. A volume mount wouldn't work well in those cases.

thockin:
I think we should not be afraid to be a bit opinionated here.  I flagged this on my readthrough.  OS-SPEC says nothing about /etc/hostname, /etc/hosts, /etc/mtab, or /etc/resolv.conf.  These are pretty important parts of most Linux systems, and hard to do some very fundamental things without.

[...]

PR #524, opened 2015-10-08
os-spec: ACE SHOULD provide /etc/hosts on Linux

jonboulle:
Partially addresses concerns raised in #144


jonboulle:
For example, as implemented in rkt in coreos/rkt#1541

philips:
LGTM

PR #528, opened 2015-10-12
discovery: add per-host HTTP headers to discovery functions

iaguis:
The discovery library doesn't have support for passing auth credentials so it will fail in servers that have authentication enabled.

This PR modifies the discovery functions to accept a map with per-host headers. If the host in the discovery URL is specified in this parameter, the corresponding headers will be applied when doing the discovery.

As an alternative, we could change the parameter to be just a header and leave converting the app to a hostname and selecting the right header to the library caller.

This is not ready for review/merge, I just want to get some feedback on the general idea.

Related: https://github.com/coreos/rkt/issues/1516

### DIFF
--- a/discovery/http.go
+++ b/discovery/http.go
@@ -31,15 +31,15 @@ var (
 	// Client is the default http.Client used for discovery requests.
 	Client *http.Client
 
-	// httpGet is the internal object used by discovery to retrieve URLs; it is
+	// httpDo is the internal object used by discovery to retrieve URLs; it is
 	// defined here so it can be overridden for testing
-	httpGet httpGetter
+	httpDo httpDoer
 )
 
-// httpGetter is an interface used to wrap http.Client for real requests and
+// httpDoer is an interface used to wrap http.Client for real requests and
 // allow easy mocking in local tests.
-type httpGetter interface {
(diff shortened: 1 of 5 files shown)


jonboulle:
Seems pretty reasonable. @sgotti ?

iaguis:
Added a test

discovery/http.go
```diff
@@ -33,13 +33,13 @@ var (
 
 	// httpGet is the internal object used by discovery to retrieve URLs; it is
```
jonboulle: needs update

discovery/http.go
```diff
@@ -52,10 +52,10 @@ func init() {
 	Client = &http.Client{
 		Transport: t,
 	}
-	httpGet = Client
+	httpDo = Client
 }
 
-func httpsOrHTTP(name string, insecure bool) (urlStr string, body io.ReadCloser, err error) {
+func httpsOrHTTP(name string, perHostHeader map[string]http.Header, insecure bool) (urlStr string, body io.ReadCloser, err error) {
```
jonboulle: how about just s/perHostHeader/hostHeaders/

jonboulle:
LGTM after minor cleanup

iaguis:
Updated

PR #534, opened 2015-10-29
schema/types: add mode, uid and gid for empty volumes

### LINKED ISSUE #526: spec: clarify empty volume permissions
Capturing a comment by @alban (coreos/rkt#1523)

> With host volumes, we don't need to specify the rights of the mount point because it comes from the existing directory on the host. But with empty volumes, maybe the manifest should tell us how the volume should be created (which owner and permissions).

Perhaps something like:

```
....
"volumes": [
    {
        "kind": "empty",
        "name": "whatever",
        "mode": "0775",
        "user": "1000",
        "group": "1005"
    }
],
...
```

With some sane defaults.

iaguis:
This allows setting which UID and GID and permission mode will be
applied to an empty volume.

We only allow UIDs and GIDs because a user/group name wouldn't make
sense in the Pod context: each app in the Pod can have a different
/etc/passwd with different mappings.

Fixes #526

### DIFF
--- a/spec/pods.md
+++ b/spec/pods.md
@@ -174,6 +174,9 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
         * **empty** - creates an empty directory on the host and bind mounts it into the container. All containers in the pod share the mount, and the lifetime of the volume is equal to the lifetime of the pod (i.e. the directory on the host machine is removed when the pod's filesystem is garbage collected)
         * **host** - fulfills a mount point with a bind mount from a **source** directory on the host.
     * **source** (string, required if **kind** is "host") absolute path on host to be bind mounted under a mount point in each app's chroot.
+    * **mode** (integer, optional, only interpreted if **kind** is "empty", defaults to "0755" if unsupplied) indicates the mode permission of the empty volume.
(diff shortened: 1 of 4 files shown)


spec/pods.md
```diff
@@ -174,6 +174,9 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
         * **empty** - creates an empty directory on the host and bind mounts it into the container. All containers in the pod share the mount, and the lifetime of the volume is equal to the lifetime of the pod (i.e. the directory on the host machine is removed when the pod's filesystem is garbage collected)
         * **host** - fulfills a mount point with a bind mount from a **source** directory on the host.
     * **source** (string, required if **kind** is "host") absolute path on host to be bind mounted under a mount point in each app's chroot.
+    * **mode** (integer, optional, only intepreted if **kind** is "empty", defaults to "0755" if unsupplied) indicates the mode permission of the empty volume.
```
jonboulle: s/intepreted/interpreted/

jonboulle:
@krobertson @vbatts any thoughts here? Seems reasonable to me but another ack on the polymorphic volume struct would be good.

philips:
lgtm

PR #541, opened 2015-11-26
schema/types: rework empty volume type code

jonboulle:
tl;dr: this patch:
- fixes a panic in the Volume type's String() method
- ensures that the serialization of a Volume to/from a string is
  reversible - i.e., `s == VolumeFromString(s).String()`

Full details:

The `Volume` type is polymorphic; it can be of several different Kinds,
and for each Kind only a select few fields are valid. When new fields
(`mode`, `uid`, `gid`) were added to the `empty` volume Kind (via
26af7cefe402259b348fdfa2299d8d2e0a595e2e), they were simply added as new
fields. However, this means that Volume structs of Kinds for which those
fields are not valid (e.g. the `host` Kind) would necessarily have
values set in those fields - since in Go any fields will be
automatically set to their zero value. To mitigate this, the previous
code set special sentinel values to those fields (e.g. -1 to the
`uid`/`mode` fields) to indicate that they were "unset"; but this is
slightly awkward since it necessitates manually setting those sentinel
values at certain points.

This patch instead change those fields to be pointers, so that it is
trivially possible to determine whether or not those fields are really
set, and disambiguate from the zeroed value case. This also means that a
"zeroed" Volume type is really unset and does not need modification to
set the sentinels appropriately.

Another change is in how exactly a Volume type is serialized to a
string. `String()` previously attempted to dereference the `readOnly`
field even if it was not necessarily set. Even though such a nil
dereference would ordinarily cause the Go runtime to panic, the
`fmt.Printf` function (and friends) uses a `recover()` to actually
_catch_ any panic occurred, so instead this results in a weird panic
trace _within the returned string_, but business otherwise continues as
usual (smdh).

As well as fixing this panic (by checking the pointer before
dereferencing), we now also construct the serialized Volume string
piece-by-piece, only if fields are set; for example, if an `empty`
Volume's `mode` field is set to the default value, we will omit it from
the serialized string.

### DIFF
--- a/schema/types/volume.go
+++ b/schema/types/volume.go
@@ -25,6 +25,12 @@ import (
 	"github.com/appc/spec/schema/common"
 )
 
+const (
+	emptyVolumeDefaultMode = "0755"
+	emptyVolumeDefaultUID  = 0
+	emptyVolumeDefaultGID  = 0
+)
+
 // Volume encapsulates a volume which should be mounted into the filesystem
 // of all apps in a PodManifest
 type Volume struct {
@@ -37,9 +43,9 @@ type Volume struct {
 	ReadOnly *bool  `json:"readOnly,omitempty"`
 
 	// currently used only by "empty"
-	Mode string `json:"mode,omitempty"`
-	UID  int    `json:"uid,omitempty"`
-	GID  int    `json:"gid,omitempty"`
+	Mode *string `json:"mode,omitempty"`
+	UID  *int    `json:"uid,omitempty"`
+	GID  *int    `json:"gid,omitempty"`
 }
 
 type volume Volume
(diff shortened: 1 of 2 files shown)


schema/types/volume_test.go
```diff
@@ -19,7 +19,15 @@ import (
 	"testing"
 )
 
-func TestVolumeFromString(t *testing.T) {
+func sp(s string) *string {
```
crawford: lol

schema/types/volume.go
```diff
@@ -106,12 +114,24 @@ func (v Volume) MarshalJSON() ([]byte, error) {
...
-	s := fmt.Sprintf("%s,kind=%s,readOnly=%t", v.Name, v.Kind, *v.ReadOnly)
+	s := fmt.Sprintf("%s,kind=%s", v.Name, v.Kind)
 	if v.Source != "" {
-		s = s + fmt.Sprintf(",source=%s", v.Source)
+		s += fmt.Sprintf(",source=%s", v.Source)
 	}
-	if v.Mode != "" && v.UID != -1 && v.GID != -1 {
-		s = s + fmt.Sprintf(",mode=%s,uid=%d,gid=%d", v.Mode, v.UID, v.GID)
+	if v.ReadOnly != nil {
+		s += fmt.Sprintf(",readOnly=%t", *v.ReadOnly)
```
jonboulle: should probably switch to a bytes Buffer for all these

ecnahc515: Could you just build a slice of strings `strings.Join` them together at the end?

vcaputo:
lgtm

schema/types/volume.go
```diff
@@ -138,6 +164,7 @@ func VolumeFromString(vp string) (*Volume, error) {
 		return nil, err
 	}
 	for key, val := range v {
+		val := val
```
steveeJ: even if it works this looks weird

jonboulle: it's idiomatic actually :/
https://golang.org/doc/effective_go.html
search for "req := req"

PR #547, opened 2015-11-30
spec: add RestartPolicy type in pod manifest.

yifan-gu:
The optional pod level `RestartPolicy` type defines the behavior of
the apps within a pod when they exit.

This PR adds 3 valid policies:

- always: The app always restarts when it exits, regardless of its
exit status.

- onFailure: The app restarts only when it exits unsuccessfully.

- never: The app never restarts in any cases.

### DIFF
--- a/examples/pod_runtime.json
+++ b/examples/pod_runtime.json
@@ -102,5 +102,6 @@
             "name": "ip-address",
             "value": "10.1.2.3"
         }
-    ]
+    ],
+    "restartPolicy": "onFailure"
 }
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -25,13 +25,14 @@ import (
 const PodManifestKind = types.ACKind("PodManifest")
 
 type PodManifest struct {
-	ACVersion   types.SemVer        `json:"acVersion"`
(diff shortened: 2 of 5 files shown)


yifan-gu:
After some discussion with @jonboulle @philips , we thought it could be less confusion if we just import the kubernetes' restart policy here.
/cc @thockin @vbatts @mpasternacki

yifan-gu:
Not specifying the exponential delay details here as I think it's rather an implementation detail than a spec. But maybe we can leave an example here?

yifan-gu:
ping ?

yifan-gu:
cc @alban @iaguis

yifan-gu:
I found a problem here is, how do we define `restart`? Is that just the restart of the process? how about the side effects left by the previous run of the process? Should we clean it? How?

@jonboulle @philips

krobertson:
Think it would be good to expand the specification some to cover some of the expected behavior, such as the Kubernetes docs do. Primarily around restart policy being a pod setting and how to handle failures of individual apps. Also perhaps around expectations when restarting, such as clearing changed filesytem, and whether `empty` volumes are cleared.

PR #548, opened 2015-12-01
actool: patch-manifest: add dependencies

alban:
Example:
```
$ actool patch-manifest --replace \
        --dependencies=example.com/image-a:example.com/image-b \
        ./etcd-v2.2.2-linux-amd64.aci
```

I would like to have this flag `--dependencies` for adding tests in rkt, related to https://github.com/coreos/rkt/issues/1752.

### DIFF
--- a/actool/manifest.go
+++ b/actool/manifest.go
@@ -50,6 +50,7 @@ var (
 	patchMounts            string
 	patchPorts             string
 	patchIsolators         string
+	patchDependencies      string
 
 	catPrettyPrint bool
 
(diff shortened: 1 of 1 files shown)


actool/manifest.go
```diff
@@ -245,6 +248,20 @@ func patchManifest(im *schema.ImageManifest) error {
 			app.Isolators = append(app.Isolators, *isolator)
 		}
 	}
+
+	if patchDependencies != "" {
+		dependencies := strings.Split(patchDependencies, ":")
```
iaguis: what about versions? they use ":" as a separator

iaguis: I see that the version is encoded as a label in Dependency field, we could somehow support that

alban:
In the end, I wrote a test without `--dependencies=...`. See https://github.com/coreos/rkt/pull/1820. I don't know if this PR is useful. Thoughts?

jonboulle:
I'd rather avoid adding new functionality to actool, in favour of moving to acbuild.

alban:
Ok, closing since I didn't use this in the end.

PR #550, opened 2015-12-02
Add p2p(peer to peer) discovery spec

liugenping:
The  discovery  specification does not meet the P2P download image now , this PR is for download images by P2P. 
The proposal is https://github.com/coreos/rkt/issues/1751   and the PR for rkt is https://github.com/coreos/rkt/pull/1819

### DIFF
--- a/spec/discovery.md
+++ b/spec/discovery.md
@@ -13,16 +13,21 @@ There are three URLs types:
 * Image URLs
 * Signature URLs
 * Public key URLs
+* Torrent URLs
+
+The Torrent URLs are torrent files for download image,signature(optional) and public key(optional) by p2p(peer to peer).Client should parse torrent file and start download image by p2p after torrent file is downloaded.
 
 Simple and Meta Discovery processes use one or more templates (predefined or derived from various sources) to render Image and Signature URLs (while the Public keys URLs aren't templates).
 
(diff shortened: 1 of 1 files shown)


jonboulle:
Hey @liugenping, thanks for the patch. I believe that the discovery semantics as they are today are sufficient to facilitate P2P. Since we do not prescribe a protocol (scheme) for the returned URL, discovery something can simply return something like a [magnet URI](https://en.wikipedia.org/wiki/Magnet_URI_scheme) - for example:
```
  <meta name="ac-discovery" content="example.com magnet:?xt=urn:btih:28ddf5885bc6cae5ckkc5ace36c70a4392f3fcda85&dn=Some+Cool+Thing&tr=udp%3A%2F%2Ftracker.example.com%3A80"
```
then a client that understands magnet links (and in this particular case, torrents) can choose to retrieve the discovered artefact via that method. 

Does that make sense?

liugenping:
@jonboulle  I think is ok. 
1.Peer can get metadata(file name, pieces, pieces length, and hash) from other peers using bittrorrent extension( http://bittorrent.org/beps/bep_0009.html)
2. "tr=" should be retured in  magnet URI as DHT is slowly

jonboulle:
Are you happy for us to close this out then?

Issue #545, opened 2015-11-30
Make insecure flag in discovery granular


krnowak:
In the beginning `insecure` only meant "I want to perform discovery over HTTP if HTTPS didn't work". But right now it means "I want to perform discovery over HTTP if HTTPS didn't work and I even want to send my credentials over HTTP". These meaning should be split.

The functions with `insecure` flag are `discovery.DiscoverPublicKeys`, `discovery.DiscoverEndpoints` and `discovery.DiscoverWalk`. I only checked `discovery/discovery.go` and `discovery/http.go`.

I didn't check if there are some other parts of this project that do similar things.

alban:
I would like an `insecure` flag to allow https with invalid certificates. Right now, it is not possible to use https with invalid certificate even if `insecure=true`. That would be useful for some tests such as https://github.com/coreos/rkt/pull/1822.

So we'd need several flags:
- `insecureTls`: don't check certificates
- `insecureHttp`: allow fallback on http
- maybe: `insecureCredentials`: allow sending credentials on http or on https with unchecked certificate.

blixtra:
What's the status of the `insecureCredentials`? Is this deemed worth having?

jonboulle:
Do we have a use case? Would someone want one but not the other?

blixtra:
IMO, we should just go with insecureHTTP. I would assume that if one uses insecureHTTP then one knows secrets may be sent in the clear and is ok with that.

jonboulle:
Seems reasonable to me.

Chris Kühl <notifications@github.com> schrieb am Fr., 8. Jan. 2016 10:40:

blixtra:
Is it fair to close this then?

PR #551, opened 2015-12-02
discovery: add insecureTls flag to skip TLS checks

alban:
The "insecure" was previously used to allow http. This is now split into
two flags "insecureHttp" and "insecureTls".

Related to https://github.com/appc/spec/issues/545

-----

/cc @krnowak 

I will need this for tests in rkt, see https://github.com/coreos/rkt/pull/1822.

### DIFF
--- a/actool/discover.go
+++ b/actool/discover.go
@@ -36,7 +36,7 @@ var (
 
 func init() {
 	cmdDiscover.Flags.BoolVar(&transportFlags.Insecure, "insecure", false,
-		"Allow insecure non-TLS downloads over http")
+		"Don't check TLS certificates and allow insecure non-TLS downloads over http")
 	cmdDiscover.Flags.BoolVar(&outputJson, "json", false,
 		"Output result as JSON")
 }
@@ -58,7 +58,11 @@ func runDiscover(args []string) (exit int) {
 			stderr("%s: %s", name, err)
 			return 1
 		}
-		eps, attempts, err := discovery.DiscoverEndpoints(*app, nil, transportFlags.Insecure)
+		insecure := discovery.InsecureNone
+		if transportFlags.Insecure {
+			insecure = discovery.InsecureTls | discovery.InsecureHttp
+		}
+		eps, attempts, err := discovery.DiscoverEndpoints(*app, nil, insecure)
 		if err != nil {
 			stderr("error fetching %s: %s", name, err)
 			return 1
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -124,7 +124,7 @@ func createTemplateVars(app App) []string {
 	return tplVars
 }
 
-func doDiscover(pre string, hostHeaders map[string]http.Header, app App, insecure bool) (*Endpoints, error) {
+func doDiscover(pre string, hostHeaders map[string]http.Header, app App, insecure InsecureOption) (*Endpoints, error) {
 	app = *app.Copy()
 	if app.Labels["version"] == "" {
 		app.Labels["version"] = defaultVersion
@@ -173,7 +173,7 @@ func doDiscover(pre string, hostHeaders map[string]http.Header, app App, insecur
 // optionally will use HTTP if insecure is set. hostHeaders specifies the
 // header to apply depending on the host (e.g. authentication). Based on the
 // response of the discoverFn it will continue to recurse up the tree.
-func DiscoverWalk(app App, hostHeaders map[string]http.Header, insecure bool, discoverFn DiscoverWalkFunc) (err error) {
+func DiscoverWalk(app App, hostHeaders map[string]http.Header, insecure InsecureOption, discoverFn DiscoverWalkFunc) (err error) {
 	var (
 		eps *Endpoints
 	)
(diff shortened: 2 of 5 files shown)


jonboulle:
Should we use bit flags rather than separate bools or are we pretty confident these are the only two there'll ever be?

krnowak:
@jonboulle: As stated in #545 by @alban - we will likely need three flags - one for allowing connections over unencrypted HTTP, one for skipping TLS certificates verification and one for allowing credentials passing over insecure connections (either HTTP or HTTPS with unverified TLS certificates).

So I guess it would be better to have bit flags.

Another solution, quicker and less dirty than 3 bools, would be separate bool-like, self-documenting types (like in https://github.com/coreos/rkt/blob/master/rkt/pubkey/pubkey.go#L42)

discovery/http.go
```diff
@@ -49,13 +52,24 @@ func init() {
 			return net.DialTimeout(n, a, defaultDialTimeout)
 		},
 	}
+	tInsecureTls := &http.Transport{
```
krnowak: You could probably do a shallow copy of `http.DefaultTransport` and modify the `TLSClientConfig` field.

```go
tInsecureTls := *http.DefaultTransport
tInsecureTls.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
ClientInsecureTls = &http.Client{
	Transport: &tInsecureTls,
}
```

discovery/discovery.go
```diff
@@ -173,7 +173,7 @@ func doDiscover(pre string, hostHeaders map[string]http.Header, app App, insecur
 // optionally will use HTTP if insecure is set. hostHeaders specifies the
 // header to apply depending on the host (e.g. authentication). Based on the
 // response of the discoverFn it will continue to recurse up the tree.
-func DiscoverWalk(app App, hostHeaders map[string]http.Header, insecure bool, discoverFn DiscoverWalkFunc) (err error) {
+func DiscoverWalk(app App, hostHeaders map[string]http.Header, insecureTls, insecureHttp bool, discoverFn DiscoverWalkFunc) (err error) {
```
krnowak: Ew, I don't like bool parameters. They are very opaque.

krnowak:
This PR doesn't allow to skip sending credentials over HTTP.

discovery/http.go
alban: updated

discovery/discovery.go
alban: updated.

alban:
Updated with a bit field for `discovery.InsecureOption`.

> This PR doesn't allow to skip sending credentials over HTTP

Can this be done separately? Now that it is a bitfield, it should be possible to add after without changing the function prototype.

actool/discover.go
```diff
@@ -58,7 +58,11 @@ func runDiscover(args []string) (exit int) {
 			stderr("%s: %s", name, err)
 			return 1
 		}
-		eps, attempts, err := discovery.DiscoverEndpoints(*app, nil, transportFlags.Insecure)
+		var insecure discovery.InsecureOption
```
krnowak: How about `insecure := discovery.InsecureNone`?

actool/discover.go
```diff
@@ -58,7 +58,11 @@ func runDiscover(args []string) (exit int) {
...
+		if transportFlags.Insecure {
+			insecure = insecure | discovery.InsecureTls | discovery.InsecureHttp
```
krnowak: I suppose you can drop the `insecure | ` part here.

discovery/discovery_test.go
```diff
@@ -248,13 +248,15 @@ func TestDiscoverEndpoints(t *testing.T) {
 
 	for i, tt := range tests {
 		httpDo = tt.do
+		httpDoInsecureTls = tt.do
```
krnowak: This test needs to be extended a bit, we are using insecure flags unconditionally, so discovery will never use `httpDo`.

discovery/http_test.go
```diff
@@ -89,7 +89,7 @@ func fakeHttpOrHttpsGet(filename string, httpSuccess bool, httpsSuccess bool, ht
 func TestHttpsOrHTTP(t *testing.T) {
 	tests := []struct {
 		name          string
-		insecure      bool
+		insecure      InsecureOption
```
krnowak: Maybe extend it, so you can test the insecure tls too?

krnowak:
Looks fine, but the testing is not sufficient.

actool/discover.go
alban: updated.

actool/discover.go
alban: updated.

discovery/discovery_test.go
alban: Tests extended.

discovery/http_test.go
alban: Tests extended.

alban:
@krnowak ready for another review :)

discovery/discovery_test.go
```diff
@@ -248,36 +249,45 @@ func TestDiscoverEndpoints(t *testing.T) {
...
+			InsecureTls,
+			InsecureHttp,
+			InsecureTls | InsecureHttp,
 		}
+		for _, insecure := range insecureList {
+			de, _, err := DiscoverEndpoints(tt.app, hostHeaders, insecure)
+			if err != nil && !tt.expectDiscoverySuccess {
+				continue
+			}
+			if err != nil {
```
krnowak: Should we handle the case when `err == nil` and `tt.expectDiscoverySuccess == false`?

krnowak:
One question, but otherwise LFAD. But won't merge it anyway, I have no power here.

<insert the obvious meme here>

discovery/discovery_test.go
alban: ok, fixed in a separate commit.

alban:
Branch updated.

alban:
@jonboulle can you review this? :)

jonboulle:
thanks!

alban:
For the record, this is in v0.7.4

Issue #538, opened 2015-11-17
dependency order unclear


alban:
[appc/spec](https://github.com/appc/spec/blob/master/spec/aci.md#image-manifest) says:

> The dependencies are applied in order and each image dependency can overwrite files from the previous dependency.

It is not really clear how it works. In https://github.com/coreos/rkt/issues/1752, we have this example of dependencies:

A->B
A->C
A->D

C->B
C->E

D->B
D->E

If we flatten the tree, depth-first, we get the list: A, B, C, E, D. If we flatten the tree, breadth-first, we get the list: A, B, C, D, E. When a file exists in all layers, should we take E's version or D's version? I think the depth-first list makes more sense and we should take E's version.

Another example, with path whitelist:

A->B
A->C
A->D

B->D
B->C

(notice how the order between C and D differs)

The depth-first list is A, B, C, D. If all layers have `/file1` and `/file2`, but B has a path whitelist={/file2}, then `/file1` is hidden from B.aci. Does it mean that we should take D's version of `/file1` but C's version of `/file2`?

jonboulle:
I think it's a little worse than that in that we don't really discuss recursive dependencies at all
https://github.com/appc/spec/blob/master/spec/ace.md#filesystem-setup

Would you mind putting together a proposal to clarify?

PR #555, opened 2015-12-09
spec: clarify image dependencies

alban:
Fixes https://github.com/appc/spec/issues/538

-----

The ordering explained in this PR is the ordering currently implemented. But the chain of `pathWhitelist` seems to be only partially implemented in [GetRenderedACIFromList()](https://github.com/appc/spec/blob/master/pkg/acirenderer/acirenderer.go#L100).

/cc @n0rad

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -27,11 +27,27 @@ Each app in a pod will start chrooted into its own unique read-write filesystem 
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
 
-1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the target directory, in the order in which they are listed.
-2. The `rootfs` contained in the ACI is extracted into the target directory
-3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed from the target directory
+1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the target directory, in the order in which they are listed. The multiple dependencies can form a tree. The dependencies are resolved in the depth-first order.
+2. The `rootfs` contained in the ACI is extracted into the target directory.
+3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed from the target directory. When extracting a file from an image, the file can be filtered by any of the `pathWhitelist` of the image itself and the images depending on the image.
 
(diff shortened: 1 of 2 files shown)


spec/ace.md
```diff
@@ -27,12 +27,28 @@ Each app in a pod will start chrooted into its own unique read-write filesystem
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
 
-1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the target directory, in the order in which they are listed.
-2. The `rootfs` contained in the ACI is extracted into the target directory
-3. If the ACI contains a non-empty `pathWhitelist` field in its `ImageManifest`, *all* paths not in the whitelist must be removed from the target directory
+1. If the ACI contains a non-empty `dependencies` field in its `ImageManifest`, the `rootfs` of each dependent image is extracted into the target directory, in the order in which they are listed. The multiple dependencies can form a tree. The dependencies are resolved in the depth-first order.
```
krnowak: Actually, if we allow dependencies like:
```
A -> [B, C]
B -> [D]
C -> [D]
```

then it is not a tree anymore, but rather a directed acyclic graph, right?

jonboulle: ping @alban

krobertson:
Bump. Was recently going over this as well. LGTM on the ordering.

jonboulle:
@krobertson you are a maintainer ;-)
What about https://github.com/appc/spec/pull/555#discussion_r47113188 ?

krobertson:
@jonboulle I was mainly bumping it since @krnowak's question was still unanswered.

I don't think there is anything wrong with anything. I'd agree its more of a DAG. If there is no protests, I can follow up with a PR to change it.

sgotti:
Sorry for the delay. I write the code but made some assumptions without documenting them. So this LGTM.

> The ordering explained in this PR is the ordering currently implemented. But the chain of pathWhitelist seems to be only partially implemented in GetRenderedACIFromList().

@alban I put a lot of (perhaps confusing) test cases in https://github.com/appc/spec/blob/master/pkg/acirenderer/acirenderer_test.go

I'm quite sure I haven't covered all the possible cases (for example looking now at the code I'm not sure if it's correctly removing unwanted directories). Do you have a case in mind not covered by the actual implementation so we can write it and see what happens?

spec/ace.md
sgotti: Perhaps I creates some confusion in the code comments. Should we also update them?

sgotti: > Actually, if we allow dependencies like:
> 
> A -> [B, C]
> B -> [D]
> C -> [D]
> 
> then it is not a tree anymore, but rather a directed acyclic graph, right?

Thinking about this, in the acirenderer code, also if there's the same imge (meaning same image ID) referenced by mutiple images, every reference is treated a a different image. Si there aren't two connection (edge) to the same image (vertex).

Issue #140, opened 2015-01-22
Should discovery be reframed as a selectable set of discovery strategies?


cdaylward:
Should the discovery process be as prescriptive as it is? I find the discovery mechanisms currently outlined in the specification to be well thought out, however I would like to see discovery reframed as "discovery strategies". Simple discovery is a strategy. Meta-discovery is a strategy. DNS discovery could be a strategy. I believe it would be beneficial to outline the structure of the various strategies, but not prescribe which ones must be used or in which order (or perhaps only a partial order).

As the specification stands now, I imagine some implementations will likely drop discovery as it is currently outlined. For instance, [simple discovery](https://github.com/appc/spec/blob/master/SPEC.md#simple-discovery) might simply be disallowed (production networks might forbid "going out to the internet"). [Meta-discovery](https://github.com/appc/spec/blob/master/SPEC.md#meta-discovery) might also not be necessary and a simple local static lookup table might be preferred (e.g. using configuration files or known directory structure for no-network discovery). Additional proposals such as the adoption of something like DNS discovery as discussed in issue/#3 might also be inappropriate. Users who want to use appc tech to share images on the internet might simply not have access to such infrastructure. These things might lead to either lower adoption or simply ignoring this part of the specification.

If the specification limited itself to outlining common strategies and how they should be structured (e.g. if labels are required in lookup, how they are used in lookup (e.g. key is of form `{name}-{version}-{os}-{arch}`), dependency matching, etc), this might promote the development and sharing of different discovery schemes/libraries/tools.

jonboulle:
+1, as the spec has evolved I would argue that the utility of simple discovery has diminished (and conversely, meta discovery has become more powerful), so I would definitely like to decouple those two for starters.

cnelson:
+1

philips:
I agree with the premise and think we should describe the label matching idea in general, then different mechanisms. I also think it would be cool to have an offline crawler/cache system like what exists at godoc.org so that you would tell the discovery mechanism: "prefix everything from aci-index.corp/" before doing the discovery process.

philips:
@cdaylward Do you want to take a swing at describing a file/table based method, the idea of a "behind firewall" prefixing system and separate out the generic idea of the template matching from the generic concept of the discovery mechanism?

sgotti:
+1.

A note: just only with simple and meta discovery some doubts exited due to discrepancies betwenn these two methods. For example:

* How to implement possible use cases with discovery + label matching? See #117.
[...]

Issue #56, opened 2014-12-19
Simple Discovery: Where does `arch`, `os`, `version` come from?


wereHamster:
When doing the Simple Discovery one is supposed to take the template `https://{name}-{version}-{os}-{arch}.aci` and fill in the variables. In the image manifest only the `name` is defined. The others are optional.

PR #559, opened 2015-12-25
spec/discovery: remove Simple Discovery entirely

jonboulle:
The value of Simple Discovery has always been somewhat dubious, and most
implementations of the spec (`apcera/kurma`, `3ofcoins/jetpack`,
`coreos/rkt`) never actually implemented it. On the other hand, meta
discovery is used actively and arguably is synonymous with discovery.

This PR proposes removing the concept of Simple Discovery entirely and
making that synonymousness official - i.e., App Container Image
Discovery is what was formerly called Meta Discovery.

References:
https://github.com/appc/spec/issues/56
https://github.com/appc/spec/issues/140
https://github.com/appc/spec/pull/118
https://github.com/coreos/rkt/pull/383

/cc @philips @krobertson

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -44,7 +44,7 @@ To make more efficient use of disk space, and improve performance, implementatio
 ### Image Discovery
 
 While the specification prescribes that the Meta Discovery process occurs over HTTP(S), it is intentionally agnostic with respect to the transport over which the discovered resource can be retrieved.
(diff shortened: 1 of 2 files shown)


krobertson:
LGTM. We'd implemented meta discovery, but never did simple discovery. Meta was just far more thorough.

PR #561, opened 2016-01-11
Make initialisms all caps

blixtra:
See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Following from conversation here: https://github.com/coreos/rkt/pull/1945#discussion_r49264529https://github.com/coreos/rkt/pull/1945#discussion_r49264529


iaguis:
LGTM

jonboulle:
LGTM, this is going to break consumers but let's rip that bandaid off

PR #568, opened 2016-02-16
Remove dependency on glibc in linux systems.

jellonek:
On linux platform - remove dependency on macros from glibc headers.
With this change - whole library can be used witouth CGO, so it can be
used to build fully statically linked binary.

### DIFF
--- a/pkg/device/device_linux.go
+++ b/pkg/device/device_linux.go
@@ -1,4 +1,4 @@
-// Copyright 2015 The appc Authors
+// Copyright 2016 The appc Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jellonek:
Btw, return value type of minor/major on linux platforms in C are unsigned int, what on both 32 and 64 bit x86 architectures means uint32 in golang nomenclature, what is not the same as uint, which under 64bit x86 golang have 64bit length.

I did not changed API there, but IMO it should be changed in separate issue.
What do you think about this @iaguis ?

krnowak:
The change looks fine to me. The API change should not be a problem - the API wasn't declared stable yet, IIRC.

pkg/device/device_linux.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2015 The appc Authors
```
krnowak: It's 2016 now. ;)

jellonek: Argh :)
Good catch :P

jonboulle:
I'm struggling to find the history for this one - wasn't there a reason we moved away from redefining to binding?

jellonek:
@jonboulle this was done in https://github.com/appc/spec/commit/39875b0830d196cb57baf0ac5d5457e22e28e12f

it was requested in comment by @alban in https://github.com/appc/spec/commit/85dac8ebb4069373c206e39ca0c0feb6dd5cdebe but still, using only 3 macros from C header files adds IMO unnecessary dependency on CGO.

jonboulle:
I agree (I could have sworn I wrote this myself but must have been in another project)..

Can you please just tweak your commit message to follow our usual format? thanks

jellonek:
Done.

jonboulle:
lgtm

PR #569, opened 2016-02-16
ace: use actual GOOS and GOARCH when building validator ACIs

mpasternacki:
Currently, `os` of built validator ACI is hardcoded to `linux`, which makes ugly patches necessary to make it run properly on Jetpack. This patch makes the build script use current `GOOS` and `GOARCH` values for the `os` and `arch` tags.


jonboulle:
Great, thanks, makes sense to me. (Btw, know I owe you an email...)

PR #576, opened 2016-03-21
discovery: correctly split DiscoverEndpoints and DiscoverPublicKeys

sgotti:
currently DiscoverEndpoints and DiscoverPublicKeys, despite their name,
returns both endpoints and keys.

Additionally their results are different.

For example: suppose in `example.com/images/app` we have this meta tag:

```
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}.{ext}">
```

while in a parent path like `example.com/images` we have this meta tag:

```
<meta name="ac-discovery-pubkeys" content="example.com https://example.com/pubkeys.gpg">
```

Calling DiscoverEndpoints will stop at the meta tags in
`example.com/images/app` and return 1 endpoint and 0 pubkeys

Calling DiscoverPublicKeys won't find any meta tag in
`example.com/images/app` so will search in `example.com/images` and
return 0 endpoints and 1 keys

To avoid these problems these functions should return just their
requested type. So the user is not tempted to use the other type just
calling one of the functions. This mean that endpoints and keys discovery
must do different http calls.

* fix actool that used only DiscoverEndpoints to also get discovery
keys.

* Remove appending of previously discovered endpoint and keys since I
can't find a logic in it (simplyfing various functions).

* Rework tests to check also this and other kind of problems and be more fine grained.


actool/discover.go
```diff
@@ -64,25 +64,40 @@ func runDiscover(args []string) (exit int) {
 		}
 		eps, attempts, err := discovery.DiscoverEndpoints(*app, nil, insecure)
 		if err != nil {
-			stderr("error fetching %s: %s", name, err)
+			stderr("error fetching endpoints for %s: %s", name, err)
 			return 1
 		}
 		for _, a := range attempts {
-			fmt.Printf("discover walk: prefix: %s error: %v\n", a.Prefix, a.Error)
+			fmt.Printf("discover endpoint swalk: prefix: %s error: %v\n", a.Prefix, a.Error)
```
sgotti: typo. Will fix this.

discovery/discovery.go
```diff
@@ -202,14 +205,13 @@ type FailedAttempt struct {
...
 
-func walker(out *Endpoints, attempts *[]FailedAttempt, testFn DiscoverWalkFunc) DiscoverWalkFunc {
-	return func(pre string, eps *Endpoints, err error) error {
+func walker(attempts *[]FailedAttempt, testFn DiscoverWalkFunc) DiscoverWalkFunc {
+	return func(pre string, dd *discoveryData, err error) error {
 		if err != nil {
 			*attempts = append(*attempts, FailedAttempt{pre, err})
 			return nil
 		}
-		out.Append(*eps)
```
jonboulle: I'm dumb and can't figure this out, can you explain it to me?

sgotti: I can't get a sense on appending the endpoint or keys obtained in the child path to the ones obtained in this path since this should never happene because if I get an endpoint at a path I should consider discovery as done and return the result without walking up to the parent path.

Perhaps I'm missing something...

jonboulle: Yeah that's what I was thinking (and also that I was also missing something ;-). Sounds fair to me.

jonboulle:
Thanks! LGTM after typo fix

discovery/discovery.go
sgotti: Probably this was needed when there was a unique function for endpoints and keys discovery to walk up if a endpoint was found but not a keys and merge them. But I'm quite sure this led to some wrong endpoints/keys in the final result depending on the structure of the meta tags in the various paths.

jonboulle: EINSUFFICIENTTESTING

sgotti: Definitely, before merging I like to rework TestDiscovery to handle more cases.

sgotti:
@jonboulle Thanks for your review. I reworked the test to cover more cases, make them for fine grained for every specific case and add some comments around them.

sgotti:
@jonboulle while waiting for merge :stuck_out_tongue: I updated the PR to rename `DiscoverEndpoints` to `DiscoverACIEndpoints` and `Keys` to `PublicKeys` around the code. Also the `actool discovery` output will show `PublicKeys` instead of `Keys`. I don't think if the actool json output is deemed to be stable. If you prefer I'll revert this renaming (or do this in a future PR).

jonboulle:
I think this is fine. Thanks so much!

PR #578, opened 2016-03-22
Godeps: unrewrite import paths

jonboulle:
Rewritten imports can make it difficult to vendor the code in other
packages. The only downside to un-rewriting (AFAICT) is that `actool` is
no longer `go get`able, but this is not something we endorsed in the
first place.


krobertson:
LGTM

Issue #575, opened 2016-03-20
[RFC] Define in the spec (or remove from code) that discovery will set version to "latest" if no version is requested


sgotti:
Actually the spec, in the discovery section, doesn't specify that, if not version is provided, it'll do discovery providing a default version as "latest". But the appc/spec code does this.

This is used for implementing the "latest" pattern. So someone could just define a single template like `https://{name}-{version}.{ext}` and, if an user just requires "myaci" it will use:

`https://myaci-latest.aci`

I have some thought on pros cons of this approach:

* It's handy since with an unique template, a server can implement "latest" just providing an url that can be a symlink/redirect to the real latest aci).

* it's different from the `noarch` approach. Here if someone wants a `noarch` aci it should provide multiple templates in this order:

```
"https://{name}-{version}-{os}-{arch}.{ext}"
"https://{name}-{version}.{ext}"
```

on discovery, not providing `os` and `arch`, the first template will fail to render, and the second will render and be used.

Using the `noarch` approach, the `latest` pattern can be implemented with another template:

```
"https://{name}-{version}-{os}-{arch}.{ext}"
"https://{name}-{os}-{arch}.{ext}"
```

So I have two proposal:

* Define in the SPEC the current behavior. No impact.

* Remove from the appc/spec code the defaulting of version to "latest" if not provided. This will impact current implementation since many expect to abtain the "latest" aci with just one template, but it's probably cleaner.

Before opening a PRs on first or second behavior I'd like to ear your thoughts.

sgotti:
Since I find the second proposal cleaner I went ahead and tried to implement it. This also showed me some additional "problems" so I opened 2 PRs for doing this #580 #581 (and also found other related problems in rkt: coreos/rkt#2311 coreos/rkt#2317).

IMHO this is cleaner than documenting an hidded "latest" behavior and can open the SPEC to some meta discovery implementations that currently will not be possibile BUT can break some implementations that rely on this.

I'd really happy to ear your thoughts on this.

mpasternacki:
The "latest" part always felt like a provisional solution to a real need. Provisional, because it overloads an actual label as a "tag" used only in discovery URL; the version label in the image manifest will be different than what was used to download. Actual need, because it's obviously useful to download the latest version. I'd even go further and say that it's very limiting to only have one "latest" pointer.

[...]

PR #580, opened 2016-03-23
discovery: match template only if satisfies all required labels

sgotti:
This is the first part that tries to implement #575 second proposal (remove the "hidden" behavior that sets a version label to "latest" if not provided)

Actually, the spec isn't clear on template rendering and matching and can block some meta tags implementation as, the current implementation, can return unwanted endpoints.

It can probably BREAK some of the current meta tags implementations. But I'd like to ear your thoughts on this since I cannot find better ideas.

--

discovery: match only templates that satisfy all required labels

Actually the spec doesn't clarify when an endpoint template should be accepted or
not. Now, the appc/spec implementation returns only endpoints that can
be fully rendered. This means that it will also accept a template if it
doesn't contain some of the required discovery labels.

This looks good, but, trying to implement some meta discovery ideas it
will bring to unwanted endpoints.

Example 1:
Suppose I want to implement the "latest" pattern behavior:

```html
<!-- ACIs with specific version -->
<!-- Latest ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest-{os}-{arch}.{ext}">
```

If, on discovery, I ask for the _name_, _os_ and _arch_
labels only the second template will be rendered (since the first cannot
be fully rendered due to missing _version_ label). So I achieved latest
pattern.

But if I'm asking for a specific _version_ both templates will be
rendered.

Example 2:
As an extension of the first example, suppose I want to create a global
meta discovery for all my images on example.com. So on the root of my
server I'll put some meta tags using example.com as prefix:

```html
<!-- ACIs with specific version -->
<!-- noarch ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}-noarch.{ext}">
<!-- Latest ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest-{os}-{arch}.{ext}">
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest.{ext}">
```

With these tags I want to implement a "latest" and also "noarch" pattern.

If, on discovery, I ask only for the _name_ and _version_ labels the
second template will be rendered. So I achieved noarch pattern.

But, unfortunately, also the last template will be rendered.

And, as the first example, if I'm asking for a specific _name_,
_version_, _os_ and _arch_ ALL the templates will be rendered.

Since the spec says:
```
Note that multiple `ac-discovery` tags MAY be returned for a given
prefix-match [snip] In this case, the client implementation MAY choose
which to use at its own discretion.
```

If an implementation chooses the second it can download the wrong image version.

This patch makes template matching stricter choosing only template that
fully satisfy the required labels.

It can probably BREAK some of the current meta tags implementations.

It also documents this behavior.

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -111,6 +111,9 @@ func renderTemplate(tpl string, kvs ...string) (string, bool) {
 	for i := 0; i < len(kvs); i += 2 {
 		k := kvs[i]
 		v := kvs[i+1]
+		if k != "{name}" && !strings.Contains(tpl, k) {
+			return tpl, false
+		}
 		tpl = strings.Replace(tpl, k, v, -1)
 	}
 	return tpl, !templateExpression.MatchString(tpl)
@@ -118,9 +121,11 @@ func renderTemplate(tpl string, kvs ...string) (string, bool) {
 
 func createTemplateVars(app App) []string {
 	tplVars := []string{"{name}", app.Name.String()}
-	// If a label is called "name", it will be ignored as it appears after
-	// in the slice
+	// Ignore labels called "name"
 	for n, v := range app.Labels {
+		if n == "name" {
+			continue
+		}
 		tplVars = append(tplVars, fmt.Sprintf("{%s}", n), v)
 	}
 	return tplVars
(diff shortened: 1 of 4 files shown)


jonboulle:
Please rebase

sgotti:
@jonboulle Rebased. Thoughts on this?

sgotti:
I reworked it to return best matching templates instead of being so strict, it will also cause much less breakage obtaining the same result. So I'm closing this in favor of #596 .

PR #583, opened 2016-03-24
spec: Fix pod spec to clarify the 'mode' field is a string

krobertson:
This corrects the pod spec to clarify that the 'mode' field under
volumes is a string and not an integer. Filing to quote the value will
break unmarshalling a manifest.

@jonboulle


krobertson:
Don't expect any protests to correcting the type to match the code. :)

Issue #479, opened 2015-09-01
Define image name format string.


sgotti:
The image name format string is used inside appc/spec/discovery and also at least by rkt. Do you think it should be part of and defined in appc/spec? If not, the related coded should probably be moved outside.

By now it's something like this

`name[,label1=value][,label2=value],...`
or
`name:version[,label1=value][,label2=value]`

where `name` type is `ACIdentifier`, label type is `ACIdentifier` and label value is string.

A doubt:
Does `name` have to absolutely contain a domain part (for example `coreos.com/etcd`)? If not meta discovery will probably fail as there's not domain to ask to.

At the same time, actually docker2aci creates images from the default docker repository with names like `busybox:latest`.

sgotti:
> Does name have to absolutely contain a domain part (for example coreos.com/etcd)? If not meta discovery will probably fail as there's not domain to ask to.

Partially answering to myself: the spec/aci.md says: `name (string, required) [...] This is not expected to be unique (see the version label) but SHOULD have a URL-like structure to facilitate App Container Image Discovery. [...]`

Since the definition of `name` in the image name string should be the same of the `name` in the image manifest looks like it's not mandatory.

iaguis:
As a side note we could change docker2aci so it generates `index.docker.io/busybox:latest` and it's not a special case.

sgotti:
Few additional thoughts:

* Apossible parsing problem of that format can be caused by labels values containing a comma `,` (as label's values are define of string type). One solution will be to require commas to be escaped (in a ways to define).

* Additionally should the format use the URL escaping rules? The actual parsing function https://github.com/appc/spec/blob/master/discovery/parse.go#L49 uses url.Parse requiring the contents to be url escaped. See for example https://github.com/coreos/rkt/blob/master/rkt/stage1hash.go#L119 that needs to escape the version because the git version contain a `+`.

iaguis:
Related #514

jonboulle:
I think we should probably just bite the bullet and put this somewhere in the spec but I'm not sure where. @sgotti any ideas?

jonboulle:
also #520

sgotti:
@jonboulle mmm. Given that it's not a foundamental part of the spec but it's coded inside the reference go implementation and used by rkt and probably other future project perhaps something in an "extra" directory like suggested by you sounds good.

About the actual format another unresolved problem is the clash with other schemeless formats like filenames: coreos/rkt#715

PR #586, opened 2016-03-29
spec: Define image format string

sgotti:
This patch tries to define image format string (see #479). It tries to handle label values being any kind of string. To do this it requires it to be URL escaped (instead of inventing another escape format). This will break some implementation (see below).

I'm open to any suggestion for other ways to escape the label value.

--
 spec: Define image format string

Add an "extra" spec section for image format string

This format is actually used by the appc spec golang reference
implementation and other tools (like rkt) but it has not been defined.

There're also various unclear cases in the actual implementationt that
this definition tries to clarify, but this is going to change some
behaviors.

The main problem is related to the label values. Since labels are
strings, they can contain any character (also special character like
carriage return, tabs etc...).

With the current implementation there are two main issues relatd to
label value being strings:
* label value containing `,` `:` and `=` cannot be used inside the image format
string. There's the need to find a way to escape them.

* Many special character (like non printable characters) are difficult
to represent in the actual format.

Instead name and label name are of type ACIIdentifier, so they can
contain only a specific set of character and in this set `,`, `:` and
`=` aren't included.

Instead of adding another ad hoc escape rule this spec requires that the
labels value must be URL escaped (percent encoding) with UTF-8 encoding.

This will break some current implementations:

For example rkt stage1 has a version like `1.2.1+gite568957-dirty` and
with the current image format parsing it works. Now, with the need to URL
escape the labels values it should be expressed as
`1.2.1%2Bgite568957-dirty` (since passing `1.2.1+gite568957-dirty` will
be interpreted as `1.2.1 gite568957-dirty`).

### DIFF
--- a/discovery/parse.go
+++ b/discovery/parse.go
@@ -79,7 +79,11 @@ func NewAppFromString(app string) (*App, error) {
 		if err != nil {
 			return nil, err
 		}
-		labels[*labelName] = val[0]
+		labelValue, err := url.QueryUnescape(val[0])
+		if err != nil {
+			return nil, err
+		}
+		labels[*labelName] = labelValue
 	}
 	a, err := NewApp(name, labels)
 	if err != nil {
--- a/discovery/parse_test.go
+++ b/discovery/parse_test.go
@@ -53,13 +53,14 @@ func TestNewAppFromString(t *testing.T) {
 			false,
 		},
 		{
-			"example.com/app:1.2.3,special=!*'();@&+$/?#[],channel=beta",
+			// URL escaped !*'();@&+$/?#[]
+			"example.com/app:1.2.3,special=%21%2A%27%28%29%3B%40%26%2B%24%2F%3F%23%5B%5D%C2%BC%C2%B5%C3%9F,channel=beta",
 
 			&App{
 				Name: "example.com/app",
 				Labels: map[types.ACIdentifier]string{
 					"version": "1.2.3",
-					"special": "!*'();@&+$/?#[]",
+					"special": "!*'();@&+$/?#[]¼µß",
 					"channel": "beta",
 				},
 			},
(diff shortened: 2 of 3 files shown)


jonboulle:
I like this clean-up a lot, but the breakage is quite unpalatable :/. Seems like we dug ourselves a hole by allowing values to be arbitrary. Hmm.

sgotti:
@jonboulle I agree.

If it'll possible to change the label value format (I don't know how many ACIs will break) we could:

* Define only a set of accepted chars (only printable? how to specify all the possible chars in the code to check them?) excluding the single quote `'`.
* Accept single quoted label values.
* Require the use of single quotes around label values when they'll contain commas or colons.


additionally, in the meantime I'll also limit the label value size to max N chars (or bytes?). This will be useful for labels printing and to avoid exceeding max argument sizes and max URL sizes for fetching.

sgotti:
@jonboulle with ACString in #590 one idea will be to just force `,`, `:` and `=` to be escaped with `\` (`\\` to get a `\`) without quoting since it will give some problems when used inside a shell (inside a shell an user can just single quote the whole app string and use `\` to escape the reserved app string chars).

sgotti:
If you have some thoughts on #590 I will update this to match the ACString type and the escaping rules described in the above comment.

lucab:
I think it may be also worth picking a schema prefix (`appc://`?) and specifying that "schema+label" is the canonical way to represent an appc discovery endpoint wherever disambiguation is needed.

sgotti:
@lucab Thanks for respawning this.

I think the scheme will be really good thing (to distinguish between similar discovery strings like the possible future OCI image spec one). There's the question of keeping backward compatibility with the current schemeless and undefined format since it's used.

I'll stay with an URI instead of an URL (or both `appc:` and `appc://`) since the discovery string format is between a location identifier (but I don't consider it a primary access mechanism because it's not a transport) and a name identifier (for the same reasons explained in coreos/rkt#2953).

PR #591, opened 2016-04-06
add an UnformatedString for use with journalctl

sjpotter:
journalctl keeps MACHINE_ID as the uuid but unformatted with "-"

### DIFF
--- a/schema/types/uuid.go
+++ b/schema/types/uuid.go
@@ -39,6 +39,10 @@ func (u UUID) String() string {
 	return fmt.Sprintf("%x-%x-%x-%x-%x", u[0:4], u[4:6], u[6:8], u[8:10], u[10:16])
 }
 
+func (u UUID) UnformattedString() string {
+	return fmt.Sprintf("%x", u)
+}
+
 func (u *UUID) Set(s string) error {
 	nu, err := NewUUID(s)
 	if err == nil {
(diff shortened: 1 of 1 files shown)


sjpotter:
@jonboulle @yifan-gu

schema/types/uuid.go
```diff
@@ -39,6 +39,10 @@ func (u UUID) String() string {
 	return fmt.Sprintf("%x-%x-%x-%x-%x", u[0:4], u[4:6], u[6:8], u[8:10], u[10:16])
 }
 
+func (u UUID) UnformatedString() string {
```
s-urbaniak: Unforma**tt**edString?

schema/types/uuid.go
```diff
@@ -39,6 +39,10 @@ func (u UUID) String() string {
...
+	return fmt.Sprintf("%x", u[0:16])
```
s-urbaniak: `return fmt.Sprintf("%x", u)` is a bit shorter

yifan-gu:
LGTM. You are using this in cAdvisor, right?

sjpotter:
@yifan-gu i was investigating it when using w/ journald as that's the _MACHINE_ID but yes, it might be used with cadvisor too.  It just seemed pointless if things are going to interacts w/ systemd/journald to have to parse it every time.

philips:
Needs a Go doc or something. Also, I don't know if "Unformatted" is the best name here. Perhaps HexString()?

Issue #592, opened 2016-04-07
"actool discover" accepts non existing images


alban:
The image "coreos.com/etcd-etcd-etcd-etcd:v2.0.10" is not supposed to exist, but `actool discover` finds it anyway by truncating the name:

```
$ actool discover coreos.com/etcd-etcd-etcd-etcd:v2.0.10
discover walk: prefix: coreos.com/etcd-etcd-etcd-etcd error: expected a 200 OK got 403
ACI: https://github.com/coreos/etcd/releases/download/v2.0.10/etcd-v2.0.10-linux-amd64.aci, ASC: https://github.com/coreos/etcd/releases/download/v2.0.10/etcd-v2.0.10-linux-amd64.aci.asc
Keys: https://coreos.com/dist/pubkeys/aci-pubkeys.gpg,https://coreos.com/dist/pubkeys/app-signing-pubkey.gpg
```

As a consequence, rkt can run it:
```
$ sudo rkt run --insecure-options=image coreos.com/etcd-etcd-etcd-etcd:v2.0.10
image: using image from local store for image name coreos.com/rkt/stage1-coreos:1.3.0
image: searching for app image coreos.com/etcd-etcd-etcd-etcd
image: remote fetching from URL "https://github.com/coreos/etcd/releases/download/v2.0.10/etcd-v2.0.10-linux-amd64.aci"
Downloading ACI: [=============================================] 3.79 MB/3.79 MB
networking: loading networks from /etc/rkt/net.d
networking: overriding "default" network with long.conf
networking: loading network default with type ptp
[334210.278023] etcd[4]: 2016/04/07 13:46:21 etcd: no data-dir provided, using default data-dir ./default.etcd
```

jonboulle:
@alban what happens without `--insecure-options=image`?

alban:
@jonboulle it gives the following error:

```
Downloading ACI: [=============================================] 3.79 MB/3.79 MB
run: 
  └─error when reading the app name: "coreos.com/etcd-etcd-etcd-etcd" expected but "coreos.com/etcd" found
```

PR #594, opened 2016-04-08
discovery: fix HasPrefix in matching app name

alban:
The following meta:
  <meta name="ac-discovery" content="coreos.com/etcd https://github.com/coreos/etcd/releases/download/{version}/etcd-{version}-{os}-{arch}.{ext}">

Should make the following discoverable:
  actool discover coreos.com/etcd/fdfdsffds:v2.0.10
  actool discover coreos.com/etcd:v2.0.10

But not the following:
  actool discover coreos.com/etcdxxxxxxx:v2.0.10

This is now fixed by checking the character '/' in addition to
HasPrefix.

Fixes https://github.com/appc/spec/issues/592

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -148,6 +148,10 @@ func doDiscover(pre string, hostHeaders map[string]http.Header, app App, insecur
 		if !strings.HasPrefix(app.Name.String(), m.prefix) {
 			continue
 		}
+		if len(app.Name.String()) > len(m.prefix) &&
+			app.Name.String()[len(m.prefix)] != '/' {
+			continue
+		}
 
 		switch m.name {
 		case "ac-discovery":
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -531,6 +531,63 @@ func TestDiscoverEndpoints(t *testing.T) {
 			[]string{"https://example.com/pubkeys.gpg"},
 			testAuthHeader,
 		},
+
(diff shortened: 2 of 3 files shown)


sgotti:
Interesting case.

Now the spec just uses `prefix-match` and says to `ensure that the prefix of the AC Name being discovered matches the prefix-match`, so it needs probably to be changed since this patch implements a more strict behavior.

philips:
Hrm, I think this is unfortunate but expected. Once the client gets the image it would find the name doesn't match.

jonboulle:
I think we should just update https://github.com/appc/spec/pull/594#issuecomment-207515664 to clarify that it should be either be an exact or `filepath.Dir()` match, rather than `HasPrefix()` , no?

sgotti:
Actually it isn't really causing a problem (since in #592 rkt doesn't complain due to the `--insecure-options=image`).

But keeping `prefix-match` and changing the logic seems confusing (since it's a path/dir match)

jonboulle:
@sgotti are you suggesting just leaving this as-is?

sgotti:
@jonboulle I'm on the fence. On one side this isn't causing problems if the discovery url consumer checks the image manifest name with the provided name, on the other side doing a path matching instead of a prefix match make more sense (in my previous sentence I tried to said that, if this is going to be changed I'll try to find a better replacement for the `prefix-match` terms inside https://github.com/appc/spec/blob/master/spec/discovery.md)

PR #596, opened 2016-04-12
discovery: return only best matching templates.

sgotti:
Actually the spec doesn't clarify when an endpoint template should be accepted or
not. Now, the appc/spec implementation returns only endpoints that can
be fully rendered. This means that it will also accept a template if it
doesn't contain some of the required discovery labels.

This looks good, but, trying to implement some meta discovery ideas it
will bring to unwanted endpoints.

Example 1:
Suppose I want to implement the "latest" pattern behavior:

```html
<!-- ACIs with specific version -->
<!-- Latest ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest-{os}-{arch}.{ext}">
```

If, on discovery, I ask for the _name_, _os_ and _arch_
labels only the second template will be rendered (since the first cannot
be fully rendered due to missing _version_ label). So I achieved latest
pattern.

But if I'm asking for a specific _version_ both templates will be
rendered.

Example 2:
As an extension of the first example, suppose I want to create a global
meta discovery for all my images on example.com. So on the root of my
server I'll put some meta tags using example.com as prefix:

```html
<!-- ACIs with specific version -->
<meta name="ac-discovery" content="example.com https://mirror.storage.example.com/{name}-{version}-{os}-{arch}.{ext}">
<!-- noarch ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}-noarch.{ext}">
<meta name="ac-discovery" content="example.com https://mirror.storage.example.com/{name}-{version}-noarch.{ext}">
<!-- Latest ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest-{os}-{arch}.{ext}">
<meta name="ac-discovery" content="example.com https://mirror.storage.example.com/{name}-latest-{os}-{arch}.{ext}">
<!-- Latest noarch ACIs -->
<meta name="ac-discovery" content="example.com https://storage.example.com/{name}-latest-noarch.{ext}">
<meta name="ac-discovery" content="example.com https://mirror.storage.example.com/{name}-latest-noarch.{ext}">
```

With this tags I want to implement global "latest" and "noarch" patterns
and also return multiple mirrors.

If, on discovery, I ask only for the _name_ and _version_ labels the
template 3 and 4 will be rendered. So I achieved a versioned noarch pattern.

But, unfortunately, also template 5, 6, 7, and 8 will be rendered.

And, as the first example, if I'm asking for a specific _name_,
_version_, _os_ and _arch_ ALL the templates will be rendered.

Since the spec says:
```
Note that multiple `ac-discovery` tags MAY be returned for a given
prefix-match [snip] In this case, the client implementation MAY choose
which to use at its own discretion.
```

If an implementation chooses the second it can download the wrong image version.

This patch makes template matching stricter choosing only best matching templates.
Best matching templates are the one where all the templates labels can
be substituted and with the highest number of substituted labels.

With this we can obtain various patterns like latest and noarch and also
keeping the ability to return multiple mirror urls. See also the tests
added in this commit.

It also documents this behavior.

It should not BREAK many of the current meta tags implementation (it
depends on how the client uses the returned endpoints, for example rkt,
currently, only uses the first one)

This is also needed to implement #575 second proposal (remove the "hidden" behavior that sets a version label to "latest" if not provided)


discovery/discovery_test.go
```diff
@@ -488,12 +489,144 @@ func TestDiscoverEndpoints(t *testing.T) {
...
+			[]string{"https://example.com/pubkeys.gpg"},
+			nil,
+		},
+
+		// Test a discovery string that has an hardcoded app name instead of using the provided {name}
+		{
+			&mockHTTPDoer{
+				doer: fakeHTTPGet(
+					[]meta{
+						{"/myapp",
```
jonboulle: formatting is a little weird (elsewhere to) - newline before " ?

discovery/discovery_test.go
```diff
@@ -461,6 +461,7 @@ func TestDiscoverEndpoints(t *testing.T) {
 			nil,
 		},
 		// Test multiple ACIEndpoints.
+		// Should render the first two endpoint, since the others match less labels
```
jonboulle: nits: endpoint->endpoints, less -> fewer

discovery/discovery.go
```diff
@@ -107,20 +107,26 @@ func extractACMeta(r io.Reader) []acMeta {
...
+	return tpl, count, !templateExpression.MatchString(tpl)
 }
 
 func createTemplateVars(app App) []string {
 	tplVars := []string{"{name}", app.Name.String()}
-	// If a label is called "name", it will be ignored as it appears after
-	// in the slice
+	// Ignore labels called "name"
 	for n, v := range app.Labels {
+		if n == "name" {
```
jonboulle: Really we should probably error if this happens

spec/discovery.md
```diff
@@ -59,6 +59,18 @@ curl $(echo "$urltmpl" | sed -e "s/{name}/$name/" -e "s/{version}/$version/" -e
 
 where _name_, _version_, _os_, and _arch_ are set to their respective values for the image, and _ext_ is either `aci` or `aci.asc` for retrieving an App Container Image or signature respectively.
 
+The client MUST accept only best matching templates. Best matching templates are the one where all the templates labels can be substituted and with the highest number of substituted labels.
```
jonboulle: I think it's a little underspecified here which labels clients should/may require. We should probably call that out specifically

jonboulle: > all the templates labels can be substituted and with the highest number of substituted labels

I don't really get the rationale here. If _all_ client-supplied labels _must_ be substituted, then won't be "highest number of substituted labels" always be equal to the number of client-supplied labels?

Or are you talking about templates where the same variable might exist multiple times? e.g. given client-supplied label `{name}`, then `{name}/{name}.{ext}` being somehow preferable to `{name}.{ext}` - if so, why?

jonboulle:
Great write-up :clap:. I mostly agree with your diagnosis and suggestion, just added a few comments on the details.

spec/discovery.md
sgotti: > If all client-supplied labels must be substituted

That was the original idea in #580, in this patch there isn't the requirement to substitute all the client supplied labels but to return only the templates that match the highest number of client-supplied labels.

Perhaps, If you agree on this new idea, the spec description needs your help for better wording :smile:

discovery/discovery_test.go
sgotti: Right. Fixed.

discovery/discovery_test.go
sgotti: Done!

spec/discovery.md
jonboulle: > in this patch there isn't the requirement to substitute all the client supplied labels 

but in the text you have

> all the templates labels can be substituted and with the highest number of substituted labels

Perhaps it should just be this instead?
> with the highest number of substituted labels

sgotti: > but in the text you have
> 
> all the templates labels can be substituted and with the highest number of substituted labels

Sorry, I'll trying to reword it:

`
all the templates expressions can be substituted by the client supplied labels and have the highest number of substituted client supplied labels.
`

sgotti:
@jonboulle Thanks.

I'm on the fence between:

* All client supplied labels must be substituted (like #580)
* Best matching: only return templates with the greater number of substituted client labels (this PR)

IMHO the former is stricter and better (since it'll fails discovery if a template doesn't contains some client provided labels) but I've the fear that it'll break some current implementations (perhaps this is a pointless fear).

I open to change it back to #580 behavior (changing this PR since it provides more tests) if you are ok.

tdavis:
@sgotti If we were to expand the ``meta`` tag format slightly, I think it would be possible to maintain flexibility without introducing the ambiguity that comes from not requiring full substitution. Consider:

    <meta name="ac-discovery" content="example.com https://storage.example.com/{name}/{version}/{arch}/{name}.{ext}" arch="noarch" version="latest">

We could supply optional default values for missing variables. This allows for #580 (which I agree is better than a "best match" which could easily result in invalid URLs) while also avoiding the need to maintain many different tags to cover all the permutations.

Additionally, while I agree in principle that all client supplied labels should be substituted, there's no way to know from the tag what those variables *are* (without implementing the RFC and parsing the URL). Another useful extension would be:

    <meta ... requires="foo,bar,baz"> 

Where the required fields would just be ones present and without default values. Taken together, here's Example 2 above:

    <meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}-{os}-{arch}.{ext}" version="latest" os="any" requires="name">
    <meta name="ac-discovery" content="example.com https://storage.example.com/{name}-{version}-{os}-{arch}.{ext}" version="latest" os="any" requires="name">

This reduces the eight required tags to two, plus gives you OS-independent permutations. It also allows the client to quickly find usable URLs just by checking the ``requires`` attribute against available labels. This also shouldn't affect existing implementations at all. Thoughts?

Issue #598, opened 2016-04-26
Define a default capability set for apps


alban:
appc/spec defines the following [Linux isolators](https://github.com/appc/spec/blob/master/spec/ace.md#linux-isolators):
- os/linux/capabilities-remove-set: list of capabilities that will be removed from the process's capabilities bounding set, all others will be included
- os/linux/capabilities-retain-set: list of capabilities that will be retained in the process's capabilities bounding set, all others will be removed

This implicitly refers to a default set of capabilities that is not clearly defined. For example, if an application wants CAP_SETUID, should the ACI author add it in the retain set? This is actually not necessary, rkt gives CAP_SETUID by default. Docker does the same. But if an application wants to have CAP_NET_ADMIN, the ACI author must add it in the retain set because it is not in the default capability set (neither in rkt or Docker).

When the default set of capabilities is not defined, ACI authors don't know what they should write in the ACI manifest.

The spec should define something like the following:

```
        // appDefaultCapabilities defines a restricted set of capabilities given to
        // apps by default.
        appDefaultCapabilities, _ = types.NewLinuxCapabilitiesRetainSet([]string{
                "CAP_AUDIT_WRITE",
                "CAP_CHOWN",
                "CAP_DAC_OVERRIDE",
                "CAP_FSETID",
                "CAP_FOWNER",
                "CAP_KILL",
                "CAP_MKNOD",
                "CAP_NET_RAW",
                "CAP_NET_BIND_SERVICE",
                "CAP_SETUID",
                "CAP_SETGID",
                "CAP_SETPCAP",
                "CAP_SETFCAP",
                "CAP_SYS_CHROOT",
        }...)
```

(coming from https://github.com/coreos/rkt/pull/2493)

jonboulle:
wanna put up a PR?

alban:
Yes: https://github.com/appc/spec/pull/600

PR #600, opened 2016-05-04
spec/ace.md: define the default capability set

alban:
Fixes https://github.com/appc/spec/issues/598

-----

/cc @iaguis @jonboulle

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -138,16 +138,40 @@ The first example is "capabilities" but this will be expanded to include things 
 
 **Parameters:**
 
-* **set** list of capabilities that will be removed from the process's capabilities bounding set, all others will be included.
+* **set** list of capabilities that will be removed from the process's capabilities bounding set, all others from the default set will be included.
+
+The default set is defined by the following 14 capabilities:
+- `CAP_AUDIT_WRITE`
+- `CAP_CHOWN`
+- `CAP_DAC_OVERRIDE`
+- `CAP_FSETID`
+- `CAP_FOWNER`
+- `CAP_KILL`
+- `CAP_MKNOD`
+- `CAP_NET_RAW`
+- `CAP_NET_BIND_SERVICE`
+- `CAP_SETUID`
+- `CAP_SETGID`
+- `CAP_SETPCAP`
+- `CAP_SETFCAP`
+- `CAP_SYS_CHROOT`
+
+When the app does not have any capability isolator, the process's capabilities bounding set is just the default set defined above.
 
 ```json
 "name": "os/linux/capabilities-remove-set",
 "value": {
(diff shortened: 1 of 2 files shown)


spec/ace.md
```diff
@@ -132,41 +132,62 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
 
-#### os/linux/capabilities-remove-set
+#### os/linux/capabilities-retain-set
 
 * Scope: app
 
 **Parameters:**
 
-* **set** list of capabilities that will be removed from the process's capabilities bounding set, all others will be included.
+* **set** list of capabilities that will be added in the process's capabilities bounding set in addition to the default app capabilities defined thereafter.
```
jonboulle: What you're describing sounds like "add-set", but this is "retain-set". I would expect "retain-set" to reference a final set of capabilities that the process should have, no?

alban: I'm not sure to understand the difference.

If it is the final set of capabilities, what is the remove set below for?

jonboulle: Well, I'd been considering them mutually exclusive. Either you use the retain-set in which case you're specifying exactly which capabilities you want, or you specify nothing in which case you get the default set, or you specify remove-set in which case you get the default set minus remove-set.

vbatts: "retain" sounds much like, "ensure these capabilities are present", but below the remove-set booleans from this list. Sounds a little confusing.

alban: ohh... that's not what has been implemented. It's possible to change this but I wonder if previous images with a retain-set of cap_net_admin would expect other caps as well but would not get them anymore.

Docker also has `--cap-add` and `--cap-drop` that can be used together:
https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

> By default, Docker has a default list of capabilities that are kept.

The default list for Docker is defined there:
https://github.com/docker/docker/blob/master/oci/defaults_linux.go#L64-L79

/cc @s-urbaniak

b-ramsey: While it is possible for earlier versions of application images may be looking for additional capabilities, my understanding of the retain set from the specification was that it allowed the developer to create a whitelist of capabilities their application should have access to.

Additionally, once the remove set is implemented in rkt it will be possible to have a smaller set of capabilities but it is far easier in cases where a small number of capabilities are required to say:
I need CAP_KILL, and CAP_CHOWN only, rather than specifying the 12 or so capabilities that they do not need from the default capability list.

alban: ok, I will update this PR following your semantics.

alban:
Patch updated, PTAL.

jonboulle:
LGTM

PR #601, opened 2016-05-06
actool: implement patch-manifest --revoke-capability

alban:
`actool patch-manifest` was able to add an isolator of type
`os/linux/capabilities-retain-set` with `--capability` but was not able to
add an isolator of type `os/linux/capabilities-remove-set`. This patch
adds this feature with `--revoke-capability`.

This will be useful for https://github.com/coreos/rkt/pull/2589

### DIFF
--- a/actool/manifest.go
+++ b/actool/manifest.go
@@ -47,6 +47,7 @@ var (
 	patchGroup             string
 	patchSupplementaryGIDs string
 	patchCaps              string
+	patchRevokeCaps        string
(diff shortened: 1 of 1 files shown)


jonboulle:
should we error if they're both set?

alban:
@jonboulle no strong opinion here. I could add a check if you would like. On the other hand, not checking if they are both set could be handy to write a rkt unit test with both set to check rkt's behaviour.

PR #603, opened 2016-05-12
spec/ace: add optional read-only rootfs field

### LINKED ISSUE #245: discussion: read-only app container filesystems
This is a discussion topic.

In chatting with @thockin about deploying application containers at scale, one potentially useful pattern that came up is the idea of making the filesystems in which applications execute (i.e. the container chroot) mostly read-only. Specifically, applications in containers can only write to a known part of the filesystem (e.g. a tmpfs mounted at `/home/` or some other location codified in the spec or manifest), and all of the files/directories supplied by the images are not writeable.

The obvious/major downside is that applications need to be designed and built to anticipate this restriction. However, it is generally a vastly simplifying assumption with respect to implementation of filesystem rendering and image management, largely due to the removal of the need to fiddle with copy-on-write mechanisms (all of which, at the time of writing on Linux, have various known issues/caveats). Strategies like making an entire set/repository of images available over network-mounted filesystems become very easy.

It seems somewhat plausible that this could be added to the spec in an optional fashion (as a simple parameter in the image manifest), although the semantics are then slightly unclear in terms of e.g. how read-only should be enforced.

tmrts:
Adds `readOnlyRootFS` field for mounting rootfs of an app as read-only.

Fixes #245 

cc @jonboulle

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -23,7 +23,8 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 
 #### Filesystem Setup
 
-Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
+Each app in a pod will start chrooted into its own unique filesystem before execution.
+If the app's entry in the `PodManifest` has the field `readOnlyRootFS` is set to `true`, the rootfs will be mounted as read-only for that app otherwise it will be mounted as read-write.
 
 An app's filesystem must be *rendered* in an empty directory by the following process (or equivalent):
 
--- a/spec/pods.md
+++ b/spec/pods.md
@@ -58,6 +58,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
                     }
                 ]
             },
+            "readOnlyRootFS": true,
             "mounts": [
                 {
                     "volume": "worklib",
@@ -163,6 +164,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
         * **name** (string, optional) name of the image (restricted to [AC Identifier](types.md#ac-identifier-type) formatting)
         * **labels** (list of objects, optional) additional labels characterizing the image
     * **app** (object, optional) substitute for the app object of the referred image's ImageManifest. See [Image Manifest Schema](aci.md#image-manifest-schema) for what the app object contains.
+    * **readOnlyRootFS** (boolean, optional, defaults to "false" if unsupplied) whether or not the root filesystem of the app will be mounted read-only.
     * **mounts** (list of objects, optional) list of mounts mapping an app mountPoint to a volume. Each mount has the following set of key-value pairs:
       * **volume** (string, required) name of the volume that will fulfill this mount (restricted to the [AC Name](types.md#ac-name-type) formatting); this is a key into the list of `volumes`, below.
       * **path** (string, required) path inside the app filesystem to mount the volume; generally this will come from one of an app's mountPoint paths. For example, if an app has a mountPoint named "work" with path "/var/lib/work", an executor should map an appropriate volume to fulfill that mountPoint by using a `mount` object with that path.
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -152,11 +152,12 @@ func (r Mount) assertValid() error {
 
 // RuntimeApp describes an application referenced in a PodManifest
 type RuntimeApp struct {
-	Name        types.ACName      `json:"name"`
-	Image       RuntimeImage      `json:"image"`
-	App         *types.App        `json:"app,omitempty"`
-	Mounts      []Mount           `json:"mounts,omitempty"`
-	Annotations types.Annotations `json:"annotations,omitempty"`
+	Name           types.ACName      `json:"name"`
+	Image          RuntimeImage      `json:"image"`
+	App            *types.App        `json:"app,omitempty"`
+	ReadOnlyRootFS bool              `json:"readOnlyRootFS,omitempty"`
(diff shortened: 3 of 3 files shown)


spec/pods.md
```diff
@@ -163,6 +163,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
         * **name** (string, optional) name of the image (restricted to [AC Identifier](types.md#ac-identifier-type) formatting)
         * **labels** (list of objects, optional) additional labels characterizing the image
     * **app** (object, optional) substitute for the app object of the referred image's ImageManifest. See [Image Manifest Schema](aci.md#image-manifest-schema) for what the app object contains.
+        * **readOnlyRootFS** (boolean, optional, defaults to "false" if unsupplied) whether or not the root filesystem of the app will be mounted read-only.
```
jonboulle: This should be unindented and live  alongside the app object

schema/types/app.go
```diff
@@ -28,6 +28,7 @@ type App struct {
 	Group             string         `json:"group"`
 	SupplementaryGIDs []int          `json:"supplementaryGIDs,omitempty"`
 	WorkingDirectory  string         `json:"workingDirectory,omitempty"`
+	ReadOnlyRootFS    bool           `json:"readOnlyRootFS,omitempty"`
```
jonboulle: This  should be changed in runtime app in pod.go

jonboulle: The idea is that it's a feature of the runtime rather than the image. So aci stays unchanged. Does that make sense?

spec/ace.md
```diff
@@ -33,6 +33,8 @@ An app's filesystem must be *rendered* in an empty directory by the following pr
 
 If during rootfs extraction a path is already present in the target directory from an earlier dependency, the previously extracted path MUST be overwritten. If the existing path is a symbolic link to a directory, the link MUST NOT be followed and it MUST be removed and replaced with the new path.
 
+If the optional field `readOnlyRootFS` is set to `true` in the ACI, the rootfs will be mounted as read-only.
```
jonboulle: Nit, could you clarify line 26 where it says read write. Perhaps just add "by default"

spec/pods.md
tmrts: ack

schema/types/app.go
tmrts: Yup, I confused the execution flow for a sec there. fixed now

spec/ace.md
tmrts: clarified it.

tmrts:
@jonboulle PTAL

spec/ace.md
```diff
@@ -23,7 +23,9 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 
 #### Filesystem Setup
 
-Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
+Each app in a pod will start chrooted into its own unique filesystem before execution.
+If the optional field `apps.readOnlyRootFS` is set to `true` in the ACI,
```
jonboulle: NACK, this is _not_ a field in the ACI - it's in the pod manifest.
"If the app's entry in the `PodManifest` has the field `readOnlyRootFS` is set to `true`, ..."

tmrts: done

spec/pods.md
```diff
@@ -163,6 +163,7 @@ JSON Schema for the Pod Manifest, conforming to [RFC4627](https://tools.ietf.org
         * **name** (string, optional) name of the image (restricted to [AC Identifier](types.md#ac-identifier-type) formatting)
         * **labels** (list of objects, optional) additional labels characterizing the image
     * **app** (object, optional) substitute for the app object of the referred image's ImageManifest. See [Image Manifest Schema](aci.md#image-manifest-schema) for what the app object contains.
+    * **readOnlyRootFS** (boolean, optional, defaults to "false" if unsupplied) whether or not the root filesystem of the app will be mounted read-only.
```
jonboulle: perhaps add an example  readOnlyRootFS: true to the example JSON here? https://github.com/appc/spec/pull/603/files#diff-7bbac9ed3bbf6dbc1688669ce478f5deL113

spec/ace.md
```diff
@@ -23,7 +23,9 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 
 #### Filesystem Setup
 
-Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
+Each app in a pod will start chrooted into its own unique filesystem before execution.
+If the app's entry in the `PodManifest` has the field `apps.readOnlyRootFS` is set to `true`,
```
jonboulle: the "apps." prefix is wrong, just remove that.

tmrts: I could've sworn I removed that

spec/pods.md
tmrts: done

spec/ace.md
```diff
@@ -23,7 +23,9 @@ This UUID is exposed to the pod through the [Metadata Service](#app-container-me
 
 #### Filesystem Setup
 
-Each app in a pod will start chrooted into its own unique read-write filesystem before execution.
+Each app in a pod will start chrooted into its own unique filesystem before execution.
+If the app's entry in the `PodManifest` has the field `readOnlyRootFS` is set to `true`,
```
jonboulle: sorry, two more things I missed:
a) put these on one line (one-sentence-per-line is still the convention here)
b) s/for each app/for that app,/

tmrts: np, done

jonboulle:
LGTM, thank you

PR #606, opened 2016-05-20
isolator_resources.go: Fix build

pquerna:
resourceValue struct uses a pointer to resource.Quantity.

Before:
```
$ go build github.com/appc/spec/schema/types
# github.com/appc/spec/schema/types
./isolator_resources.go:158: cannot use req (type resource.Quantity) as type *resource.Quantity in field value
./isolator_resources.go:159: cannot use lim (type resource.Quantity) as type *resource.Quantity in field value
./isolator_resources.go:212: cannot use req (type resource.Quantity) as type *resource.Quantity in field value
./isolator_resources.go:213: cannot use lim (type resource.Quantity) as type *resource.Quantity in field value
```

After: It compiles.

PR #607, opened 2016-05-20
Godeps: Update k8s.io packages

dlorenc:
- Change the location of the speter.net package, it has moved and no longer exists in that location.
- Fix isolator_resources.go to handle the new k8s.io/kubernetes/pkg/api/resource interface.


dlorenc:
Whoops, just noticed that this fixes the same problem addressed by #606. I think this is complete though.

jonboulle:
LGTM. Thanks for the patch!

PR #613, opened 2016-05-25
deps:  gogo/protobuf: use v0.2

steveeJ:
Motivation is to vendor only tagged dependencies.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,7 @@
 {
 	"ImportPath": "github.com/appc/spec",
 	"GoVersion": "go1.5",
-	"GodepVersion": "v63",
+	"GodepVersion": "v70",
 	"Packages": [
 		"./..."
 	],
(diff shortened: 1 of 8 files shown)


alban:
@lucab @onlyjob: rkt already depends on an unspecified version of protobuf:
https://anonscm.debian.org/cgit/pkg-go/packages/rkt.git/tree/debian/control#n50

Does it mean it is unnecessary to switch to a tagged version for this specific package?

alban:
For the context: this is downgrading the dependency on gogo/protobuf to avoid a dependency on an untagged version. But v0.2 is quite old. A new tag has been requested on https://github.com/gogo/protobuf/issues/182

onlyjob:
> rkt (package) already depends on an unspecified version of protobuf:

`golang-goprotobuf-dev` provides `github.com/golang/protobuf`.
If _rkt_ gained _direct_ dependency on `github.com/gogo/protobuf` then `golang-github-gogo-protobuf-dev` package (currently v0.2) should be added to rkt's Build-Depends.

Properly versioning dependencies is very helpful.

If only some of _rkt_ dependency packages use `github.com/gogo/protobuf` then corresponding packages should be updated with versioned dependency on `golang-github-gogo-protobuf-dev (>= 0.2)` (or higher, as required).

onlyjob:
> But v0.2 is quite old.

That's rapid development, isn't it?  ;)
gogo-protobuf 0.2 is just about 2 months (and 17 commits) "old"!  :)

lucab:
> For the context: this is downgrading the dependency on gogo/protobuf to avoid a dependency on an untagged version.

This doesn't look like a downgrade to me. Current revision falls between 0.1 and 0.2. If it is not causing any new regressions, I think this should be ok to be merged. Debian already has `golang-github-gogo-protobuf-dev` 0.2 anyway.

jonboulle:
lgtm

PR #615, opened 2016-05-26
travis: stop travis manipulating godeps

jonboulle:


.travis.yml
```diff
@@ -7,5 +7,9 @@ matrix:
     - go: 1.5.1
     - go: 1.6
```
steveeJ: Should we also update to 1.5.3 and 1.6.2?

PR #618, opened 2016-05-26
deps:  use native vendoring

steveeJ:
Closes #617. 

/cc @jonboulle


jonboulle:
LGTM. Thanks for the fix!

PR #619, opened 2016-05-26
build/test: fail on Go1 .5

jonboulle:
Go 1.5 needs the environment flag `GO15VENDOREXPERIMENT=1` set to detect the deps in the _vendor/_ directory. Ideally it would be done dynamically, but could just be added to the scripts statically too.

/cc @jonboulle

### DIFF
--- a/build
+++ b/build
@@ -11,7 +11,7 @@ fi
 
 export GOBIN=${PWD}/bin
 export GOPATH=${PWD}/gopath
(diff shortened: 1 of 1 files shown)


jonboulle:
@steveeJ

steveeJ:
To test this you could remove the `GO15VENDOREXPERIMENT=1` from the travis configuration.

PR #620, opened 2016-05-31
[WIP] ace: initial seccomp proposal

lucab:
> WORK IN PROGRESS

An initial proposal for seccomp support as an ACE Linux-specific isolator.

### DIFF
--- a/schema/types/isolator_linux_specific.go
+++ b/schema/types/isolator_linux_specific.go
@@ -17,12 +17,15 @@ package types
 import (
 	"encoding/json"
 	"errors"
+	"unicode"
 )
 
 const (
 	LinuxCapabilitiesRetainSetName = "os/linux/capabilities-retain-set"
 	LinuxCapabilitiesRevokeSetName = "os/linux/capabilities-remove-set"
 	LinuxNoNewPrivilegesName       = "os/linux/no-new-privileges"
+	LinuxSeccompRemoveSetName      = "os/linux/seccomp-remove-set"
+	LinuxSeccompRetainSetName      = "os/linux/seccomp-retain-set"
 )
 
 var LinuxIsolatorNames = make(map[ACIdentifier]struct{})
@@ -32,6 +35,8 @@ func init() {
 		LinuxCapabilitiesRevokeSetName: func() IsolatorValue { return &LinuxCapabilitiesRevokeSet{} },
 		LinuxCapabilitiesRetainSetName: func() IsolatorValue { return &LinuxCapabilitiesRetainSet{} },
 		LinuxNoNewPrivilegesName:       func() IsolatorValue { v := LinuxNoNewPrivileges(false); return &v },
+		LinuxSeccompRemoveSetName:      func() IsolatorValue { return &LinuxSeccompRemoveSet{} },
+		LinuxSeccompRetainSetName:      func() IsolatorValue { return &LinuxSeccompRetainSet{} },
 	} {
 		AddIsolatorName(name, LinuxIsolatorNames)
 		AddIsolatorValueConstructor(name, con)
@@ -54,6 +59,10 @@ func (l *LinuxNoNewPrivileges) UnmarshalJSON(b []byte) error {
 	*l = LinuxNoNewPrivileges(v)
 
 	return nil
+}
+
+type AsIsolator interface {
+	AsIsolator() (*Isolator, error)
 }
 
 type LinuxCapabilitiesSet interface {
@@ -115,17 +124,17 @@ func NewLinuxCapabilitiesRetainSet(caps ...string) (*LinuxCapabilitiesRetainSet,
 	return &l, nil
 }
 
-func (l LinuxCapabilitiesRetainSet) AsIsolator() Isolator {
+func (l LinuxCapabilitiesRetainSet) AsIsolator() (*Isolator, error) {
 	b, err := json.Marshal(l.linuxCapabilitiesSetBase.val)
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 	rm := json.RawMessage(b)
-	return Isolator{
+	return &Isolator{
 		Name:     LinuxCapabilitiesRetainSetName,
 		ValueRaw: &rm,
 		value:    &l,
-	}
+	}, nil
 }
 
 type LinuxCapabilitiesRevokeSet struct {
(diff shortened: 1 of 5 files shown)


jonboulle:
Looks fantastic!

lucab:
This also captures some prior offline discussions with @s-urbaniak, which may want to chime in to double-check.

schema/types/isolator_linux_specific.go
```diff
@@ -161,3 +170,122 @@ func (l LinuxCapabilitiesRevokeSet) AsIsolator() Isolator {
...
+func (l *linuxSeccompBase) UnmarshalJSON(b []byte) error {
+	var v linuxSeccompValue
+	err := json.Unmarshal(b, &v)
+	if err != nil {
+		return err
+	}
+
+	l.val = v
+
+	return err
```
s-urbaniak: `return nil`, err cannot be non-nil here

schema/types/isolator_linux_specific.go
```diff
@@ -161,3 +170,122 @@ func (l LinuxCapabilitiesRevokeSet) AsIsolator() Isolator {
...
+	if err := l.AssertValid(); err != nil {
+		return nil, err
+	}
+	return &l, nil
+}
+
+func (l LinuxSeccompRetainSet) AsIsolator() Isolator {
+	b, err := json.Marshal(l.linuxSeccompBase.val)
+	if err != nil {
+		panic(err)
```
s-urbaniak: I suggest to return `err` here instead of panic'ing, can we extend the signature of the `Isolator` interface for this?

schema/types/isolator_linux_specific.go
```diff
@@ -161,3 +170,122 @@ func (l LinuxCapabilitiesRevokeSet) AsIsolator() Isolator {
...
+	if err := l.AssertValid(); err != nil {
+		return nil, err
+	}
+	return &l, nil
+}
+
+func (l LinuxSeccompRemoveSet) AsIsolator() Isolator {
+	b, err := json.Marshal(l.linuxSeccompBase.val)
+	if err != nil {
+		panic(err)
```
s-urbaniak: same here, see above

schema/types/isolator_linux_specific.go
```diff
@@ -161,3 +170,122 @@ func (l LinuxCapabilitiesRevokeSet) AsIsolator() Isolator {
...
+	Set   []LinuxSeccompEntry `json:"set"`
+	Errno LinuxSeccompErrno   `json:"errno"`
+}
+
+type linuxSeccompBase struct {
+	val linuxSeccompValue
+}
+
+func (l linuxSeccompBase) AssertValid() error {
+	if l.val.Errno != "" {
```
s-urbaniak: let's flatten this a bit:
```
if l.val.Errno == "" {
  return nil
}

for _, c := range l.val.Errno {
...

```

s-urbaniak:
@lucab just nits really :-)

As far as I remember we discussed that the seccomp filter doesn't really make sense without used in conjunction with the NoNewPrivileges isolator. Is this something we should document/code?

lucab:
@s-urbaniak yes, in the bigger picture there is some relationship between seccomp, capabilities and no-new-privs to avoid unprivileged user re-gaining enough privileges (CAP_SYS_ADMIN) to revert the filter. But I don't think this should be part of the spec (or maybe just a "MAY" note).

lucab:
I'll let this sediment this here for some time, waiting for major concerns on the spec side. I'll then move it to non-wip with a final refinement pass.

schema/types/isolator_linux_specific.go
jonboulle: @s-urbaniak why? if something goes wrong with marshalling IMHO we should fail hard - it's either a serious programming error or runtime failure

lucab: @jonboulle I think his point is that it should be an hard failure, but not here. The caller should be able to handle this gracefully, instead of having the library blowing up.

s-urbaniak: @jonboulle my intuitive reaction was based on the fact that this marshaling operation is based on user input.

We do maintain the invariant of providing a string type only for the values to be marshaled, hence Marshal cannot return [`UnsupportedTypeError`](https://godoc.org/encoding/json#UnsupportedTypeError)/[`UnsupportedValueError`](https://godoc.org/encoding/json#UnsupportedValueError) but this assumes knowledge of the implementation details and invariants of `encoding/json` which might change unnoticed.

I'll leave @lucab the final decision ;-)

lucab:
Moved to proposal stage, spec and code now at #621.

spec/ace.md
```diff
@@ -131,7 +131,68 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+
+**Example:**
+```json
+"name": "os/linux/seccomp-remove-set",
+"value": {
+  "errno": "ENOTSUP",
+  "set": [
+    "chown",
+    "chmod"
```
alban: Should we give an example with some syscalls that should reasonably not be given to containers, so that it is more realistic? Such as:
- reboot
- init_module, finit_module, delete_module
- kexec_load

Some of them found from [systemd-nspawn](https://github.com/systemd/systemd/blob/master/src/nspawn/nspawn-seccomp.c#L46).

spec/ace.md
```diff
@@ -131,7 +131,68 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+* **set** case-sensitive list of syscall names that will be whitelisted (ie. not blocked). All syscalls specified in this set MUST NOT be blocked. This set MAY be augmented by an implementation-specific default whitelist set (see notes below). Syscalls not specified in the union of these two sets MUST be blocked.
+* **errno** all-uppercase name of a single [errno code](http://man7.org/linux/man-pages/man3/errno.3.html) that will be returned by blocked system calls, instead of terminating. If missing or empty string, by default blocked syscalls MUST result in app termination via `SIGSYS` signal.
+
+**Notes:**
+ 1. The `os/linux/seccomp-retain-set` isolator cannot be used in conjunction with the `os/linux/seccomp-remove-set` isolator.
+ 2. When an app does not have any seccomp isolator (neither `os/linux/seccomp-remove-set` nor `os/linux/seccomp-retain-set` are specified), implementations MAY apply their own set of blocked syscalls.
+ 3. The implementation-specific default whitelist MUST be the minimum set of syscalls required for app life-cycle. For example, implementations may require the `exit(2)` syscall to be always allowed for clean app termination. This set MAY be empty.
```
alban: When would that implementation-specific default whitelist be used? When there is a `os/linux/seccomp-retain-set` isolator defined but with `"set": []`?

spec/ace.md
```diff
@@ -131,7 +131,68 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc/` top-level scope is reserved for usage in this specification. Implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
+  * `@appc/all` represents the set of all available syscalls.
```
alban: Could you give an example how `@appc/all` could be used? I don't see a use case.

With `os/linux/seccomp-remove-set`, would that block _exit and sigreturn as well?

With `os/linux/seccomp-retain-set`, is it equivalent to not have any isolator?

spec/ace.md
lucab: Whenever `os/linux/seccomp-retain-set` isolator is enabled. The resulting whitelist is the union of the two sets (the one specified in the app isolator and the implementation-specific one):
> This set MAY be augmented by an implementation-specific default whitelist set (see notes below). Syscalls not specified in the union of these two sets MUST be blocked.

spec/ace.md
lucab: I think it will make sense, yes. Another source for examples is [Docker blacklist](https://github.com/docker/docker/blob/master/docs/security/seccomp.md)

spec/ace.md
lucab: > Could you give an example how @appc/all could be used? I don't see a use case.

I reworded this in the final proposal as I realized it was not clear.

> With os/linux/seccomp-remove-set, would that block _exit and sigreturn as well?

This was my mistake in the original spec, I failed at describing it. The new one should address it.

> With os/linux/seccomp-retain-set, is it equivalent to not have any isolator?

it is equivalent to not apply any seccomp filter. This is specifically to opt-out from default filtering.

PR #621, opened 2016-06-02
ace: add Linux-specific seccomp isolator

### LINKED ISSUE #529: spec: add seccomp isolator for Linux
It would be nice to expose seccomp isolators to allow apps/pods running on Linux to restrict the permitted set of system calls.

http://man7.org/linux/man-pages/man2/seccomp.2.html
https://en.wikipedia.org/wiki/Seccomp

lucab:
This is a proposal for two Linux-specific security isolators based on [seccomp filters](https://www.kernel.org/doc/Documentation/prctl/seccomp_filter.txt). It follows closely the capabilities isolators, where two modes are provided: "remove-set" (blacklisting) and "retain-set" (whitelisting).

Given the huge number of syscalls that can be specified this way and the on-going discussion about sensible defaults, this specification doesn't hardcode any list but let implementations some freedom to provide groups/wildcards via `@`-prefixed special values. Users can decide to provide their own custom filters, or just opt-in/opt-out those implementations-provided filters.

Fixes #529

### DIFF
--- a/actool/manifest.go
+++ b/actool/manifest.go
@@ -51,6 +51,8 @@ var (
 	patchMounts            string
 	patchPorts             string
 	patchIsolators         string
+	patchSeccompMode       string
+	patchSeccompSet        string
 
 	catPrettyPrint bool
 
@@ -69,6 +71,8 @@ var (
 		  [--ports=query,protocol=tcp,port=8080[:query2,...]]
 		  [--supplementary-groups=gid1,gid2,...]
 		  [--isolators=resource/cpu,request=50m,limit=100m[:resource/memory,...]]
+		  [--seccomp-mode=remove|retain[,errno=EPERM]]
+		  [--seccomp-set=syscall1,syscall2,...]]
 		  [--replace]
 		  INPUT_ACI_FILE
 		  [OUTPUT_ACI_FILE]`,
@@ -99,6 +103,8 @@ func init() {
 	cmdPatchManifest.Flags.StringVar(&patchMounts, "mounts", "", "Replace mount points")
 	cmdPatchManifest.Flags.StringVar(&patchPorts, "ports", "", "Replace ports")
 	cmdPatchManifest.Flags.StringVar(&patchIsolators, "isolators", "", "Replace isolators")
+	cmdPatchManifest.Flags.StringVar(&patchSeccompMode, "seccomp-mode", "", "Enable and configure seccomp isolator")
+	cmdPatchManifest.Flags.StringVar(&patchSeccompSet, "seccomp-set", "", "Set of syscalls for seccomp isolator enforcing")
 
 	cmdCatManifest.Flags.BoolVar(&catPrettyPrint, "pretty-print", false, "Print with better style")
 }
@@ -211,7 +217,11 @@ func patchManifest(im *schema.ImageManifest) error {
 		if err != nil {
 			return fmt.Errorf("cannot parse capability %q: %v", patchCaps, err)
 		}
-		app.Isolators = append(app.Isolators, caps.AsIsolator())
+		isolator, err = caps.AsIsolator()
+		if err != nil {
+			return err
+		}
+		app.Isolators = append(app.Isolators, *isolator)
 	}
 	if patchRevokeCaps != "" {
 		isolator := app.Isolators.GetByName(types.LinuxCapabilitiesRevokeSetName)
@@ -225,7 +235,11 @@ func patchManifest(im *schema.ImageManifest) error {
 		if err != nil {
 			return fmt.Errorf("cannot parse capability %q: %v", patchRevokeCaps, err)
 		}
-		app.Isolators = append(app.Isolators, caps.AsIsolator())
+		isolator, err = caps.AsIsolator()
+		if err != nil {
+			return err
+		}
+		app.Isolators = append(app.Isolators, *isolator)
 	}
 
 	if patchMounts != "" {
@@ -250,6 +264,18 @@ func patchManifest(im *schema.ImageManifest) error {
 		}
 	}
 
+	// Parse seccomp args and override existing seccomp isolators
+	if patchSeccompMode != "" {
+		seccompIsolator, err := parseSeccompArgs(patchSeccompMode, patchSeccompSet)
+		if err != nil {
+			return err
+		}
+		seccompReps := []types.ACIdentifier{types.LinuxSeccompRemoveSetName, types.LinuxSeccompRetainSetName}
+		app.Isolators.ReplaceIsolatorsByName(*seccompIsolator, seccompReps)
+	} else if patchSeccompSet != "" {
+		return fmt.Errorf("--seccomp-set specified without --seccomp-mode")
+	}
+
 	if patchIsolators != "" {
 		isolators := strings.Split(patchIsolators, ":")
 		for _, is := range isolators {
@@ -260,14 +286,16 @@ func patchManifest(im *schema.ImageManifest) error {
 
 			_, ok := types.ResourceIsolatorNames[name]
 
-			if name == types.LinuxNoNewPrivilegesName {
+			switch name {
+			case types.LinuxNoNewPrivilegesName:
 				ok = true
 				kv := strings.Split(is, ",")
 				if len(kv) != 2 {
 					return fmt.Errorf("isolator %s: invalid format", name)
 				}
-
 				isolatorStr = fmt.Sprintf(`{ "name": "%s", "value": %s }`, name, kv[1])
+			case types.LinuxSeccompRemoveSetName, types.LinuxSeccompRetainSetName:
+				ok = false
 			}
 
 			if !ok {
@@ -282,6 +310,51 @@ func patchManifest(im *schema.ImageManifest) error {
 		}
 	}
 	return nil
+}
+
+// parseSeccompArgs parses seccomp mode and set CLI flags, preparing an
+// appropriate seccomp isolator.
+func parseSeccompArgs(patchSeccompMode string, patchSeccompSet string) (*types.Isolator, error) {
+	// Parse mode flag and additional keyed arguments.
+	var errno, mode string
+	args := strings.Split(patchSeccompMode, ",")
+	for _, a := range args {
+		kv := strings.Split(a, "=")
+		switch len(kv) {
+		case 1:
+			// mode, either "remove" or "retain"
+			mode = kv[0]
+		case 2:
+			// k=v argument, only "errno" allowed for now
+			if kv[0] == "errno" {
+				errno = kv[1]
+			} else {
+				return nil, fmt.Errorf("invalid seccomp-mode optional argument: %s", a)
+			}
+		default:
+			return nil, fmt.Errorf("cannot parse seccomp-mode argument: %s", a)
+		}
+	}
+
+	// Instantiate an Isolator with the content specified by the --seccomp-set parameter.
+	var err error
+	var seccomp types.AsIsolator
+	switch mode {
+	case "remove":
+		seccomp, err = types.NewLinuxSeccompRemoveSet(errno, strings.Split(patchSeccompSet, ",")...)
+	case "retain":
+		seccomp, err = types.NewLinuxSeccompRetainSet(errno, strings.Split(patchSeccompSet, ",")...)
+	default:
+		err = fmt.Errorf("unknown seccomp mode %s", mode)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse seccomp isolator: %s", err)
+	}
+	seccompIsolator, err := seccomp.AsIsolator()
+	if err != nil {
+		return nil, err
+	}
+	return seccompIsolator, nil
 }
 
 // extractManifest iterates over the tar reader and locate the manifest. Once
--- a/examples/image.json
+++ b/examples/image.json
@@ -60,6 +60,17 @@
             {
                 "name": "os/linux/no-new-privileges",
                 "value": true
+            },
+            {
+                "name": "os/linux/seccomp-remove-set",
+                "value": {
+                    "errno": "EACCESS",
+                    "set": [
+                        "clock_settime",
+                        "clock_adjtime",
+                        "reboot"
+                    ]
+                }
             }
         ],
         "mountPoints": [
--- a/schema/types/isolator.go
+++ b/schema/types/isolator.go
@@ -16,10 +16,14 @@ package types
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 )
 
 var (
 	isolatorMap map[ACIdentifier]IsolatorValueConstructor
+
+	ErrIncompatibleIsolator = errors.New("isolators set contains incompatible types")
 )
 
 func init() {
@@ -40,6 +44,29 @@ func AddIsolatorName(n ACIdentifier, ns map[ACIdentifier]struct{}) {
 // and PodManifest schemas.
 type Isolators []Isolator
 
+// assertValid checks that every single isolator is valid and that
+// the whole set is well built
+func (isolators Isolators) assertValid() error {
+	typesMap := make(map[ACIdentifier]bool)
+	for _, i := range isolators {
+		if err := i.Value().AssertValid(); err != nil {
+			return err
+		}
+		if _, ok := typesMap[i.Name]; ok {
+			if !i.Value().multipleAllowed() {
+				return fmt.Errorf(`isolators set contains too many instances of type %s"`, i.Name)
+			}
+		}
+		for _, c := range i.Value().Conflicts() {
+			if _, found := typesMap[c]; found {
+				return ErrIncompatibleIsolator
+			}
+		}
+		typesMap[i.Name] = true
+	}
+	return nil
+}
+
 // GetByName returns the last isolator in the list by the given name.
 func (is *Isolators) GetByName(name ACIdentifier) *Isolator {
 	var i Isolator
@@ -50,6 +77,22 @@ func (is *Isolators) GetByName(name ACIdentifier) *Isolator {
 		}
 	}
 	return nil
+}
+
+// ReplaceIsolatorsByName overrides matching isolator types with a new
+// isolator, deleting them all and appending the new one instead
+func (is *Isolators) ReplaceIsolatorsByName(newIs Isolator, oldNames []ACIdentifier) {
+	var i Isolator
+	for j := len(*is) - 1; j >= 0; j-- {
+		i = []Isolator(*is)[j]
+		for _, name := range oldNames {
+			if i.Name == name {
+				*is = append((*is)[:j], (*is)[j+1:]...)
+			}
+		}
+	}
+	*is = append((*is)[:], newIs)
+	return
 }
 
 // Unrecognized returns a set of isolators that are not recognized.
@@ -69,8 +112,17 @@ func (is *Isolators) Unrecognized() Isolators {
 // serialized as any arbitrary JSON blob. Specific Isolator types should
 // implement this interface to facilitate unmarshalling and validation.
 type IsolatorValue interface {
+	// UnmarshalJSON unserialize a JSON-encoded isolator
 	UnmarshalJSON(b []byte) error
+	// AssertValid returns a non-nil error value if an IsolatorValue is not valid
+	// according to appc spec
 	AssertValid() error
+	// Conflicts returns a list of conflicting isolators types, which cannot co-exist
+	// together with this IsolatorValue
+	Conflicts() []ACIdentifier
+	// multipleAllowed specifies whether multiple isolator instances are allowed
+	// for this isolator type
+	multipleAllowed() bool
 }
 
 // Isolator is a model for unmarshalling isolator types from their JSON-encoded
--- a/schema/types/isolator_linux_specific.go
+++ b/schema/types/isolator_linux_specific.go
@@ -17,12 +17,15 @@ package types
 import (
 	"encoding/json"
 	"errors"
+	"unicode"
 )
 
 const (
 	LinuxCapabilitiesRetainSetName = "os/linux/capabilities-retain-set"
 	LinuxCapabilitiesRevokeSetName = "os/linux/capabilities-remove-set"
 	LinuxNoNewPrivilegesName       = "os/linux/no-new-privileges"
+	LinuxSeccompRemoveSetName      = "os/linux/seccomp-remove-set"
+	LinuxSeccompRetainSetName      = "os/linux/seccomp-retain-set"
 )
 
 var LinuxIsolatorNames = make(map[ACIdentifier]struct{})
@@ -32,6 +35,8 @@ func init() {
 		LinuxCapabilitiesRevokeSetName: func() IsolatorValue { return &LinuxCapabilitiesRevokeSet{} },
 		LinuxCapabilitiesRetainSetName: func() IsolatorValue { return &LinuxCapabilitiesRetainSet{} },
 		LinuxNoNewPrivilegesName:       func() IsolatorValue { v := LinuxNoNewPrivileges(false); return &v },
+		LinuxSeccompRemoveSetName:      func() IsolatorValue { return &LinuxSeccompRemoveSet{} },
+		LinuxSeccompRetainSetName:      func() IsolatorValue { return &LinuxSeccompRetainSet{} },
 	} {
 		AddIsolatorName(name, LinuxIsolatorNames)
 		AddIsolatorValueConstructor(name, con)
@@ -41,6 +46,15 @@ func init() {
 type LinuxNoNewPrivileges bool
 
 func (l LinuxNoNewPrivileges) AssertValid() error {
+	return nil
+}
+
+// TODO(lucab): both need to be clarified in spec,
+// see https://github.com/appc/spec/issues/625
+func (l LinuxNoNewPrivileges) multipleAllowed() bool {
+	return true
+}
+func (l LinuxNoNewPrivileges) Conflicts() []ACIdentifier {
 	return nil
 }
 
@@ -54,6 +68,10 @@ func (l *LinuxNoNewPrivileges) UnmarshalJSON(b []byte) error {
 	*l = LinuxNoNewPrivileges(v)
 
 	return nil
+}
+
+type AsIsolator interface {
+	AsIsolator() (*Isolator, error)
 }
 
 type LinuxCapabilitiesSet interface {
@@ -75,6 +93,15 @@ func (l linuxCapabilitiesSetBase) AssertValid() error {
 	if len(l.val.Set) == 0 {
 		return errors.New("set must be non-empty")
 	}
+	return nil
+}
+
+// TODO(lucab): both need to be clarified in spec,
+// see https://github.com/appc/spec/issues/625
+func (l linuxCapabilitiesSetBase) multipleAllowed() bool {
+	return true
+}
+func (l linuxCapabilitiesSetBase) Conflicts() []ACIdentifier {
 	return nil
 }
 
@@ -115,17 +142,17 @@ func NewLinuxCapabilitiesRetainSet(caps ...string) (*LinuxCapabilitiesRetainSet,
 	return &l, nil
 }
 
-func (l LinuxCapabilitiesRetainSet) AsIsolator() Isolator {
+func (l LinuxCapabilitiesRetainSet) AsIsolator() (*Isolator, error) {
 	b, err := json.Marshal(l.linuxCapabilitiesSetBase.val)
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 	rm := json.RawMessage(b)
-	return Isolator{
+	return &Isolator{
 		Name:     LinuxCapabilitiesRetainSetName,
 		ValueRaw: &rm,
 		value:    &l,
-	}
+	}, nil
 }
 
 type LinuxCapabilitiesRevokeSet struct {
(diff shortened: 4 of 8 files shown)


actool/manifest.go
```diff
@@ -284,6 +319,45 @@ func patchManifest(im *schema.ImageManifest) error {
...
+			return nil, fmt.Errorf("cannot parse seccomp-mode argument: %s", a)
+		}
+	}
+
+	// Instantiate an Isolator with the content specified by the --seccomp-set parameter.
+	var err error
+	var seccomp types.AsIsolator
+	switch mode {
+	case "remove":
+		seccomp, err = types.NewLinuxSeccompRemoveSet(errno, strings.Split(patchSeccompSet, ",")...)
```
jellonek: Looks like errno could be unset in this place if it's not in cmdline parameters. Is this desired?
Also this looks like errno should be declared as `errno := "EPERM"` around 326 line.

jellonek: Other possibility - command line help could be misinterpreted as me as example shows ;)

jellonek: Ok, docs are more clear in this, so IMO cmdline help is not so clear.

lucab: `errno` is optional, so an empty string is a desired default. Command line help marks it as an optional parameter:
> [--seccomp-mode=remove|retain[,errno=EPERM]]

In which way you don't find it clear? Any way I can improve it to be more immediate to grasp?

spec/ace.md
```diff
@@ -130,8 +130,69 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+* Scope: app
+
+**Parameters:**
+
+* **set** case-sensitive list of syscall names that will be blacklisted (ie. blocked); values starting with `@` MUST be handled as scoped special values (see notes below). All syscalls specified in this set MUST be blocked. This set MAY be augmented by an implementation-specific default blacklist set (see notes below). Syscalls not specified in the union of these two sets MUST NOT be blocked.
+* **errno** all-uppercase name of a single [errno code](http://man7.org/linux/man-pages/man3/errno.3.html) that will be returned by blocked system calls, instead of terminating. If missing or empty, by default blocked syscalls MUST result in app termination via `SIGSYS` signal.
+
+**Notes:**
+ 1. The `os/linux/seccomp-remove-set` isolator cannot be used in conjunction with the `os/linux/seccomp-retain-set` isolator.
```
alban: The spec should say whether having several isolators of type `os/linux/seccomp-remove-set` is allowed. I guess we should not allow that: otherwise, conflicting errno for the same syscalls would make it ambiguous.

Ditto for the retain-set.

actool/manifest.go
```diff
@@ -250,6 +264,23 @@ func patchManifest(im *schema.ImageManifest) error {
 		}
 	}
 
+	if patchSeccompMode != "" {
```
alban: `actool patch-manifest` could return an error when patchSeccompMode == "" but patchSeccompSet != "".

spec/ace.md
lucab: Ah, good point.

actool/manifest.go
jellonek: Sorry for mess. It was unclear for me which errno values could be used in this place. After a break - it's clear for me to comment AFTER reading whole diff...

actool/manifest.go
```diff
@@ -284,6 +319,45 @@ func patchManifest(im *schema.ImageManifest) error {
...
+		kv := strings.Split(a, "=")
+		switch len(kv) {
+		case 1:
+			// mode, either "remove" or "retain"
+			mode = kv[0]
+		case 2:
+			// k=v argument, only "errno" allowed for now
+			if kv[0] == "errno" {
+				errno = kv[1]
+			}
```
lucab: Defensively, sounds like a good idea.

spec/ace.md
```diff
@@ -130,8 +130,69 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+**Notes:**
+ 1. The `os/linux/seccomp-retain-set` isolator cannot be used in conjunction with the `os/linux/seccomp-remove-set` isolator.
+ 2. When an app does not have any seccomp isolator (neither `os/linux/seccomp-remove-set` nor `os/linux/seccomp-retain-set` are specified), implementations MAY apply their own set of blocked syscalls.
+ 3. The implementation-specific default whitelist MUST be the minimum set of syscalls required for app life-cycle. For example, implementations may require the `exit(2)` syscall to be always allowed for clean app termination. This set MAY be empty.
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc/` top-level scope is reserved for usage in this specification. Implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
+  * `@appc/all` represents the set of all available syscalls. In the context of `seccomp-retain-set`, this wildcard masks all other values specified in `set` and can be effectively used to opt-out seccomp filtering.
```
lucab: Two remarks here:
 * I think there should be in the spec some way for users to opt-out. We may reword this as a SHOULD and mention sensible security defaults or side-effects with other isolators, but keep it here.
 * From OOB discussion you are referring to [this systemd snippet](https://github.com/systemd/systemd/blob/master/src/nspawn/nspawn-seccomp.c#L39). It was introduced by [this commit](https://github.com/systemd/systemd/commit/9a71b1122c6e49dd9227f82b2f53837c7ea13019) and I think this is only partially relevant, as it further enforces normal capability behavior via seccomp (ie. directly blocking syscalls with EPERM when missing capabilities). No process termination there, and no visible effects if the app lacks the capability in the first place. I'd consider this an implementation detail of capabilities enforcing, which just happens to use seccomp but shouldn't practically hinder further filters setup.

actool/manifest.go
lucab: I realized I didn't make it clear and my code is making some assumptions that `set` may be empty. On a second thought, I think I am not sure there are actual usecases for this. Should I make it explicit that `set MUST NOT be empty` and just kill this corner case?

spec/ace.md
lucab: Still relevant for the discussion, may be worth a second opinion.

actool/manifest.go
alban: I was referring to the other side: if the set is not empty but the mode is not given (e.g. the user called `actool patch-manifest --seccomp-set=chroot`). In this case, the current code in `actool` would just ignore the seccomp parameter and not add any isolator. I think `actool` should return an error in that case instead of silently doing nothing.

lucab: Ah, fixed that. Opinions regarding the other question?

alban: I'm ok with saying `set MUST NOT be empty`. It's easier to change that in that direction if we ever need to allow empty set later.

lucab: Ack, done.

lucab:
@alban amended as per your suggestions.

Maintainers PTAL, if fine I'd like to see this landing soon.

actool/manifest.go
```diff
@@ -250,6 +264,31 @@ func patchManifest(im *schema.ImageManifest) error {
 		}
 	}
 
+	if patchSeccompMode != "" {
+		i := 0
```
jonboulle: I'm confused by this change; seems like you're being careful to remove any existing isolator by the given name - but we weren't (and still aren't) doing this for the other cases (like `patchRevokeCaps`), instead just appending to the list. 

I suppose it's underspecified in the spec that the list can contain multiple entries of a particular name, and we're arguably just skirting around this today by having GetByName return the last in the list. But can we split this out so that we perform consistently for the new isolator types, and then clear that up in a subsequent change?

lucab: Correct, I made it explicit in the two seccomp isolators, but behavior for others is not specified. At which level do you want to split this out? Something like introducing a `ReplaceIsolatorsByName(Isolator, []string)`?

spec/ace.md
```diff
@@ -130,8 +130,74 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
 
 ### Linux Isolators
 
-These isolators are specific to the Linux kernel and are impossible to represent as a 1-to-1 mapping on other kernels.
-The first example is "capabilities" but this will be expanded to include things such as SELinux, SMACK or AppArmor.
+These isolators are specific to the Linux kernel and are impossible to re wpresent as a 1-to-1 mapping on other kernels.
```
jonboulle: represent

spec/ace.md
```diff
@@ -130,8 +130,74 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+ 3. The implementation-specific default blacklist SHOULD be a security-focused set of syscalls typically not required by apps. For example, implementations may require the `reboot(2)` syscall to be always blocked. This set MAY be empty.
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc/` scope is reserved for usage in this specification, implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
+  * `@appc/empty` represents the empty set. In the context of `seccomp-remove-set`, this wildcard masks all other values specified in `set` and can be effectively used to opt-in implementation-specific seccomp filtering.
+
+
+**Example:**
+```json
+"name": "os/linux/seccomp-remove-set",
+"value": {
+  "errno": "ENOTSUP",
```
jonboulle: could you indent these a little more, similar to other examples?

spec/ace.md
```diff
@@ -130,8 +130,74 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+  "set": [
+    "clock_adjtime",
+    "clock_settime",
+    "reboot"
+  ]
+}
+```
+
+In the example above, the process will not be allowed to invoke `clock_adjtime(2)`, `clock_settime(2)` and `reboot(2)` (and any other syscall in the implementation-specific blacklist). When invoked, such syscalls will immediately return a `ENOTSUP` error code. All other syscalls will behave as usual.
```
jonboulle: nit, [serial comma](https://en.wikipedia.org/wiki/Serial_comma) please

actool/manifest.go
jonboulle: hmm; I don't want to drag you into too much of a yakshave here. you've just unfortunately highlighted this issue that's now going to nag me :-)

Perhaps the minimal change to this diff to appease my pedantry would be to add sections to the other isolators to say that they "may be listed multiple times but only the last will be considered" or something to that effect. Then we're not conflating that behavioural change with the addition of seccomp stuff. 

Then, subsequently, we could have `ReplaceIsolatorsByName` as you suggest, and tighten up the requirements for the other isolators too. WDYT?

spec/ace.md
```diff
@@ -130,8 +130,74 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+* Scope: app
+
+**Parameters:**
+
+* **set** case-sensitive list of syscall names that will be blacklisted (ie. blocked); values starting with `@` MUST be handled as scoped special values (see notes below). This field MUST NOT be empty. All syscalls specified in this set MUST be blocked. This set MAY be augmented by an implementation-specific default blacklist set (see notes below). Syscalls not specified in the union of these two sets MUST NOT be blocked.
+* **errno** all-uppercase name of a single [errno code](http://man7.org/linux/man-pages/man3/errno.3.html) that will be returned by blocked system calls, instead of terminating. If missing or empty, by default blocked syscalls MUST result in app termination via `SIGSYS` signal.
+
+**Notes:**
+ 1. Only a single `os/linux/seccomp-remove-set` isolator can be specified per-app, and it cannot be used in conjunction with the `os/linux/seccomp-retain-set` isolator.
```
jonboulle: Manifest validation should check for these requirements

actool/manifest.go
lucab: I need to rework this anyway: as you tangentially highlighted, I'm replacing just the last isolators in the array instead of all of them. As such, I need something like `ReplaceIsolatorsByName` even if just for seccomp.

I don't think it makes much sense to partially clarify an unspecified behavior now in a half-baked way, just to re-visit and perhaps change it soon later. Plus, there are several other corner-cases which may be worth to address (eg. precedence of multiple isolators in pod+app scopes). Some details have been unspecified so far and I also like to disambiguate them, but I don't agree on rushing to do it here.

jonboulle: I don't think it's half-baked, it's just specifying the current (unfortunate) behaviour. How do you want to move forward here?

lucab: Leave it unspecified for the moment and then clarify a sane behavior once for all. I think somewhere else in the code there are already checks/assumptions about unique isolator entries, and I pretty much prefer tightening the specs instead of codifying accidental behaviors.

jonboulle: OK, can you file a follow up for that and we can get this in?

lucab: Sure!

jonboulle:
@lucab ready for review?

lucab:
@jonboulle rebased once more as travis catched an additional error case. It should be ready for another round now, yes.

schema/types/isolator.go
```diff
@@ -40,16 +44,55 @@ func AddIsolatorName(n ACIdentifier, ns map[ACIdentifier]struct{}) {
...
 
+// assertValid checks that every single isolator is valid and that
+// the whole set is well built
+func (isolators Isolators) assertValid() error {
+	typesMap := make(map[ACIdentifier]uint)
+	for _, i := range isolators {
+		if err := i.Value().AssertValid(); err != nil {
+			return err
+		}
+		if typesMap[i.Name]+1 > i.Value().MaxInstances() {
```
jonboulle: This check returning an error about duplicates presumes that MaxInstances is always 1 - perhaps "isolators set contains too many of type %s" or something

schema/types/isolator.go
```diff
@@ -40,16 +44,55 @@ func AddIsolatorName(n ACIdentifier, ns map[ACIdentifier]struct{}) {
...
+			}
+		}
+		typesMap[i.Name] += 1
+	}
+	return nil
+}
+
 // GetByName returns the last isolator in the list by the given name.
-func (is *Isolators) GetByName(name ACIdentifier) *Isolator {
+func (is *Isolators) GetByName(name ACIdentifier) (*Isolator, int) {
```
jonboulle: Never using this index, just remove it?

schema/pod.go
```diff
@@ -116,6 +116,11 @@ func (al AppList) assertValid() error {
 			return fmt.Errorf(`duplicate apps of name %q`, a.Name)
 		}
 		seen[a.Name] = true
+		if a.App != nil {
```
jonboulle: This shouldn't be necessary - the individual app marshalling should handle it

actool/manifest.go
```diff
@@ -284,6 +323,47 @@ func patchManifest(im *schema.ImageManifest) error {
...
+	// Instantiate an Isolator with the content specified by the --seccomp-set parameter.
+	var err error
+	var seccomp types.AsIsolator
+	switch mode {
+	case "remove":
+		seccomp, err = types.NewLinuxSeccompRemoveSet(errno, strings.Split(patchSeccompSet, ",")...)
+	case "retain":
+		seccomp, err = types.NewLinuxSeccompRetainSet(errno, strings.Split(patchSeccompSet, ",")...)
+	default:
+		err = fmt.Errorf("Unknown seccomp mode %s", mode)
```
jonboulle: nit, error strings start lowercase

actool/manifest.go
```diff
@@ -260,14 +295,18 @@ func patchManifest(im *schema.ImageManifest) error {
...
+			case types.LinuxNoNewPrivilegesName:
 				ok = true
 				kv := strings.Split(is, ",")
 				if len(kv) != 2 {
 					return fmt.Errorf("isolator %s: invalid format", name)
 				}
-
 				isolatorStr = fmt.Sprintf(`{ "name": "%s", "value": %s }`, name, kv[1])
+			case types.LinuxSeccompRemoveSetName:
+				fallthrough
```
jonboulle: nit, for the trivial fallthrough case, grouping cases is slightly more idiomatic
```
case types.LinuxSeccompRemoveSetName, types.LinuxSeccompRetainSetName:
    ok = false
```

actool/manifest.go
```diff
@@ -250,6 +264,27 @@ func patchManifest(im *schema.ImageManifest) error {
...
+		if err != nil {
+			return err
+		}
+		seccompReps := []types.ACIdentifier{types.LinuxSeccompRemoveSetName, types.LinuxSeccompRetainSetName}
+		_, err = app.Isolators.ReplaceIsolatorsByName(*seccompIsolator, seccompReps)
+		if err != nil {
+			return err
+		}
+	} else {
+		if patchSeccompSet != "" {
```
jonboulle: collapse into 282?

schema/pod.go
lucab: It looks like we have a test-case exercising exactly this code-path, and panics here.

schema/types/isolator.go
lucab: Ack, changed to a more neutral wording.

spec/ace.md
lucab: Added.

lucab:
@jonboulle latest revision accounts for your further review (also left some additional comments inline).

actool/manifest.go
```diff
@@ -284,6 +319,47 @@ func patchManifest(im *schema.ImageManifest) error {
 	return nil
 }
 
+// parseSeccompArgs parses seccomp mode and set CLI flags, preparing an
+// appropriate seccomp isolator.
+func parseSeccompArgs(patchSeccompMode string, patchSeccompSet string) (types.AsIsolator, error) {
```
jonboulle: Why does this return AsIsolator and not just Isolator?

schema/pod.go
jonboulle: I don't understand, reference?

schema/types/isolator.go
```diff
@@ -40,6 +44,27 @@ func AddIsolatorName(n ACIdentifier, ns map[ACIdentifier]struct{}) {
...
+// assertValid checks that every single isolator is valid and that
+// the whole set is well built
+func (isolators Isolators) assertValid() error {
+	typesMap := make(map[ACIdentifier]uint)
+	for _, i := range isolators {
+		if err := i.Value().AssertValid(); err != nil {
+			return err
+		}
+		if typesMap[i.Name]+1 > i.Value().MaxInstances() {
+			return fmt.Errorf(`isolators set contains too many instances of type %s"`, i.Name)
```
jonboulle: Check out https://github.com/appc/spec/blob/master/schema/types/errors.go#L26 for an alternative way to go here

schema/types/isolator.go
```diff
@@ -52,6 +77,24 @@ func (is *Isolators) GetByName(name ACIdentifier) *Isolator {
 	return nil
 }
 
+// ReplaceIsolatorsByName overrides matching isolator types with a new
+// isolator, deleting them all and appending the new one instead
+func (is *Isolators) ReplaceIsolatorsByName(newIs Isolator, oldNames []ACIdentifier) (uint, error) {
```
jonboulle: There's still no error condition here

schema/types/isolator.go
```diff
@@ -71,6 +114,8 @@ func (is *Isolators) Unrecognized() Isolators {
 type IsolatorValue interface {
 	UnmarshalJSON(b []byte) error
 	AssertValid() error
+	MaxInstances() uint
```
jonboulle: Please document these

schema/types/isolator_linux_specific.go
```diff
@@ -149,15 +174,148 @@ func NewLinuxCapabilitiesRevokeSet(caps ...string) (*LinuxCapabilitiesRevokeSet,
...
+
+func (l linuxSeccompBase) AssertValid() error {
+	if len(l.val.Set) == 0 {
+		return errors.New("empty set")
+	}
+	if l.val.Errno == "" {
+		return nil
+	}
+	for _, c := range l.val.Errno {
+		if !unicode.IsUpper(c) {
```
jonboulle: Oh, never mind, found it

spec/ace.md
```diff
@@ -131,7 +131,75 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+This set includes Linux-specific isolators, which may relies on kernel technologies like seccomp, SELinux, SMACK or AppArmor.
+
+#### os/linux/seccomp-remove-set
+
+* Scope: app
+
+**Parameters:**
+
+* **set** case-sensitive list of syscall names that will be blacklisted (ie. blocked); values starting with `@` MUST be handled as scoped special values (see notes below). This field MUST NOT be empty. All syscalls specified in this set MUST be blocked. This set MAY be augmented by an implementation-specific default blacklist set (see notes below). Syscalls not specified in the union of these two sets MUST NOT be blocked.
+* **errno** all-uppercase name of a single [errno code](http://man7.org/linux/man-pages/man3/errno.3.html) that will be returned by blocked system calls, instead of terminating. If missing or empty, by default blocked syscalls MUST result in app termination via `SIGSYS` signal.
```
jonboulle: Perhaps specify what an implementation should do if an invalid errno is supplied

schema/types/isolator_linux_specific.go
jonboulle: I'd rather return a more specific error - e.g "Errno must be an upper case string" or something

jonboulle: `if l.val.Errno != strings.ToUpper(l.val.Errno)`

schema/types/isolator_linux_specific.go
```diff
@@ -149,15 +174,148 @@ func NewLinuxCapabilitiesRevokeSet(caps ...string) (*LinuxCapabilitiesRevokeSet,
...
+	val linuxSeccompValue
+}
+
+func (l linuxSeccompBase) MaxInstances() uint {
+	return 1
+}
+
+func (l linuxSeccompBase) AssertValid() error {
+	if len(l.val.Set) == 0 {
+		return errors.New("empty set")
```
jonboulle: for consistency , "set must be non-empty"

spec/ace.md
```diff
@@ -131,7 +131,75 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+**Notes:**
+ 1. Only a single `os/linux/seccomp-remove-set` isolator can be specified per-app, and it cannot be used in conjunction with the `os/linux/seccomp-retain-set` isolator.
+ 2. When an app does not have any seccomp isolator (neither `os/linux/seccomp-remove-set` nor `os/linux/seccomp-retain-set` are specified), implementations MAY apply their own set of blocked syscalls.
```
jonboulle: English nit, prefer "neither .. nor .. is" to "neither .. nor .. are"

spec/ace.md
```diff
@@ -131,7 +131,75 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+ 3. The implementation-specific default blacklist SHOULD be a security-focused set of syscalls typically not required by apps. For example, implementations may require the `reboot(2)` syscall to be always blocked. This set MAY be empty.
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc/` scope is reserved for usage in this specification, implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
```
jonboulle: specification, implementations => specification; implementations

spec/ace.md
```diff
@@ -131,7 +131,75 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+**Notes:**
+ 1. Only a single `os/linux/seccomp-retain-set` isolator can be specified per-app, and it cannot be used in conjunction with the `os/linux/seccomp-remove-set` isolator.
+ 2. When an app does not have any seccomp isolator (neither `os/linux/seccomp-remove-set` nor `os/linux/seccomp-retain-set` are specified), implementations MAY apply their own set of blocked syscalls.
+ 3. The implementation-specific default whitelist MUST be the minimum set of syscalls required for app life-cycle. For example, implementations may require the `exit(2)` syscall to be always allowed for clean app termination. This set MAY be empty.
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc/` top-level scope is reserved for usage in this specification. Implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
+  * `@appc/all` represents the set of all available syscalls. In the context of `seccomp-retain-set`, this wildcard masks all other values specified in `set` and can be effectively used to opt-out seccomp filtering.
```
jonboulle: Please namespace this as "@appc.io/all"

schema/types/isolator_linux_specific.go
lucab: Agreed on the more specific error, but not on the code change (no need to manipulate input before it has been validated, and no need to get into unicode normalization nuances here).

schema/pod.go
lucab: [TestAppList](https://github.com/appc/spec/blob/9bdac40da49bfee6737cd4e267c41c201a60f321/schema/pod_test.go#L39) would [trigger a panic](https://travis-ci.org/appc/spec/jobs/137769458) here.

spec/ace.md
lucab: That would make the field invalid, and the behavior would be the same as for any other invalid field. Do you have any specific idea in mind, ie. why do you want to special case this? (validator is not currently validating set-membership, only the uppercase constraint, as I'm still looking for a way to go string->uintptr without hardcoding the map here)

spec/ace.md
lucab: Done, even though it was my impression (from the rest of the spec) that url-like scopes are only employed in discoverable labels.

schema/types/isolator.go
lucab: Thanks for the hint, I see this is part of what we were discussing today. I think it won't fit with the surrounding style here, but I'll consider it for the future.

spec/ace.md
jonboulle: Hmm, do you have any examples?

spec/ace.md
jonboulle: Well I assumed you didn't want to specify actual fields here because you don't want the spec to have to mirror and chase the upstream set of Linux signals - i.e. the field is basically opaque as far as the spec is concerned. So what should an executor implementation do if a provided signal isn't available on the particular platform?

schema/types/isolator_linux_specific.go
jonboulle: *grumbles about lack of strings.IsUpper*

spec/ace.md
lucab: Moving this to a MUST for POSIX.1 and C99 errnos, and a SHOULD for all the others. This will account for future entries, whenever Linux introduces new custom errno (as the list may be kernel-version dependent).

lucab:
@jonboulle would you mind another round?

schema/pod.go
jonboulle: I think I'm dumb, I'm really not seeing what would cause this panic (why is .App being dereferenced? the go marshaller should skip nil pointers). Can you explain it to me?

lucab: I'm not sure about every single detail, but just by reading the existing code this is my understanding of the flow:
 1. TestAppList directly builds its own [RuntimeApp instances](https://github.com/appc/spec/blob/9bdac40da49bfee6737cd4e267c41c201a60f321/schema/pod_test.go#L44) without any `App` member.
 2. [Validation](https://github.com/lucab/spec/blob/b183ae94f9bdd4c8e0e590ecd2f0ff298fd79967/schema/pod.go#L106) is invoked before marshaling
 3. `App` is `nil` at this point, and [marshaling is called on it](https://github.com/lucab/spec/blob/b183ae94f9bdd4c8e0e590ecd2f0ff298fd79967/schema/pod.go#L120)
 4. panic due to nil pointer dereference

> (why is .App being dereferenced? the go marshaller should skip nil pointers)

This is a validation routine, used for sanity checks both for marshaling and unmarshaling. In the specific marshaling codepath, validation happens *before* marshaling (marshaller will later skip nil pointers, but validator panics on them first).

jonboulle: But look at the diff. Marshalling wasn't being called on it before you added it. I'm saying, why do you need to add this stanza at all?

lucab: This links `AppList` validation to validation of `RuntimeApp` instances contained therein.

But I think I see your point now: I was considering this validation function per-se, while in the bigger context it already makes sense as every layer performs its own validation. Given that this is a private validation method only uses by the marshaller, I'll remove this un-necessary snippet.

lucab:
@jonboulle @alban anything else needing rework here?

schema/types/isolator.go
```diff
@@ -52,6 +77,24 @@ func (is *Isolators) GetByName(name ACIdentifier) *Isolator {
 	return nil
 }
 
+// ReplaceIsolatorsByName overrides matching isolator types with a new
+// isolator, deleting them all and appending the new one instead
+func (is *Isolators) ReplaceIsolatorsByName(newIs Isolator, oldNames []ACIdentifier) uint {
+	var numReplacements uint = 0
```
alban: Why bothering counting numReplacements? the only caller of this function ignore the returned value.

examples/image.json
```diff
@@ -60,6 +60,10 @@
             {
                 "name": "os/linux/no-new-privileges",
                 "value": true
+            },
+            {
+                "name": "os/linux/seccomp-remove-set",
+                "value": {"errno": "EACCESS", "set": ["clock_settime", "clock_adjtime", "reboot"]}
```
alban: could this line be pretty-printed?
```
{
   "name" : "os/linux/seccomp-remove-set",
   "value" : {
      "errno" : "EACCESS",
      "set" : [
         "clock_settime",
         "clock_adjtime",
         "reboot"
      ]
   }
}
```

schema/types/isolator.go
lucab: That's its only way to signal if it actually replaced anything. While the *current* caller doesn't use it, I don't think it's a good idea to remove this observable return value completely.

schema/types/isolator.go
```diff
@@ -69,8 +112,17 @@ func (is *Isolators) Unrecognized() Isolators {
...
 // implement this interface to facilitate unmarshalling and validation.
 type IsolatorValue interface {
+	// UnmarshalJSON unserialize a JSON-encoded isolator
 	UnmarshalJSON(b []byte) error
+	// AssertValid returns a non-nil error value if an IsolatorValue is not valid
+	// according to appc spec
 	AssertValid() error
+	// MaxInstances specifies the maximum number of allowed isolator instances
+	// for this isolator type
+	MaxInstances() uint
```
alban: MaxInstances always returns either `1` or `^uint(0)` (aka `math.MaxUint32`?). I don't see a use case for other values. Could it be `multipleAllowed() bool` instead?

schema/types/isolator_linux_specific.go
```diff
@@ -17,12 +17,15 @@ package types
...
 	"errors"
+	"unicode"
 )
 
 const (
 	LinuxCapabilitiesRetainSetName = "os/linux/capabilities-retain-set"
 	LinuxCapabilitiesRevokeSetName = "os/linux/capabilities-remove-set"
 	LinuxNoNewPrivilegesName       = "os/linux/no-new-privileges"
+	LinuxSeccompRemoveSetName      = "os/linux/seccomp-remove-set"
+	LinuxSeccompRetainSetName      = "os/linux/seccomp-retain-set"
```
alban: So we have Remove and Revoke. Is the plan to migrate to Remove?

schema/types/isolator_linux_specific.go
```diff
@@ -78,6 +95,14 @@ func (l linuxCapabilitiesSetBase) AssertValid() error {
 	return nil
 }
 
+// TODO(lucab): both need to be clarified in spec
+func (l linuxCapabilitiesSetBase) MaxInstances() uint {
+	return ^uint(0)
+}
+func (l linuxCapabilitiesSetBase) Conflicts() []ACIdentifier {
+	return nil
```
alban: https://github.com/appc/spec/blob/master/spec/ace.md#oslinuxcapabilities-retain-set says:

> The retain set cannot be used in conjunction with the remove set.

schema/types/isolator_linux_specific.go
lucab: Yes (also to better align constant name with textual label).

schema/types/isolator_linux_specific.go
lucab: I opened https://github.com/appc/spec/issues/625 to clarify all those points at once, at this seccomp PR was starting to touch way too many things. If it is ok for you, I'd fix that there.

alban: Sure. The TODO comment could include the link to #625.

schema/types/isolator_linux_specific_test.go
```diff
@@ -94,3 +94,113 @@ func TestNewLinuxCapabilitiesRevokeSet(t *testing.T) {
...
 }
+
+func TestNewLinuxSeccompRemoveSet(t *testing.T) {
+	tests := []struct {
+		set   []string
+		errno string
+
+		wset   []LinuxSeccompEntry
+		werrno LinuxSeccompErrno
+		werr   bool
```
alban: What does "w" stand for?

spec/ace.md
```diff
@@ -131,7 +131,75 @@ An executor MAY implement a "strict mode" where an image cannot run unless all i
...
+
+**Notes:**
+ 1. Only a single `os/linux/seccomp-remove-set` isolator can be specified per-app, and it cannot be used in conjunction with the `os/linux/seccomp-retain-set` isolator.
+ 2. When an app does not have any seccomp isolator (neither `os/linux/seccomp-remove-set` nor `os/linux/seccomp-retain-set` is specified), implementations MAY apply their own set of blocked syscalls.
+ 3. The implementation-specific default blacklist SHOULD be a security-focused set of syscalls typically not required by apps. For example, implementations may require the `reboot(2)` syscall to be always blocked. This set MAY be empty.
+ 4. Values starting with `@` identify special wildcards and are scoped by the `/` separator. The `@appc.io/` scope is reserved for usage in this specification; implementations MAY provide additional wildcards in their own namespace (eg. `@implementation/wildcard`). The following special values are currently defined:
+  * `@appc.io/empty` represents the empty set. In the context of `seccomp-remove-set`, this wildcard masks all other values specified in `set` and can be effectively used to opt-in implementation-specific seccomp filtering.
+
+
```
alban: 2 empty lines

schema/types/isolator.go
lucab: > (aka math.MaxUint32?)

No, value depends on arch-specific uint size.

> I don't see a use case for other values. Could it be multipleAllowed() bool instead?

To be honest I don't have a usecase to propose, but I (generally) prefer to sketch those interfaces with an open/close approach. No strong opinions here, I can rework it if you find it prettier. Thoughts?

schema/types/isolator_linux_specific_test.go
lucab: To me it stands for "wanted" (as in got/want everywhere else), but I was just following the convention of the surrounding tests (so it may be something different).

jonboulle: yep, that's what it is

jonboulle: would be clearer if the messages used "wanted" not "expected" :p

lucab: I have to rebase this PR once more, so let me just change variable names to something more readable.

schema/types/isolator.go
jonboulle: I don't feel particularly strongly about this, but generally speaking in Go we err on the side of minimal possible interfaces (including for function prototypes) until something more complex is demonstrably necessary - since typically dependencies are vendored and builds are static, so refactorings are straightforward. 

Similar logic applies when considering whether to export methods (e.g. in this case `maxInstances`/`multipleAllowed` can probably remain package-local for now)

https://github.com/appc/spec/pull/621/files#r67835328 is similar

jonboulle:
A couple small open questions around style but generally this LGTM. @lucab let me know if you want to change those or just move forward with this as-is.

schema/types/isolator.go
lucab: I see. I'll try to align myself here. Let me quickly adjust all those.

schema/types/isolator_linux_specific_test.go
```diff
@@ -94,3 +94,113 @@ func TestNewLinuxCapabilitiesRevokeSet(t *testing.T) {
 	}
 
 }
+
+func TestNewLinuxSeccompRemoveSet(t *testing.T) {
+	tests := []struct {
+		set   []string
+		errno string
+
+		expectedSet   []LinuxSeccompEntry
```
jonboulle: you went in the wrong direction here - the rest of the file is pretty consistent with w/want :-(

jonboulle:
LGTM

PR #622, opened 2016-06-03
[WIP] types/volume: add recursive flag

steveeJ:
- [x] Add Flag
- [ ] Decide on default value
- [ ] Tests

### DIFF
--- a/schema/types/volume.go
+++ b/schema/types/volume.go
@@ -40,8 +40,9 @@ type Volume struct {
 
 	// currently used only by "host"
 	// TODO(jonboulle): factor out?
-	Source   string `json:"source,omitempty"`
-	ReadOnly *bool  `json:"readOnly,omitempty"`
+	Source    string `json:"source,omitempty"`
+	ReadOnly  *bool  `json:"readOnly,omitempty"`
+	Recursive *bool  `json:"recursive,omitempty"`
 
 	// currently used only by "empty"
 	Mode *string `json:"mode,omitempty"`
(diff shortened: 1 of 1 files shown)


iaguis:
It seems Docker does [recursive by default](https://github.com/docker/docker/blob/v1.11.2/daemon/oci_linux.go#L504)

steveeJ:
@appc/owners-team opinions?

alban:
systemd-nspawn has [options --bind= --bind-ro=](https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html#--bind=) and since systemd-v226, they are recursive by default, unless "norbind" is specified. See also https://github.com/coreos/rkt/issues/1358 and https://github.com/systemd/systemd/pull/649.

rkt with systemd-nspawn flavors don't specify norbind or rbind, so the volumes are bind-mounted recursively volumes as well.

If we take this PR, I suggest that recursive should be the default in appc/spec as well.

alban:
Actually, I changed my mind: if the recursive boolean is not given in the pod manifest, can we say in the spec that implementations can choose what should be the default? The reasons not to say in the spec what should be the default are listed by @steveeJ in https://github.com/coreos/rkt/issues/2567#issuecomment-223600490:
- normal pods should usually use recursive bind mounts for security, to avoid revealing directories normally hidden by mounts.
- rkt-fly should use non-recursive bind mounts to reduce bind-mount explosions

alban:
TODO:
- [ ] The change in "volume" has to be documented in the [pod manifest schema](https://github.com/appc/spec/blob/master/spec/pods.md#pod-manifest-schema)

jonboulle:
This change is missing a corresponding update to the spec text regarding volumes..

jonboulle:
Could you turn this into a non-WIP proposal please?

alban:
I am working on https://github.com/appc/spec/pull/630

steveeJ:
Thanks for picking this one up @alban! Once #630 is merged this will be automatically closed.

PR #626, opened 2016-06-15
schema/app: add parameter to set support of sd_notify()

alepuccetti:
supportNotify exposes if or not an application supports notifications
using sd_notify(). This information may be used to signal the init
process when a service is ready.

### DIFF
--- a/examples/image.json
+++ b/examples/image.json
@@ -128,6 +128,10 @@
         {
             "name": "homepage",
             "value": "https://example.com"
+        },
+        {
+            "name": "appc.io/executor/supports-systemd-notify",
+            "value": "false"
         }
     ]
 }
--- a/schema/types/annotations_test.go
+++ b/schema/types/annotations_test.go
@@ -79,6 +79,12 @@ func TestAnnotationsAssertValid(t *testing.T) {
 			},
 			false,
 		},
+		{
+			[]Annotation{
+				makeAnno("appc.io/executor/supports-systemd-notify", "false"),
+			},
+			false,
+		},
 		// empty is OK
 		{
 			[]Annotation{},
--- a/spec/aci.md
+++ b/spec/aci.md
@@ -211,6 +211,10 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
         {
             "name": "homepage",
             "value": "https://example.com"
+        },
+        {
+            "name": "appc.io/executor/supports-systemd-notify",
+            "value": false
         }
     ]
 }
@@ -250,6 +254,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
     * **authors** contact details of the people or organization responsible for the image (freeform string)
     * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
+    * **appc.io/executor/supports-systemd-notify** (boolean, optional, defaults to "false" if unset) if set to true, the application SHOULD use the sd\_notify mechanism to signal when it is ready. Also it SHOULD be able to detect if the executor had not set up the sd\_notify mechanism and skip the notification without error ([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) from libsystemd does that automatically).
 
 #### Dependency Matching
 


spec/aci.md
```diff
@@ -239,6 +240,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
         * **count** (integer, optional, defaults to 1 if unset) specifies a range of ports, starting with "port" and ending with "port" + "count" - 1. Must be >=1.
         * **socketActivated** (boolean, optional, defaults to "false" if unsupplied) if set to true, the application expects to be [socket activated](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) on these ports. To perform socket activation, the ACE MUST pass file descriptors using the [socket activation protocol](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) that are listening on these ports when starting this app. If multiple apps in the same pod are using socket activation then the ACE must match the sockets to the correct apps using getsockopt() and getsockname().
+    * **supportNotify** (boolean, optional, defaults to "false" if unset) if set to true, the application should use [sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) to signal when it is ready. Setting this parameter to true for an application that does not send a ready signal with `sd_notify()` may result in misbehave.
```
jonboulle: "may result in misbehave [sic]" is too vague - let's define a policy here. 
What MUST executors do if an application sets this parameter to true but does not send a ready signal?
Also, what SHOULD executors do if an application sets this parameter to true but the executor does not support sd_notify?

jonboulle: Nit, prefer "supportsNotify" to "supportNotify"

alepuccetti: > What MUST executors do if an application sets this parameter to true but does not send a ready signal?
Also, what SHOULD executors do if an application sets this parameter to true but the executor does not support sd_notify?

In both case ma answer is that this should not happen, because setting to true `supportNotify` means that the app will be responsible for the notification. For example, in systemd, if the init system do not get the ready message after a timeout will fill the app.

I could write something like: "Setting this parameter to true for an application that does not send a ready signal with `sd_notify()` may result in starvation or termination of the app. When setting to true the app will be responsible to provide the minimum notifications to the init system"

euank: perhaps it should be rephrased as "if set to true the application MUST use sd_notify" to more clearly express that the alternate behaviour is not well defined.

I think the second case can happen more easily. For example, with `stage1-fly` there is not always a systemd to notify. Should `stage1-fly` refuse to run an application with `supportsNotify=true`, or should `stage1-fly` ignore it and the application be able to detect that it's not running under systemd and run normally?

alepuccetti: On Friday, June 17, 2016, Euan Kemp <notifications@github.com> wrote:

> In spec/aci.md
> <https://github.com/appc/spec/pull/626#discussion_r67568121>:
>
> > @@ -239,6 +240,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
> >          * **port** (integer, required) port number that will be used; see also **count**. Must be >=1 and <=65535.
> >          * **count** (integer, optional, defaults to 1 if unset) specifies a range of ports, starting with "port" and ending with "port" + "count" - 1. Must be >=1.
> >          * **socketActivated** (boolean, optional, defaults to "false" if unsupplied) if set to true, the application expects to be [socket activated](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) on these ports. To perform socket activation, the ACE MUST pass file descriptors using the [socket activation protocol](http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) that are listening on these ports when starting this app. If multiple apps in the same pod are using socket activation then the ACE must match the sockets to the correct apps using getsockopt() and getsockname().
> > +    * **supportNotify** (boolean, optional, defaults to "false" if unset) if set to true, the application should use [sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) to signal when it is ready. Setting this parameter to true for an application that does not send a ready signal with `sd_notify()` may result in misbehave.
>
> perhaps it should be rephrased as "if set to true the application MUST use
> sd_notify" to more clearly express that the alternate behaviour is not well
> defined.
>
This sound good.

> I think the second case can happen more easily. For example, with
> stage1-fly there is not always a systemd to notify. Should stage1-fly
> refuse to run an application with supportsNotify=true, or should
> stage1-fly ignore it and the application be able to detect that it's not
> running under systemd and run normally?
>
I don't k ow how stage1-fly works, but for the moment the manly reason to
support notification is to have a a clear signal when the service in
the pod is ready. This allow to sort out dependencies between pods not
easily.

euank: stage1-fly, unlike stage1-coreos, does not use systemd (nor any init process at all) but instead naively execs the entrypoint of the application.

I think the options for an ACE that doesn't use systemd upon seeing this flag are: 

1. error out that it will not be able to support running this app, or
2. proceed to run the app as normal, letting notify calls fail (which practically all programs using sd_notify behave well under).

euank:
Since this option is somewhat tied to implementation (systemd), I'm not sure if this should be part of the spec or part of specific ACEs (e.g. rkt). The spec also doesn't clearly define that an ACE must provide a means to do things like show the status of an application nor talk about state transitions, which I think would also be helpful to have if this is to be specified.

The alternative is to support it in rkt, and any other ACE that desires, as metadata that lives outside of the ACI (e.g. as an additional flag or via interpreting some well-known annotation).

Other opinions would be quite welcome :)

jonboulle:
@euank my thinking was it would be reasonable to have this in the spec because we already have a precedent with `socketActivated`, but I admit that's arguably a lazy rationale, and I do see the argument it should be per implementation (shame on me for not considering annotations for this).

I'm still noodling on the best way to move forward..

alepuccetti:
@jonboulle and @euank Honestly I do not know other init system and if they implement notification also if stage1_fly do not use any init system this option do no have any effect on it. Again I am not familiar with stage1_fly or other init system. I am working on a patch for rkt to allow services inside a pod to notify when they are ready to better handle inter-pod dependencies. I do not have a strong opinion where expose this features in appc/spec.

spec/aci.md
alban: sd_notify() does not fail and just return zero when the executor has not set up the notification socket (i.e. $NOTIFY_SOCKET is undefined).

So the stage1-fly executor can just ignore that field and the app must deal with it.

What about the following wording:

> * **supportsNotify** (boolean, optional, defaults to "false" if unset) if set to true, the application MUST use [the sd_notify mechanism](https://www.freedesktop.org/software/systemd/man/sd_notify.html) to signal when it is ready. Executors MAY ignore **supportNotify**. An application with supportsNotify=true MUST be able to detect if the executor had not set up the sd_notify mechanism and skip the notification without error ([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) from libsystemd does that automatically).

jonboulle:
any other opinions - @vbatts ?

vbatts:
notifying the init is an interesting concept, but to my knowledge is only something implemented by systemd.
While we did already add the socket activation as a first class field, perhaps this would be a good place to start deferring more to be an annotation. Especially if the executor can ignore it.

alepuccetti:
Ok then, if we put it in annotations we can describe it a little better.  We can define a **systemd** "class" of annotations and so `supportsNotify` become `systemdNotifySupport`. The convention will be to prefix with the string "systemd" all the systemd related parameters. This could be extended to other init systems or other tool in general. I think it is a good convention to allow easy extensibility.

@vbatts what do you think about it? Should we also push `socketActivated` in annotations? In a different PR of course.

jonboulle:
For backwards compatibility we can't really remove socketActivated, but we could add it as an annotation as well (and just fall back to checking it if the standard one isn't set). 

Perhaps something like `appc.io/executor/supportsSystemdNotify` and `appc.io/executor/socketActivated` - IMO the "class" (namespace) is executor features, rather than systemd per se

vbatts:
there may need to be a one-off for `appc.io/executor/socketActivated` of ensuring if set then both the annotation and the field are set and matching. As a backward-compat thing.

alepuccetti:
@jonboulle 
> For backwards compatibility we can't really remove socketActivated, but we could add it as an annotation as well (and just fall back to checking it if the standard one isn't set).

We could have both and giev priority to the annotation if set and then remove the first class field in the future.

alepuccetti:
> Perhaps something like appc.io/executor/supportsSystemdNotify and appc.io/executor/socketActivated - IMO the "class" (namespace) is executor features, rather than systemd per se

Do you mean to have a first class field named "executor"?

jonboulle:
@alepuccetti No, I'm saying having well-known annotations with those names; since annotations are free-form/arbitrary, we're just using a DNS-like name to provide some kind of namespacing

jonboulle:
e.g. `{"annotations": [ "name": "appc.io/executor/socketActivated", "value": "true" ] }`

alepuccetti:
>No, I'm saying having well-known annotations with those names; since annotations are free-form/arbitrary, we're just using a DNS-like name to provide some kind of namespacing

@jonboulle Gotcha, I like that. So for the notification: `{"annotations": [ {"name": "appc.io/executor/supportsSystemdNotify"}]}`

I will make the change for `supportsSystemdNotify` and then submit an other PR for `socketActivated`

alepuccetti:
@jonboulle I will use `"appc.io/executor/systemd/support-notify"` because I get this error when I use uppercases. `panic: ACIdentifier must contain only lower case alphanumeric characters plus "-._~/"`

alepuccetti:
I just noticed that annotations may be applied to all the apps, but this is a per app feature.

alepuccetti:
@jonboulle 
> e.g. {"annotations": [ "name": "appc.io/executor/support-notify", "value": "true" ] }

Annotations refer to the pod, but this feature is per app. So If we have in the same pod  an app supporting notifications and a second one that it does not the pod will be never marked as `active` due to the fact the pod's systemd is waiting for all the apps to send the ready notification.

One solution it would be to document that you should set this annotation to `true` ONLY IF  all the apps in the pod supports notification. If you agree I will update the PR so we can get this thing done.

alban:
@alepuccetti 

> Annotations refer to the pod, but this feature is per app.

Annotations can be added both in the pod manifest (as defined [here](https://github.com/appc/spec/blob/master/spec/pods.md)) and in the app manifest (as defined [here](https://github.com/appc/spec/blob/master/spec/aci.md)).

As you said, the sd_notify feature is per app. Your patch defines "supportsSystemdNotify" in the app manifest and not in the pod manifest, so this looks correct to me.

alepuccetti:
@jonboulle If there is nothing more about this, can we merge this?

examples/image.json
```diff
@@ -128,6 +128,10 @@
         {
             "name": "homepage",
             "value": "https://example.com"
+        },
+        {
+            "name": "appc.io/executor/systemd/support-notify",
```
jonboulle: The name is not correct here

schema/types/annotations_test.go
```diff
@@ -79,6 +79,12 @@ func TestAnnotationsAssertValid(t *testing.T) {
 			},
 			false,
 		},
+		{
+			[]Annotation{
+				makeAnno("appc.io/executor/systemd/support-notify", "false"),
```
jonboulle: Likewise

examples/image.json
alepuccetti: This made me think that it could be a good choice to use `appc.io/executor/systemd/` as prefix for all the systemd related annotation. I will change the docs accordingly.

alepuccetti:
Patch updated

spec/aci.md
```diff
@@ -169,7 +169,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
                 "count": 1000,
                 "protocol": "tcp"
             }
-        ]
+        ],
```
euank: the comma makes this invalid json

jonboulle:
Two nits:
i) can you fix the name in the commit message - it's different from the actual ones now
ii) since this is an annotation I'm a little wary of the MUST, can it be SHOULD instead?

After that, this LGTM.

Note since this is not changing the spec itself, it can be implemented in rkt without being blocked by this.

jonboulle:
LGTM. Thanks!

PR #629, opened 2016-06-21
discovery: add port parameters

dgonyeo:
Currently the ACI spec only supports discovery on ports 443 and 80, this
does not allow using a discovery server on any other ports.

This commit adds an argument allowing for discovery on arbitrary ports.
If the argument is not specified (as in, is 0), the default ports are
used.

Replaces https://github.com/appc/spec/pull/110#issuecomment-227428866.
Prompted by https://github.com/appc/acpush/issues/6.

### DIFF
--- a/actool/discover.go
+++ b/actool/discover.go
@@ -29,9 +29,10 @@ var (
 		Name:        "discover",
 		Description: "Discover the download URLs for an app",
 		Summary:     "Discover the download URLs for one or more app container images",
-		Usage:       "[--json] APP...",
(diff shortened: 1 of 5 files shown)


jonboulle:
needs a rebase?

dgonyeo:
Don't know how I messed that one up. Should build now.

jonboulle:
LGTM

iaguis:
I guess we can finally close https://github.com/appc/spec/pull/110 :grin:

PR #631, opened 2016-06-23
*: appc v0.8.5

jonboulle:


jonboulle:
apparently some bugs in isolatorvalues, stand by

lucab:
@jonboulle it turned out this may not be a real bug in validator, but rkt misusing it. See https://github.com/appc/spec/pull/633#issuecomment-228107843 for reference, but it only adds a graceful handler to an error case.

jonboulle:
@lucab updated

euank:
LGTM

PR #633, opened 2016-06-23
schema/types: don't panic validating incomplete isolators

lucab:
Even though Isolator instances should be constructed via
unmarshaling, library users may still construct one directly
with a nil value.
This commit fixes the validator to handle this error gracefully
instead of panicing.

### DIFF
--- a/schema/types/isolator.go
+++ b/schema/types/isolator.go
@@ -23,7 +23,12 @@ import (
 var (
 	isolatorMap map[ACIdentifier]IsolatorValueConstructor
 
+	// ErrIncompatibleIsolator is returned whenever an Isolators set contains
+	// conflicting IsolatorValue instances
 	ErrIncompatibleIsolator = errors.New("isolators set contains incompatible types")
+	// ErrInvalidIsolator is returned upon validation failures due to improper
+	// or partially constructed Isolator instances (eg. from incomplete direct construction)
+	ErrInvalidIsolator = errors.New("invalid isolator")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


lucab:
It turns out rkt has [some tests](https://github.com/coreos/rkt/blob/master/tests/rkt_run_pod_manifest_test.go#L776) doing exactly what godoc [says not to do](https://github.com/appc/spec/blob/master/schema/types/isolator.go#L133:L138). I'm going to fix those tests, but this makes validation a bit more graceful. I'm not sure if there is an easy way to forbid this at all without making the type hierarchy even more complex.

euank:
LGTM. I have no problem with making it fail more gracefully in this case of user-error.

jonboulle:
Ideally ValueRaw would not be exported at all, but we're hamstrung there by Go's json marshalling rules

schema/types/isolator.go
```diff
@@ -24,6 +24,7 @@ var (
 	isolatorMap map[ACIdentifier]IsolatorValueConstructor
 
 	ErrIncompatibleIsolator = errors.New("isolators set contains incompatible types")
+	ErrInvalidIsolator      = errors.New("invalid isolator")
```
jonboulle: Could you add a docstring for this error? hinting that it might be thrown if an Isolator is constructed incorrectly for example

lucab: Yup, the one above too? In general I haven't seen lot of docstring around error constants, but I will happy to have more.

jonboulle: Yeah - in this case in particular I think it helps because as a consumer of the package it's unclear why I'd ever encounter this (it's unfortunate that we allow them to even get into this situation)

lucab:
Amended

schema/types/isolator.go
```diff
@@ -23,7 +23,12 @@ import (
 var (
 	isolatorMap map[ACIdentifier]IsolatorValueConstructor
 
+	// ErrIncompatibleIsolator is thrown whenever an Isolators set contains
+	// conflicting IsolatorValue instances
 	ErrIncompatibleIsolator = errors.New("isolators set contains incompatible types")
+	// ErrInvalidIsolator is thrown upon validation failures due to improper
+	// or partially constructed Isolator instances (ie. not constructed via unmarshaling)
```
jonboulle: The ie is wrong (since valid isolators can be constructed from the New* functions). 

Also errors are returned, not thrown.

lucab: Uhm, right.

PR #634, opened 2016-07-05
Added amd64 as a valid architecture on darwin

mbhinder:
Added "amd64" as a valid OS architecture for darwin.

Based on the output of `go env`, I believe this should be a valid architecture for darwin

```
go env | grep GOARCH

GOARCH="amd64"
```

My OS X version:

```
sw_vers
ProductName:    Mac OS X
ProductVersion: 10.10.5
BuildVersion:   14F1808
```


lucab:
I think this is already defined as "darwin/x86_64". That label applies to ACI images, and is unrelated to whatever label golang (or any other language/tool) uses for its own purposes.

PR #639, opened 2016-07-13
schema: add ppc64le/ppc64 as valid linux archs

lucab:
This adds labeling for the two endian variants of ppc64:
`ppc64le` for little-endian and `ppc64` for big-endian.

These explicitly correspond to `uname -m` on Linux systems.

### DIFF
--- a/schema/types/labels.go
+++ b/schema/types/labels.go
@@ -21,7 +21,7 @@ import (
 )
 
 var ValidOSArch = map[string][]string{
-	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv6l", "armv7l", "armv7b"},
+	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv6l", "armv7l", "armv7b", "ppc64", "ppc64le"},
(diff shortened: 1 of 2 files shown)


s-urbaniak:
lgtm

glevand:
Two wrongs don't make a right, so I would say use ppc64 and ppc64le.

lucab:
For reference, this is how they are labeled in other places:
 * Debian: ppc64, ppc64el
 * GOARCH: ppc64, ppc64le
 * uname -m: ppc64, ppc64le

Bikeshedding level, may need a tie-breaker. @jonboulle?

jonboulle:
I think we should go with @glevand 's suggestion/`uname -m` in this case; we do want to be consistent with external sources where appropriate (and arguably `ppc64`/`ppc64el` is consistent with appc's `aarch64`/`aarch64_be` anyway - one "normal", one with endianness specifically called out). We went with the weird arm7 labels at the recommendation of ARM engineers.

lucab:
Updated to use ppc64le and ppc64.

jonboulle:
LGTM

PR #642, opened 2016-07-22
Expand ACI acronym so that it's clear

polettix:
The "ACI" acronym starts to be relevant since the very beginning, but it's not defined anywhere in the starting page. I think it's important to have its expansion early to improve clarity.


dgonyeo:
LGTM

iaguis:
LGTM

PR #644, opened 2016-08-02
discovery: increase default connection timeout

lucab:
This commit increases default connection timeout from 5s to 20s,
in order to gracefully accomodate for retrials. This will help in
overcoming soft-failure cases, like skipping a single faulty
resolvers in resolv.conf.

This is still a quite aggressive timeout, well below the Linux
timeout of 127s for new TCP connections.

### DIFF
--- a/discovery/http.go
+++ b/discovery/http.go
@@ -28,7 +28,7 @@ import (
 type InsecureOption int
 
 const (
-	defaultDialTimeout = 5 * time.Second
+	defaultDialTimeout = 20 * time.Second
 )
 
 const (


lucab:
Coming from https://github.com/coreos/rkt/issues/3009#issuecomment-236852933

There is still some room for increase here, but I'm not sure about why the original value was picked so low (and #112 lacks a bit of context). For reference, nginx has [30s timeout just for DNS resolution](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver_timeout) while default Linux `tcp_syn_retries` result in [127s timeout for new connections](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt).

iaguis:
This looks reasonable to me

lucab:
@jonboulle may have some insights here.

jonboulle:
LRTM

PR #647, opened 2016-08-17
spec: add os/unix/sysctl isolator

### LINKED ISSUE #645: ace: add sysctl isolator
Applications running inside pods may need to tweak several namespaced sysctl parameters. There is a need for executor to support this (https://github.com/coreos/rkt/issues/2694), and OCI Linux runtime manifest already features a sysctl stanza (https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#sysctl).

In short, ACE should provide an `os/linux/sysctl` isolator.

lucab:
This PR introduces an "os/unix/sysctl" isolator, which applies at pod-level.

Closes #645

---

Arbitrary/controversial choices made here:
 * isolator type, as it applies inside pod-context (like other isolators) instead of the host (like top-level stanzas: volumes, ports)
 * ~~new "environment" isolator category, as it can also accommodate "ulimit" ones~~

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -383,6 +383,36 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
 
 **NOTE**: Network limits MUST NOT apply to localhost communication between apps in a pod.
 
+### Unix Isolators
+
+These isolators take care of configuring several settings that are typically available in UNIX-like environments.
+This set includes isolators to tweak some standard POSIX features as well as other technologies that have been ported across multiple kernel families.
+
+#### os/unix/sysctl
+
+Sysctl parameters allow an application to configure kernel level options for its environment.
+A number of these values are namespace-aware as well, which makes them a natural fit for for an ACE to configure.
+
+* Scope: pod
+
+**Parameters:**
+
+* dictionary of `sysctl(8)` key-value entries to set kernel runtime parameters.
+
+```json
+"name": "os/unix/sysctl",
+"value": {
+    "net.ipv4.tcp_mtu_probing": "1",
+    "net.ipv4.tcp_keepalive_time": "300",
+    "net.ipv4.tcp_keepalive_probes": "5",
+    "net.ipv4.tcp_keepalive_intvl": "15"
+}
+```
+**Notes**:
+ 1. Only a single `os/unix/sysctl` isolator can be specified per-pod.
+ 2. Implementations SHOULD validate keys before applying sysctl parameters.
+    For example, a Linux implementation may want to only allow entries related to a specific namespaced subtree like `net.*`.
+    In such case, implementations MUST either ignore forbidden parameters or refuse to run the pod alltogether.
 
 ## App Container Metadata Service
 
--- a/examples/pod_runtime.json
+++ b/examples/pod_runtime.json
@@ -98,6 +98,12 @@
         {
             "name": "resource/memory",
             "value": {"limit": "4G"}
+        },
+        {
+            "name": "os/unix/sysctl",
+            "value": {
+                "net.ipv4.tcp_fin_timeout": "6"
+            }
         }
     ],
     "annotations": [
--- a/schema/types/isolator_test.go
+++ b/schema/types/isolator_test.go
@@ -172,6 +172,27 @@ func TestIsolatorUnmarshal(t *testing.T) {
 			}`,
 			true,
 		},
+		{
+			`{
+				"name": "os/unix/sysctl",
+				"value": {"net.ipv4.tcp_rfc1337": "1"}
+			}`,
+			false,
+		},
+		{
+			`{
+				"name": "os/unix/sysctl",
+				"value": {"net.ipv4.tcp_rfc1337": 1}
+			}`,
+			true,
+		},
+		{
+			`{
+				"name": "os/unix/sysctl",
+				"value": {["net.ipv4.tcp_rfc1337"]}
+			}`,
+			true,
+		},
 	}
 
 	for i, tt := range tests {
@@ -205,6 +226,10 @@ func TestIsolatorsGetByName(t *testing.T) {
 			{
 				"name": "os/linux/no-new-privileges",
 				"value": true
+			},
+			{
+				"name": "os/unix/sysctl",
+				"value": {"net.ipv4.tcp_rfc1337": "1"}
 			}
 		]
 	`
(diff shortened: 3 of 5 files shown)


spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
 
 **NOTE**: Network limits MUST NOT apply to localhost communication between apps in a pod.
 
+### Environment Isolators
+
+An application or a group of applications may often require some tweaking to the environment where they run, in order to perform in an optimal way.
```
euank: Awkward wording I think. I'll suggest "An application may want or need some changes to the system environment it is run in."

At the least, nix often.

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
...
+Environment isolators take care of setting up the runtime environment as needed.
+
+#### environment/sysctl
+
```
euank: Paragraph explaining what sysctls roughly are and that some are namespaced / attached to certain namespaces.

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
...
+  "sysctl": {
+    "net.ipv4.tcp_mtu_probing": "1",
+    "net.ipv4.tcp_keepalive_time": "300",
+    "net.ipv4.tcp_keepalive_probes": "5",
+    "net.ipv4.tcp_keepalive_intvl": "15",
+  }
+}
+```
+**Notes**:
+ 1. Only a single `environment/sysctl` isolator can be specified per-pod.
```
euank: Why not merge?

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
...
+ 2. Implementations MAY sanity-check keys before applying sysctl parameters. For example, only entries related to isolated subsystems (e.g. `net.*` on Linux) may be allowed.
```
euank: Perhaps we should have:

3. Implementations SHOULD only allow sysctl parameters that are safely namespaced and will not impact other pods.

euank:
I tossed on some nits, but I think this makes sense, both in terms of naming and as an isolator.

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
 
 **NOTE**: Network limits MUST NOT apply to localhost communication between apps in a pod.
 
+### Environment Isolators
+
+An application or a group of applications may often require some tweaking to the environment where they run, in order to perform in an optimal way.
+Environment isolators take care of setting up the runtime environment as needed.
+
+#### environment/sysctl
```
jonboulle: Do we want to namespace this more specifically to Linux or *nix? /cc @mpasternacki if he has any thoughts from the BSD perspective

spec/ace.md
jonboulle: maybe "For example, a Linux implementation might only allow entries related to a specific subtree like `net.*`"

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
...
+* dictionary of `sysctl(8)` key-value entries to set kernel runtime parameters.
+
+```json
+"name": "environment/sysctl",
+"value": {
+  "sysctl": {
+    "net.ipv4.tcp_mtu_probing": "1",
+    "net.ipv4.tcp_keepalive_time": "300",
+    "net.ipv4.tcp_keepalive_probes": "5",
+    "net.ipv4.tcp_keepalive_intvl": "15",
```
jonboulle: trailing ,

spec/ace.md
lucab: I checked before venturing here, and at least FreeBSD has sysctl tuning support: https://www.freebsd.org/doc/handbook/configtuning-sysctl.html

jonboulle: Right, I'm just wondering if that suggests we could make this e.g. `os/unix/environment/sysctl`

spec/ace.md
sttts: "will not impact other pods" is a very fuzzy term (compare https://github.com/kubernetes/kubernetes/pull/26057/files#diff-8dc23ab258695ee42154d4d1238c36efR323). Being namespaced in certain namespaces is much clearer defined (might differ by kernel version though).

Also note, that being namespaced of course depends on which namespaces are actually used (ipc, net, etc.) and not shared with the host. This should have influence on the allowed sysctls.

sttts: What about the error behavior after validation? Should a container refuse to launch?

spec/ace.md
lucab: Ack.

spec/ace.md
lucab: I agree this is missing some introduction text here. And thanks for the suggestion!

spec/ace.md
lucab: Mostly because KISS and because merging behavior requires some kind of arbitrary priority ordering.

spec/ace.md
lucab: Summing up a bit of discussion, it is desirable for this to be a SHOULD but without getting into specifying low-level stuff. Consumers are also interested in different behaviors after validation (apply anyway, ignore, fail-hard), but this is something that shouldn't be decided by spec, and left to runtime. As such, the middle-ground would be:
```
Implementations SHOULD validate keys before applying sysctl parameters. 
For example, a Linux implementation might only allow entries related to a 
specific namespaced subtree like `net.*`. In such case, implementations MUST
either ignore forbidden parameters or refuse to run the pod alltogether.
```

spec/ace.md
lucab: I'm more or less neutral here and your scoping also makes sense (maybe dropping the `environment` part). Just for future proofing we may have a `posix-limits` and a `linux-limits` upcoming after this, how would you scope those too in you mind?

spec/ace.md
```diff
@@ -383,6 +383,33 @@ Small quantities can be represented directly as decimals (e.g., 0.3), or using m
...
+* Scope: pod
+
+**Parameters:**
+
+* dictionary of `sysctl(8)` key-value entries to set kernel runtime parameters.
+
+```json
+"name": "environment/sysctl",
+"value": {
+  "sysctl": {
```
lucab: Heads-up: spurious label, this was not supposed to be here.

jonboulle:
bump - this still needs a rework right?

lucab:
It does, plus a decision on https://github.com/appc/spec/pull/647#discussion_r75478277

spec/ace.md
jonboulle: @lucab I would have expected them to be `os/linux/limits` and `os/posix/limits` or similar, per the existing ones.

lucab: Ack, I guess we can settle on `os/unix/sysctl` - `os/unix/limits` - `os/linux/limits`. This is aligned with current naming trend.

lucab:
Second iteration of this: renamed isolator to `os/unix/sysctl` and reworded the key validation note. PTAL.

jonboulle:
LGTM
@vbatts any comments?

euank:
LGTM as well

PR #655, opened 2016-09-16
schema: Change the type of annotation's Name from ACIdentifier to string

yifan-gu:
This allows arbitrary strings to be used as annotation, which improves
the usability of the annotations.

/cc @jonboulle @philips @lucab @s-urbaniak

PR #656, opened 2016-09-19
Update appc data structures for CRI

squeed:
@yifan-gu @lucab @s-urbaniak I also updated the appc spec for annotations, along with port forwards. Take a look and we can update glide to pull from appc/spec's CRI branch.

### DIFF
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -28,13 +28,15 @@ import (
 const PodManifestKind = types.ACKind("PodManifest")
 
 type PodManifest struct {
-	ACVersion   types.SemVer        `json:"acVersion"`
-	ACKind      types.ACKind        `json:"acKind"`
-	Apps        AppList             `json:"apps"`
-	Volumes     []types.Volume      `json:"volumes"`
-	Isolators   []types.Isolator    `json:"isolators"`
-	Annotations types.Annotations   `json:"annotations"`
-	Ports       []types.ExposedPort `json:"ports"`
+	ACVersion      types.SemVer         `json:"acVersion"`
+	ACKind         types.ACKind         `json:"acKind"`
+	Apps           AppList              `json:"apps"`
+	Volumes        []types.Volume       `json:"volumes"`
+	Isolators      []types.Isolator     `json:"isolators"`
+	Annotations    types.Annotations    `json:"annotations"`
+	Ports          []types.ExposedPort  `json:"ports"`
+	CRIAnnotations types.CRIAnnotations `json:"criAnnotations,omitempty"`
+	CRILabels      types.CRILabels      `json:"criLabels,omitempty"`
 }
 
 // podManifest is a model to facilitate extra validation during the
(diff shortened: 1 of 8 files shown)


squeed:
@yifan-gu this supersedes #655, with a slightly cleaner interface. Sorry for stepping on your toes.

lucab:
For reference, this PR goes to a new `cri` branch which is aimed at being a playground and an early opportunity for reviewing changes related to supporting the new [Kubernetes CRI runtime interface](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/api/v1alpha1/runtime/api.proto) CRI includes a slightly different handling of pods and apps, and work is ongoing on the rkt side to support it.

schema/pod.go
```diff
@@ -28,13 +28,15 @@ import (
...
-	Annotations types.Annotations   `json:"annotations"`
-	Ports       []types.ExposedPort `json:"ports"`
+	ACVersion      types.SemVer         `json:"acVersion"`
+	ACKind         types.ACKind         `json:"acKind"`
+	Apps           AppList              `json:"apps"`
+	Volumes        []types.Volume       `json:"volumes"`
+	Isolators      []types.Isolator     `json:"isolators"`
+	Annotations    types.Annotations    `json:"annotations"`
+	Ports          []types.ExposedPort  `json:"ports"`
+	CRIAnnotations types.CRIAnnotations `json:"criAnnotations,omitempty"`
```
yifan-gu: @squeed CRIAnnotation and CRILabels sounds too specific for CRI? Why not just using the old Annotations and changing the type of the Key from ACIdentifier to string?
Both Label or Annotation can be muxed into annotations with prefixes.

squeed: @yifan-gu I thought about that, but the appc spec already defines some app annotations with special parsing rules, e.g. 'created'.
If we're going to add a type that allow arbitrary user-specified Kubernetes labels, we can't go adding more meaning to them.

squeed:
We need to make progress on other parts of the CRI push, so I'm merging this, but we don't need to stop the discussion.

schema/pod.go
lucab: This seems to be inline with previous discussion about this at https://github.com/coreos/rkt/issues/3010#issuecomment-238212558

yifan-gu:
@squeed We also need to add the same fields in the app level.

schema/types/cri_labels.go
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "fmt"
+
+type CRILabels map[ACIdentifier]string
```
yifan-gu: @squeed Why labels keys are not arbitrary strings as well? Today, the [labels](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/labels.go#L34-L35) injected by kubelet is not ACIdentifier compliant. 
Although there is a TODO to change that...

yifan-gu:
> @squeed We also need to add the same fields in the app level.

@squeed NVM, I saw them in the RuntimeApp.App...

PR #657, opened 2016-09-21
CRI: Change CRILabels type to map[string]string.

yifan-gu:
Follow https://github.com/appc/spec/pull/656#discussion_r79746750

Here I see two solutions:
1, use `CRIAnnotations` for those labels I mentioned in the above comments.
2, Merge this PR to allow arbitrary string in labels.

I am fine with both.

/cc @squeed @lucab @s-urbaniak


s-urbaniak:
LGTM

squeed:
@yifan-gu LGTM. This is funny - I was trying to do the right thing by adding validation :-).

PR #660, opened 2016-09-29
spec: add appData volume type

jonboulle:
Add a new volume type, similar to `empty` but prepopulated with data
from the app filesystem, similar to how [Docker
volumes](https://docs.docker.com/engine/tutorials/dockervolumes/#/data-volumes)
work.

@lucab wdyt?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,18 @@
+#<img src="http://kubernetes.io/kubernetes/img/warning.png" alt="WARNING" width="25" height="25"> Disclaimer <img src="http://kubernetes.io/kubernetes/img/warning.png" alt="WARNING" width="25" height="25">
+
+With the formation of the [Open Container Initiative (OCI)](https://www.opencontainers.org/), the industry has come together in a single location to define specifications around applications containers.
+OCI is intended to incorporate the best elements of existing container efforts like appc, and several of the appc maintainers are participating in OCI projects as maintainers and on the Technical Oversight Board (TOB).
+Accordingly, as of late 2016, appc is no longer being actively developed, other than minor tweaks to support existing implementations.
+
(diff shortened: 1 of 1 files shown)


euank:
Sounds like a sane thing to have to me.

I'm slightly iffy on the name because it implies the use of a copy-up filesystem, which really is an implementation detail. Maybe something like 'appData' or 'existingData'? Neither of those seem particularly better tan copyUp to me though, and I'm fine with copyUp too.

lucab:
rkt have a usecase  for this (mostly for inter-operability), so thanks for adding it. I'm not much into bikeshedding here, so feel free to pick the name you prefer. (If you actually need a tie-breaker, `appData` sound like the most obvious keyword to me)

squeed:
What happens if multiple apps in the pod mount the volume and have data to initialize?

We may need to declare this as a MountPoint option as well.

jonboulle:
Renamed to appData

jonboulle:
@squeed can you take this over?

squeed:
Hyup.

jonboulle:
@squeed bump?

squeed:
@jonboulle any thoughts on the ordering semantics?

jonboulle:
@squeed honestly I'd lean towards saying that's unspecified and discouraged behaviour. Is there a sensible usecase for something more controlled?

lucab:
As much as I don't like UB, I think this is probably one of those cases where it has to be. At best, this is related to application add/start ordering inside a both. Do we say explicitly say anything about that somewhere else in the spec? In that case, we could just refer to it.

jonboulle:
branch is at https://github.com/jonboulle/spec/tree/volthiing

PR #662, opened 2016-10-07
CRI: Move volume override from MountPoint to Mount

squeed:


schema/types/mountpoint.go
```diff
@@ -27,10 +27,9 @@ import (
 // The Volume is optional. If missing, the pod-level Volume of the
```
s-urbaniak: There is no `Volume` field in this struct any more.

PR #671, opened 2016-10-24
vendor: Update kubernetes dependency

euank:
third_party/forked/reflect is gone from upstream, pick up that change


squeed:
Cool, thanks!

squeed:
@euank looks like you forgot to add some new files that were incorporated as part of this change - I'll fix it.

squeed:
@euank My run of `glide-update` (with the changes in coreos/rkt#3313) generates a different result, see the [diff](https://github.com/euank/spec/compare/vendor-update-k8s...squeed:vendor-update-k8s). Any chance you just forgot to `git add -A vendor`?

euank:
@squeed I didn't apply the changes from that rkt pr, and without them + glide v0.12.3 I reliably get what's PR'd here. I did not forget to `git add` stuff.

I don't think we need the `lock-file` bit for glide-vc in this repo since we're not vendoring tools, so it's fine to let it strip a little more.

euank:
I cleaned up the glide script and removed the reference to `pkg/conversion` (that showed up in your diff from the lockfile but isn't referenced anywhere in this repo). Good catch on `pkg/conversion` being a bit odd there too!

jonboulle:
LGTM

jonboulle:
LGTM

PR #672, opened 2016-10-25
resource: re-export MaxMilliValue for consumers

euank:
Based on discussion in https://github.com/coreos/rkt/pull/3312, it seems like a nice thing to have available for consumers of this.


jonboulle:
```
gopath/src/github.com/appc/spec/schema/types/isolator_resources.go:46: const initializer resource.MaxMilliValue is not a constant
```

euank:
I'm closing this since it doesn't solve a real problem and I think this change would make more sense as part of a larger "no longer depend on K8s resource semantics" refactor in appc (which might never happen).

PR #673, opened 2016-11-14
spec: add os/linux/selinux-context isolator

lucab:
This commit introduces an "os/linux/selinux-context" isolator, which
can be applied at both app and pod scope.

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -200,6 +200,38 @@ In the example above, the process will not be allowed to invoke `clock_adjtime(2
 
 In the example above, the process will be only allowed to invoke syscalls specified in the custom network-related set (and any other syscall in the implementation-specific whitelist). All other syscalls will result in app termination via `SIGSYS` signal.
 
+#### os/linux/selinux-context
+
+* Scope: app/pod
+
+**Parameters:**
+
+* **user** case-sensitive string containing the user portion of the SELinux security context to be used to label the current pod or application.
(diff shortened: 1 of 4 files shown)


lucab:
@jonboulle @squeed PTAL.

spec/ace.md
```diff
@@ -200,6 +200,37 @@ In the example above, the process will not be allowed to invoke `clock_adjtime(2
...
+
+**Parameters:**
+
+* **user** case-sensitive string containing the user portion of the SELinux security context to be used to label the current pod or application.
+* **role** case-sensitive string containing the role portion of the SELinux security context to be used to label the current pod or application.
+* **type** case-sensitive string containing the type/domain portion of the SELinux security context to be used to label the current pod or application.
+* **level** case-sensitive string containing the level portion of the SELinux security context to be used to label the current pod or application.
+
+**Notes:**
+ 1. Only a single `os/linux/selinux-context` isolator can be specified per-pod.
```
jonboulle: If a context is applied to a pod, does this just mean it's for all apps (unless they have overrides), or also for any executor processes involved in running the pod? (i.e. stage1 for rkt)

jonboulle:
LGTM modulo clarification on pod thing

spec/ace.md
lucab: It also applies to other processes in the pod context. Thanks for highlighting this, I'll make it explicit in the spec.

lucab:
Amended with the clarification regarding pod-context.

squeed:
:+1:

PR #674, opened 2016-11-17
ace validator: Remove extra environment variables

uiri:
These environment variables are no longer mandated by the spec per
    appc/spec#273. It makes sense to remove them from the validator.


jonboulle:
LGTM

PR #676, opened 2016-11-20
spec/ace: Add HTTP methods for metadata service

uiri:
The metadata service endpoints for creating and verifying an HMAC
signature already specify use of the POST method in their
descriptions. The other metadata service endpoints make no mention of
which HTTP methods are valid for them.

The current `rkt` implementation uses POST for the HMAC-related
endpoints and GET for all the others. This makes such behaviour
explicit in the spec.


jonboulle:
LGTM, thanks for the clarification

PR #677, opened 2016-11-21
spec/ace: Specify Content-Type for metadata service.

uiri:
Some of the metadata service endpoints explicitly specify a
Content-Type of `text/plain; charset=us-ascii`. For most other
endpoints, the ACE validator expects a Content-Type of
`application/json`.

The one exception is the HMAC verify endpoint which makes no mention
of its Content-Type nor even the contents of its response body. The
current implementation in `rkt` sets the Content-Type to
`text/plain; charset=us-ascii` which is now reflected in the spec.

### DIFF
--- a/spec/ace.md
+++ b/spec/ace.md
@@ -474,11 +474,11 @@ Information about the pod that this app is executing in.
 
 Retrievable at `$AC_METADATA_URL/acMetadata/v1/pod`
 
-| Entry       | Method | Description |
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM, thanks (I was wondering whether these should be in backticks, as you've done in the commit message, but can see arguments either way, so will leave it for now - feel free to follow up if you want to change that).

PR #682, opened 2017-01-13
switch to github errorutil

deads2k:
`github.com/abh/errorutil` is identical to `go4.org/errorutil` and go4.org is currently down and only used for this single dependency.  I found this in a downstream project (kubernetes) that could no longer godep.

This is my first time using glide, so I think I've done it all correctly, but double checking would be appreciated.

@vbatts @eparis


glide.yaml
```diff
@@ -25,3 +21,4 @@ import:
   version: ~1.4.0
   subpackages:
   - pkg/api/resource
+- package: github.com/abh/errorutil
```
deads2k: It wasn't created automatically, so I don't actually know.  Without any branches or tags to specify an actual version, I'm not sure what it would buy us over the lock file.

euank:
The source package on `go4.org` is already on [github](https://github.com/camlistore/go4/tree/master/errorutil) too. Using glide's `repo` key in glide.yaml, we can specify to fetch the go4.org package from github while still using its canonical (and broken) import path.
However, that breaks go-gettability and people using other vendoring systems like godeps I think.

It might all be a moot point though since in the time it took me to type this comment, go4.org is back up again 😄

deads2k:
> It might all be a moot point though since in the time it took me to type this comment, go4.org is back up again

16 hours down and it comes back right as a I finally make glide sort of run.  Figures.

euank:
@deads2k I commented on the k8s issue too. It might be worth switching it still just based on it having a shaky track record at this point.

eparis:
I see no reason to use an indirection like `go4.org`/`camlistore/go4` when we actually want the real thing. Why would we want to use a collection of packages more slowly updated instead of just going to the real source? Even if it seems like the real source is unlikely to move.

Even thought go4 came back to life, I think this is the right thing to do.

vbatts:
The indirection is _can_ be handy, but often frustrating. Not all turtles stack equally.

vbatts:
how come the LICENSE was dropped? I see that the newly pointed repo doesn't have it, but it seems the attribution should still be there.

eparis:
What attribution? go4.org (aka https://github.com/camlistore/go4) didn't write any of this code. It is just a curated collection of others work. We aren't using the collection. We are just using the one thing. If the one thing doesn't have a LICENSE why should be vendor some other random LICENSE from an completely unrelated repository?

eparis:
actually, as I look, maybe you are right...  go4 isn't 'unrelated'...

vbatts:
so should we switch or not?

euank:
I think we do want the `version` specified to a commit if we do switch over, so this PR would need an additional slight change.

Since it hasn't been directly linked here, the Kubernetes issue that resulted in this change is https://github.com/kubernetes/kubernetes/issues/39868.

On that issue, the consensus seems to roughly be that it's really K8s responsibility to better avoid this sort of error causing so much blockage/pain.

I'm indifferent to this change.

eparis:
Since I now believe that camlistore/go4 is in fact the parent package and abh/errorutil is in fact the clone/child I believe that appc/spec is pulling from the right upstream. Thus i think we should close this.

PR #687, opened 2017-04-28
pod: now reject duplicate volume names in manifest

lucab:
Noteworthy: validation check now rejects manifest with duplicate volume names.

This PR adds a validation check on the `volumes` array to ensure that it doesn't contain any duplicate-named volumes (and corresponding tests). Also, it clarifies string-typing of annotations.

### DIFF
--- a/schema/pod.go
+++ b/schema/pod.go
@@ -83,6 +83,15 @@ func (pm *PodManifest) assertValid() error {
 	if pm.ACKind != PodManifestKind {
 		return pmKindError
 	}
+
+	// ensure volumes names are unique (unique key)
+	volNames := make(map[types.ACName]bool, len(pm.Volumes))
+	for _, vol := range pm.Volumes {
+		if volNames[vol.Name] {
+			return fmt.Errorf("duplicate volume name %q", vol.Name)
+		}
+		volNames[vol.Name] = true
+	}
 	return nil
 }
 
--- a/spec/aci.md
+++ b/spec/aci.md
@@ -214,7 +214,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
         },
         {
             "name": "appc.io/executor/supports-systemd-notify",
-            "value": false
+            "value": "false"
         }
     ]
 }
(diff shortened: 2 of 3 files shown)


spec/aci.md
```diff
@@ -256,7 +256,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
...
     * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **appc.io/executor/supports-systemd-notify** (boolean, optional, defaults to "false" if unset) if set to true, the application SHOULD use the sd\_notify mechanism to signal when it is ready. Also it SHOULD be able to detect if the executor had not set up the sd\_notify mechanism and skip the notification without error ([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) from libsystemd does that automatically).
+    * **appc.io/executor/supports-systemd-notify** (string boolean, optional, defaults to "false" if unset) if set to "true", the application SHOULD use the sd\_notify mechanism to signal when it is ready. Also it SHOULD be able to detect if the executor had not set up the sd\_notify mechanism and skip the notification without error ([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html) from libsystemd does that automatically).
```
dgonyeo: Wait, why is this a string, if it can only be `"true"` or `"false"`?

Was it before this PR, and this is just a docs update? I don't see the change to the types here.

jonboulle:
Because it's an annotation

Am 28.04.2017 20:24 schrieb "Derek Gonyeo" <notifications@github.com>:

*@dgonyeo* commented on this pull request.
------------------------------

In spec/aci.md <https://github.com/appc/spec/pull/687#discussion_r113991462>
:

> @@ -256,7 +256,7 @@ JSON Schema for the Image Manifest (app image manifest, ACI manifest), conformin
     * **authors** contact details of the people or organization
responsible for the image (freeform string)
     * **homepage** URL to find more information on the image (string,
must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image
(string, must be a URL with scheme HTTP or HTTPS)
-    * **appc.io/executor/supports-systemd-notify** (boolean,
optional, defaults to "false" if unset) if set to true, the
application SHOULD use the sd\_notify mechanism to signal when it is
ready. Also it SHOULD be able to detect if the executor had not set up
the sd\_notify mechanism and skip the notification without error
([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html)
from libsystemd does that automatically).
+    * **appc.io/executor/supports-systemd-notify** (string boolean,
optional, defaults to "false" if unset) if set to "true", the
application SHOULD use the sd\_notify mechanism to signal when it is
ready. Also it SHOULD be able to detect if the executor had not set up
the sd\_notify mechanism and skip the notification without error
([sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html)
from libsystemd does that automatically).

Wait, why is this a string, if it can only be "true" or "false"?

Was it before this PR, and this is just a docs update? I don't see the
change to the types here.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/appc/spec/pull/687#pullrequestreview-35429324>, or mute
the thread

schema/pod.go
```diff
@@ -83,6 +83,15 @@ func (pm *PodManifest) assertValid() error {
 	if pm.ACKind != PodManifestKind {
 		return pmKindError
 	}
+
+	// ensure volumes names are unique (unique key)
+	volNames := make(map[types.ACName]bool, len(pm.Volumes))
```
euank: dumb nit, we can save a couple bits if we use `struct{}` in place of `bool`. I'm fine with it either way.

spec/aci.md
lucab: As @jonboulle said, this just fixes a wrong example/wording conflicting with the previous definition of annotations a couple of lines above: `value is an arbitrary string`.

schema/pod.go
lucab: I thought this was golang-idiomatic to make the if-has-key check more compact. The same pattern is used in other places in appc/spec, so I'd leave it as is.

lucab:
@euank thanks for feedback, I've updated title and PR comment to highlight that this is now rejecting out-of-spec manifests that were previously accepted. This way the information should also propagate into release notes.

euank:
rkt's release notes are what really need this imo. Still 👍 from me

PR #688, opened 2017-05-10
ace: Fix manifest arch error

glevand:
The manifest contains values for the ACI arch and OS, not the go
language values.  To avoid confusion, change the symbols 'GOOS" and
'GOARCH' to 'ACI_OS' and 'ACI_ARCH', and provide conversion as
needed.

Fixes errors like these when run an ARM64 machines:

  bad arch "arm64" for linux (must be one of: [amd64 i386 aarch64 aarch64_be armv6l armv7l armv7b ppc64 ppc64le s390x])


squeed:
LGTM too.

PR #692, opened 2017-09-06
linuxSeccompBase: Fix errno check

glevand:
errno symbols can contain numeric characters.

### DIFF
--- a/schema/types/isolator_linux_specific.go
+++ b/schema/types/isolator_linux_specific.go
@@ -226,9 +226,13 @@ func (l linuxSeccompBase) AssertValid() error {
 	if l.val.Errno == "" {
 		return nil
 	}
-	for _, c := range l.val.Errno {
-		if !unicode.IsUpper(c) {
-			return errors.New("errno must be an upper case string")
+	for i, c := range l.val.Errno {
+		if i == 0 && c != 'E' {
+			s := fmt.Sprintf("errno must start with an 'E' character, got %s", l.val.Errno)
+			return errors.New(s)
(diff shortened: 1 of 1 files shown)


alban:
LGTM

Examples:
```
$ errno --list|gawk '{print $1}'|grep [0-9]
E2BIG
EL2NSYNC
EL3HLT
EL3RST
EL2HLT
```

schema/types/isolator_linux_specific.go
```diff
@@ -227,7 +227,7 @@ func (l linuxSeccompBase) AssertValid() error {
 		return nil
 	}
 	for _, c := range l.val.Errno {
-		if !unicode.IsUpper(c) {
+		if !(unicode.IsUpper(c) || unicode.IsNumber(c)) {
 			return errors.New("errno must be an upper case string")
```
jellonek: This error message could be also updated there...
E.g. "errno must be a string containing only digits and upper case characters"

lucab:
Definitely my bad, I forgot some of them contain numbers. Perhaps a better check would be `starts with uppercase "E" and contains only uppercase or digits`.

schema/types/isolator_linux_specific.go
lucab: As we wish, I think the additional OR-condition doesn't change the spirit of the previous error label. @jellonek's suggestion is strictly speaking more correct.

glevand:
Updated patch to do the checks as suggested.

cc: @lucab

lucab:
(go fmt is unhappy about the latest rebase)

glevand:
Fixed gofmt.

PR #695, opened 2017-12-09
spec: add os/linux/apparmor-profile isolator

cyphar:
AppArmor is a very widely used isolator in the GNU/Linux community,
providing an "easier to configure" alternative to SELinux. From the
perspective of AppC this is a fairly simple addition (profile data is
not included in the spec and must already be loaded into the kernel).

This feature is required for feature parity for ACI containers with OCI
ones (as AppArmor is also supported as a first-class citizen in OCI).

/cc @flavio @vrothberg
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/schema/types/isolator_linux_specific.go
+++ b/schema/types/isolator_linux_specific.go
@@ -31,6 +31,7 @@ const (
 	LinuxOOMScoreAdjName           = "os/linux/oom-score-adj"
 	LinuxCPUSharesName             = "os/linux/cpu-shares"
 	LinuxSELinuxContextName        = "os/linux/selinux-context"
+	LinuxAppArmorProfileName       = "os/linux/apparmor-profile"
 )
 
 var LinuxIsolatorNames = make(map[ACIdentifier]struct{})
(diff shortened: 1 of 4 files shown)


jonboulle:
Changeset looks fine - but curious on the motivation/appropriateness given appc is not really being developed any more. I guess this is for https://github.com/rkt/rkt/issues/865 ? Do you know of an extant use case?

cyphar:
The "higher level" plan is that I'd like to have (at least some) feature parity from the ACI side so that implementing an OCI stage1 can be done by just mapping the ACI bits to OCI. AppArmor support is one of the only things that are quite important that are missing (for us at least -- though the seccomp profile scheme in ACI isn't quite as full-featured as the support in OCI which makes parity not as nice as it could be). There already is a patchset that adds AppArmor support to `rkt` in rkt/rkt#3765, but it's done in a way that wouldn't help other `stage1` implementations use it.

jonboulle:
Thanks for the context. Given that this LGTM. Obviously projects will need to update to make any use of this.

